PR #1, opened 2020-01-05
Extract the OpenSlides 3 client

tsiegleauq:
Extracts the client for OpenSlides 4

Add a docker config for both
production build and developing toolchain

Set and adjust the readme


README.md
```diff
@@ -1,3 +1,164 @@
-# OpenSlides client
+# OpenSlides Client
 
-Webclient for OpenSlides which ...
+Web client for OpenSlides which is written in Angular.
```
normanjaeckel: Hm. The ... which ... clause was ment to describe what this part of OpenSlides is doing. But if "client" is self explaining you may write only "Web client for OpenSlides 4+". Please also adjust repository description in GitHub.

README.md
```diff
@@ -1,3 +1,164 @@
-# OpenSlides client
+# OpenSlides Client
```
normanjaeckel: OpenSlides client or OpenSlides Client? @emanuelschuetze

normanjaeckel:
Maybe also add ``.travis.yml``. After that we can enable status checks for further pull requests.

README.md
emanuelschuetze: Let us use always capital letters for service names! As in our [architecture diagram](https://raw.githubusercontent.com/OpenSlides/OpenSlides/openslides4-dev/docs/OpenSlides4-systemarchitecture.png). E.g. "OpenSlides Backend Service", not "OpenSlides backend service".

tsiegleauq:
Added `.travis.yml`
Adjusted the readme even more
Extended the "which ..." clause

README.md
normanjaeckel: Ok.

normanjaeckel:
Ok.

PR #9, opened 2020-07-14
Adds action.service

GabrielInTheWorld:
- Replaces previous http-requests with current available actions
- Make an `actions.md` for documentation-purposes. This document contains a list with current available actions and a list with actions, that have to implement.


Makefile
```diff
@@ -2,10 +2,10 @@ build-dev:
 	docker build -t openslides-client-dev -f Dockerfile.dev .
 
 run-dev: | build-dev
-	docker run -t -v `pwd`/client:/app -p 127.0.0.1:4200:4200/tcp openslides-client-dev
+	docker run -ti -v `pwd`/client:/app -p 127.0.0.1:4200:9001/tcp openslides-client-dev
```
FinnStutzenstein, line 5: please also change the port in the `package.json`

client/src/app/core/core-services/action.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    MOTION_UPDATE_METADATA = 'motion.update_metadata',
+    COMMITEE_CREATE = 'commitee.create',
+    MEETING_CREATE = 'meeting.create',
+    MEETING_UPDATE = 'meeting.update',
+    MEETING_DELETE = 'meeting.delete'
+}
+
+export interface RequestInfo<T> {
+    action: ActionType;
+    // data: (keyof T)[];
```
FinnStutzenstein, line 27: please remove

client/src/app/core/core-services/action.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    data: any;
+}
+
+@Injectable({
+    providedIn: 'root'
+})
+export class ActionService {
+    private readonly BACKEND_URL = '/system/action/handle_request';
```
FinnStutzenstein, line 35: `ACTION_URL`

Makefile
jsangmeister: I thought the port 9001 was needed for the full setup? I specifically only touched this to be able to access it. If you want, I can set the host port to 9001, too.

client/src/app/core/repositories/base-repository.ts
```diff
@@ -99,8 +99,8 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
         return this.repositoryServiceCollector.DS;
     }
 
-    protected get dataSend(): DataSendService {
-        return this.repositoryServiceCollector.dataSend;
+    protected get actions(): ActionService {
```
FinnStutzenstein, line 102: Pleas euse the singular term. But a direct `actionService` would be the best

client/src/app/core/repositories/base-repository.ts
```diff
@@ -253,21 +253,28 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
...
 
     /**
      * patches an existing model with new data,
      * rather than sending a full update
      *
      * @param update the update to send
-     * @param viewModel the motion to update
+     * @param viewModel the view model to update
      */
     public async patch(update: Partial<M>, viewModel: V): Promise<void> {
```
FinnStutzenstein, line 266: Can we eliminate `patch` here, or is it too much work?

If so, please add a `@depricated` and just call `update` internally. Note that the id-injection has to be done in the update.

client/src/app/core/repositories/motions/motion-repository.service.ts
```diff
@@ -1,9 +1,11 @@
 import { Injectable } from '@angular/core';
 
 import { Observable } from 'rxjs';
+import { Action } from 'rxjs/internal/scheduler/Action';
```
FinnStutzenstein, line 4: This doesn't seem right

client/src/app/core/repositories/users/user-repository.service.ts
```diff
@@ -313,55 +310,57 @@ export class UserRepositoryService extends BaseRepository<ViewUser, User> {
...
-                const lastUsername = usernames.pop();
-                userString = usernames.join(', ') + ' ' + this.translate.instant('and') + ' ' + lastUsername;
-            } else {
-                userString = usernames.join(', ');
-            }
-            msg = msg.replace('%user%', userString);
-        }
-
-        return msg;
+        // const response = await this.httpService.post<{ count: number; no_email_ids: number[] }>(
```
FinnStutzenstein, line 313: Please adjust the comments to `/* */`

client/src/app/site/common/components/start/start.component.ts
```diff
@@ -142,7 +130,7 @@ export class StartComponent extends BaseComponent implements OnInit {
     }
 
     public test(): void {
```
FinnStutzenstein, line 132: I do not need these test anymore. Can you remove it completly (incl. the button in the ui)?

Makefile
FinnStutzenstein: Yes, then it is consistent at all places

client/src/app/core/repositories/base-repository.ts
jsangmeister: Already done ;)

jsangmeister:
Done.

client/src/app/core/core-services/action.service.ts
```diff
@@ -0,0 +1,47 @@
...
+
+@Injectable({
+    providedIn: 'root'
+})
+export class ActionService {
+    private readonly ACTION_URL = '/system/action/handle_request';
+
+    private constructor(private http: HttpService) {}
+
+    public async sendRequest<T>(action: ActionType, data: any[]): Promise<T> {
```
FinnStutzenstein, line 38: One little thing to discuss: In most cases `data: any` would be sufficient. We need this feature for bulk-updates. We should provide an extra function for this, so in the usual cases, the user of this service does not have to wonder why an array is needed.

client/src/app/core/core-services/action.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    MOTION_DELETE = 'motion.delete',
+    MOTION_SORT = 'motion.sort',
+    MOTION_UPDATE_METADATA = 'motion.update_metadata',
+    COMMITEE_CREATE = 'commitee.create',
+    MEETING_CREATE = 'meeting.create',
+    MEETING_UPDATE = 'meeting.update',
+    MEETING_DELETE = 'meeting.delete'
+}
+
+export interface RequestInfo {
```
jsangmeister, line 25: @GabrielInTheWorld Since the Type Variable wasn't used, I removed it. If you need it for something, feel free to add it back

jsangmeister:
I'm currently not able to test the code since the backend refuses to do what I want... ;) Will look into this next week.

PR #10, opened 2020-07-17
Fix agendaitem<->tag relation

FinnStutzenstein:


FinnStutzenstein:
Regarding the changelog: This is a general "Redoing it, move to OS4" thing. Until we start implementing actual features, this should not bother us ;)

PR #12, opened 2020-07-22
Repair topic and user view

tsiegleauq:
Add the new autoupdate routine to topic and user detail view.
Some slight restructure for model contexts.

### DIFF
--- a/client/src/app/site/tags/components/tag-list/tag-list.component.ts
+++ b/client/src/app/site/tags/components/tag-list/tag-list.component.ts
@@ -85,7 +85,7 @@ export class TagListComponent extends BaseListViewComponent<ViewTag> implements 
                     follow: [
                         {
                             idField: 'tagged_ids',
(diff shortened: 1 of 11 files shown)


client/src/app/site/tags/components/tag-list/tag-list.component.ts
```diff
@@ -85,7 +85,7 @@ export class TagListComponent extends BaseListViewComponent<ViewTag> implements
                     follow: [
                         {
                             idField: 'tagged_ids',
-                            fieldset: ['title'] // TODO
+                            fieldset: ['title', 'name'] // TODO
```
FinnStutzenstein, line 88: This is the fieldset of the tagged objects. Plase change it from the array to a string specifing the default fieldset -> For generic lookups we do not exactly know, which fields are required -> If all models have a common fieldset like `display`/`list` just for displaying purposes, this fieldset is the right one to use here.

PR #15, opened 2020-07-23
Adjust assignments to new autoupdate

tsiegleauq:
Adjust assignment list and detail to be compatible to the new
autoupdate system

### DIFF
--- a/client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
+++ b/client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
@@ -14,7 +14,7 @@ import { ComponentServiceCollector } from 'app/core/ui-services/component-servic
 import { PromptService } from 'app/core/ui-services/prompt.service';
 import { Assignment } from 'app/shared/models/assignments/assignment';
 import { ViewAgendaItem } from 'app/site/agenda/models/view-agenda-item';
-import { BaseComponent } from 'app/site/base/components/base.component';
+import { BaseModelContextComponent } from 'app/site/base/components/base-model-context.component';
 import { ViewMediafile } from 'app/site/mediafiles/models/view-mediafile';
 import { PermissionsService } from 'app/site/motions/services/permissions.service';
 import { ViewTag } from 'app/site/tags/models/view-tag';
@@ -25,6 +25,7 @@ import { AssignmentPollService } from '../../services/assignment-poll.service';
 import { AssignmentPhases, ViewAssignment } from '../../models/view-assignment';
 import { ViewAssignmentCandidate } from '../../models/view-assignment-candidate';
 import { ViewAssignmentPoll } from '../../models/view-assignment-poll';
+import { SPEAKER_BUTTON_FOLLOW } from 'app/shared/components/speaker-button/speaker-button.component';
 
 /**
  * Component for the assignment detail view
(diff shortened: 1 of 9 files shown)


tsiegleauq:
> Number of candidates badge shows 0

was apparently fixed by your commit

> reload doesn't work
> candidates are not shown

Fixed thouse. Polls seem to be a little more complicated to resolve.
Please note the `FIXME`, not entirely sure if I am using the candidates / users the intended way.

client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
```diff
@@ -111,6 +112,16 @@ export class AssignmentDetailComponent extends BaseComponent implements OnInit {
      */
     private _assignment: ViewAssignment;
 
+    public get candidates(): ViewUser[] {
```
FinnStutzenstein, line 115: Please undo this. `candidates` is auto-generated and the type `ViewAssignmentCandidate[]` is correct, so you can access the m2m model (analog to submitters). If you directly need users (It's just a shortcut...), use `this.candidatesAsUsers` (not `this.assignment.candidatesAsUsers`, it is not defined for the class `Assignment`).

client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
```diff
@@ -111,6 +112,16 @@ export class AssignmentDetailComponent extends BaseComponent implements OnInit {
...
+        /**
+         * FIXME:
+         * Confirm: Not sure if candidatesAsUsers or candidates (as AssignmentCandidate)
+         * would be required for the lists.
+         * AssignmentCandidate .getTitle always returns "Candidate" instead of the users name
```
FinnStutzenstein, line 120: This is ok, since it is an m2m model and (potentially) the user is not loaded.

client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
```diff
@@ -111,6 +112,16 @@ export class AssignmentDetailComponent extends BaseComponent implements OnInit {
      */
     private _assignment: ViewAssignment;
 
+    public get candidates(): ViewUser[] {
+        /**
+         * FIXME:
+         * Confirm: Not sure if candidatesAsUsers or candidates (as AssignmentCandidate)
```
FinnStutzenstein, line 118: `candidatesAsUsers` is ok: It is just a `map` from the m2m model (`ViewAssignmentCandidate`) to the linked `ViewUser`. It's should be correctly sorted[1]. So if you just want to display, it is enough. For sorting the m2m model should be used.

[1] I just checked the `relations.ts`: The order is missing. I'll add it to this PR.

FinnStutzenstein: Done.

FinnStutzenstein:
> > Number of candidates badge shows 0
> 
> was apparently fixed by your commit

I cannot confirm this. If I reload the list, both badges show 0. You should check, if the counting relies on just the id-array (like `candidate_ids.length`) or actually counting view-models (something like `candidates.length`). In the latter case, you need to either change the logic, or request the candidates, too (without users should be sufficient).

A note about the fieldset you changed: It is not worthy to include id-fields there. They will not be resolved and most time, they are useless without resolving them. And for things like the list, where you might need just the id-array, you can include it in `additionalFields`. THis is a list of fields to request next to the fields in the fieldset.

tsiegleauq:
> See comments above. The detail page works fine (except for polls, this is another thing for later).

no comments to resolve except 

> I cannot confirm this

Which is pretty odd. Perhaps due to the client?

FinnStutzenstein:
> no comments to resolve except

The fixme is ok, so it can be removed.

Regarding the candidate-amount-badge: Let's talk about this after the meetings later.

tsiegleauq:
> Number of candidates badge shows 0
ok. I confused the speaker badge with the amount of candidates.

tsiegleauq:
could be done now.
I added the candidate_ids (array of numbers) to the list, since it's the most reliable and direct way to determine the amount of candidates on the list view.

PR #17, opened 2020-07-27
 Adjust Motion To New Autoupdate

tsiegleauq:


client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
```diff
@@ -567,6 +561,29 @@ export class MotionDetailComponent extends BaseComponent implements OnInit, OnDe
         });
     }
 
+    private registerSubjects(): void {
+        // get required information from the repositories
+        this.tagObserver = this.tagRepo.getViewModelListBehaviorSubject();
```
FinnStutzenstein, line 566: If you do not need `this.tagObserver`, ..., I have a proposal: Make an array of repos and loop through them.

client/src/app/core/repositories/motions/motion-repository.service.ts
```diff
@@ -104,12 +104,32 @@ export class MotionRepositoryService extends BaseIsAgendaItemAndListOfSpeakersCo
             'category_weight',
             'created',
             'last_modified',
-            'sequential_number'
+            'sequential_number',
+            'workflow_id',
```
FinnStutzenstein, line 108: Same as in the other PR: Do you have an advantage of having id fields declared here?

client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
tsiegleauq: you need to access the subjects in the view aswell.

tsiegleauq:
still need to rebase

@FinnStutzenstein ping, can be reviewed at most.
most stuff works, and can reload without crashing.
"previous and next motion" is still a little difficult.
Some stuff kinda depends on the groups to work.
Motion-state-detail does not show the state, please see the @fixme

client/src/app/site/motions/modules/motion-workflow/components/workflow-detail/workflow-detail.component.ts
```diff
@@ -107,7 +107,12 @@ export class WorkflowDetailComponent extends BaseComponent implements OnInit {
         { name: 'Allow support', selector: 'allow_support', type: 'check' },
         { name: 'Allow create poll', selector: 'allow_create_poll', type: 'check' },
         { name: 'Allow submitter edit', selector: 'allow_submitter_edit', type: 'check' },
-        { name: 'Do not set number', selector: 'dont_set_number', type: 'check' },
+        /**
+         * FIXME:
+         * Discuss:
+         * Is it now "do" or "do not" set number?
```
FinnStutzenstein, line 113: it is set_number. We just want to have positive names.

FinnStutzenstein:
@tsiegleauq The issues were mostly incorrect example data; I fixed it in the main repo. You have to `make stop-dev` to reset the db to get the new example data.

> "previous and next motion" is still a little difficult.

Yes - does it include the custom sporting from the list view?

tsiegleauq:
> Yes - does it include the custom sporting from the list view?

It should, but I did not implement it on OS4 yet. I suppose I have to make another call of requestModels with the previous motion id and the next motion id, while requesting only the number(identifier) and title.
The sorting you are using on the list view directly reflects what the current previous and next motion of your current motion is

FinnStutzenstein:
Ok, not in this PR. This one is tricky because: You need to have all motions (e.g. only the title/indentifier) and the fields to sort after. E.g. if the user sorts by category, you additionally need `categeory_id` and `category_wight` (hopefully I'm correct here). Then you have to sort the list returned by the server, search the current motion and go index+-1. The problem is that you must have the complete list to get the next/previous motion, if custom sorting is used...

But this should not be a performance problem: I mean the list view also queries all motions and sorts them. We just do not query the text/reason for each motion and we are fine i guess.

tsiegleauq:
sounds like material for later indeed

PR #22, opened 2020-07-31
Add mediafile AU

tsiegleauq:
browsing example data works
Selecting mediafiles in a search value selector was not covered here

### DIFF
--- a/client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.html
+++ b/client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.html
@@ -41,7 +41,7 @@
                     *ngIf="!last"
                 >
                     <span class="folder-text">
-                        {{ directory.filename }}
+                        {{ directory.title }}
                     </span>
                 </button>
                 <button
@@ -53,11 +53,11 @@
                     *ngIf="last && showFileMenu(directory)"
                 >
                     <os-icon-container icon="arrow_drop_down" swap="true" size="large" [noWrap]="true">
-                        {{ directory.filename }}
+                        {{ directory.title }}
                     </os-icon-container>
                 </button>
                 <span class="folder fake-folder folder-text" *ngIf="last && !showFileMenu(directory)">
-                    {{ directory.filename }}
+                    {{ directory.title }}
                 </span>
             </span>
         </span>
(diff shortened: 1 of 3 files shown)


client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.ts
```diff
@@ -216,18 +218,24 @@ export class MediafileListComponent extends BaseListViewComponent<ViewMediafile>
...
     public ngOnInit(): void {
+        super.ngOnInit();
         super.setTitle('Files');
 
         this.repo.getDirectoryIdByPath(this.route.snapshot.url.map(x => x.path)).then(directoryId => {
             this.changeDirectory(directoryId);
         });
 
         // Observe the logo actions
+        /**
```
FinnStutzenstein, line 229: The logo/font handling has changed completly: It should also be done in this PR.

Mediafiles have template fields `used_as_logo_$_in_meeting` and `used_as_font_$_in_meeting`. The values to replace are the places for the mediafiles. E.g. for logos there can be `web_header`, `pdf_header`, `pdf_footer`, ... (see https://github.com/OpenSlides/OpenSlides/blob/master/openslides/core/config_variables.py#L255 without the `logo_` prefix)

This list must now be defined in the client - the server does not know about these fields anymore. It implies, that the client can define fields, as it likes without server changes.

So, if a mediafile is used as the `web_header` and `pdf_header`, the fields `used_as_logo_web_header_in_meeting` and `used_as_logo_pdf_header_in_meeting` are set.

Maybe the `MediaManageService` can save the list of possible places and do some handling of that.

client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.html
```diff
@@ -41,7 +41,7 @@ <h2>{{ 'Files' | translate }}</h2>
                     *ngIf="!last"
                 >
                     <span class="folder-text">
-                        {{ directory.filename }}
+                        {{ directory.name }}
```
FinnStutzenstein, line 44: please change to title

client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.ts
FinnStutzenstein: For selecting a place in the dot-menu: There will be a dedicated action for setting a mediafile in a meeting as X (e.g. `projector_header`).

tsiegleauq: It's still not clear how and from where to get the information about the currently used logo(s) and fonts, this this feature is way beyond the scope of this pull request.

PR #23, opened 2020-08-03
Add build action

tsiegleauq:
Try to ensure that `make build-dev` works all the time


tsiegleauq:
Not sure if start should also be tested. I suppose testing build would be enough for the time being.
@FinnStutzenstein any opinion on this?

FinnStutzenstein:
You should add thinks like `check-formatting` (analog to `make run-cleanup`, but just for the exit code) and call `make run-tests`. I'm not sure, if `make run-tests` currently works. We should not start it, but should build the client (`make build-dev` just builds the container, not the application). The prod build can be deferred to the point, where we enable prod in all services...

tsiegleauq:
okay, so far we covered:
- deps
- lint
- prettier
- debug build
- docker build dev

I am not entirely sure if are able to ensure that the docker container starts successfully, since it tries to keep angulars dev server up and running.

tsiegleauq:
tests are now executed inside the docker container using nothing but `make run-tests`
should be done now

tsiegleauq:
I suppose it's as good and as fast as it gets for the moment.
Unfortunately, we cannot really run the tests in parallel, since we cannot pass the docker container between the jobs. We could, however, run certain tests in parallel outside of docker. I might think about that later.

PR #24, opened 2020-08-03
Replace media manage service

tsiegleauq:
Adjusts the selection of manageable media

### DIFF
--- a/client/src/app/core/ui-services/media-manage.service.ts
+++ b/client/src/app/core/ui-services/media-manage.service.ts
@@ -1,15 +1,12 @@
 import { Injectable } from '@angular/core';
 
-import { Observable } from 'rxjs';
-
-import { HttpService } from 'app/core/core-services/http.service';
 import { OrganisationSettingsService } from 'app/core/ui-services/organisation-settings.service';
 import { ViewMediafile } from '../../site/mediafiles/models/view-mediafile';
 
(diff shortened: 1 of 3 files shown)


client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.ts
```diff
@@ -397,7 +386,8 @@ export class MediafileListComponent extends BaseListViewComponent<ViewMediafile>
      * @returns the display name of the selected action
      */
     public getNameOfAction(mediaFileAction: string): string {
```
FinnStutzenstein, line 388: A quick request: Can you find another name than `action`. It raises confusion with our new concepts.. E.g. logos and fonts can have usages/places where they are used. You'll find something appropriate I think :)

client/src/app/core/ui-services/media-manage.service.ts
```diff
@@ -43,6 +53,14 @@ interface ManagementPayload {
     providedIn: 'root'
 })
 export class MediaManageService {
+    public get logoSettings(): string[] {
```
FinnStutzenstein, line 56: The same naming should be adopted here. (I wonder that the client was inconsistent with the naming before: A mix of `action` and `setting`?)

PR #26, opened 2020-08-05
Request all requires updates in motion create

tsiegleauq:
Fix an error in new motion view
Motion detail will request all/most required updates

### DIFF
--- a/client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
+++ b/client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
@@ -39,6 +39,7 @@ import { Motion } from 'app/shared/models/motions/motion';
 import { ViewUnifiedChange } from 'app/shared/models/motions/view-unified-change';
 import { infoDialogSettings, mediumDialogSettings } from 'app/shared/utils/dialog-settings';
 import { BaseModelContextComponent } from 'app/site/base/components/base-model-context.component';
+import { ViewMeeting } from 'app/site/event-management/models/view-meeting';
 import { CreateMotion } from 'app/site/motions/models/create-motion';
 import { ViewCreateMotion } from 'app/site/motions/models/view-create-motion';
 import { ViewMotion } from 'app/site/motions/models/view-motion';
@@ -457,6 +458,7 @@ export class MotionDetailComponent extends BaseModelContextComponent implements 
      * Sets all required subjects and fills in the required information
      */
     public ngOnInit(): void {
+        this.requestUpdates();
         this.registerSubjects();
         this.createForm();
         this.observeRoute();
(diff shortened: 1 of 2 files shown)


client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
```diff
@@ -588,6 +590,64 @@ export class MotionDetailComponent extends BaseModelContextComponent implements
         );
     }
 
+    private requestUpdates(): void {
+        this.requestUserUpdates();
```
FinnStutzenstein, line 594: Please merge those requests together: Each update request will issue a real request to the server. Maybe you can collect follows together and make a single request then.

FinnStutzenstein:
Hi, are these changes due to the search-value-selector?

I have a proposal for it: For displaying, all data is fetched in the main component, e.g. for the category in the motion detail, there is a follow when requesting the motion. So the category can be displayed.

But we do not need to fetch all categories directly. Only when the search-value-selector opens, all categories can be fetched. So next tof injecting an observable into each s-v-selector, we can also add a model request. So the requests are lazy in the sense that all values are only fetched, if the s-v-selector is acutally used.

Whats your opinion about this?

tsiegleauq:
the cleanest solution would probably be to let the `os-search-value-selector` take care of the fetching. This would probably require a rewrite of that component. The `SearchValueSelectorComponent` class cannot trivially extend `BaseModelContextComponent` since it currently extends `BaseFormControlComponent`. Thus, some restructure is required. I.e replacing `BaseModelContextComponent` with an interface (which would have been the cleaner approcach itfp).
Furthermore, I am not entirely sure if the request and an observable would be sufficient or if the whole repository would be required. In any case, it seems to involve a whole lot of changes.

FinnStutzenstein:
Maybe we just should not let the s-v-selector inherit from `BaseModelContextComponent`.

This parameters for the s-v-selector would also be enough:
- The request
- The viewmodel (or collection string, ...) -> We can get the repo by VM/collection via the `CollectionMapperService`
- Optional: A filter function. E.g. you want to exclude the default group, or something else.

Maybe the changes in the s-v-selector are not so big, since it is just an extra request, when it is opened. But it has to be adopted in all places. Nevertheless we can remove the mass of observables e.g. in the motion detail, but have to write dedicated model requests.

Maybe those requests are always the same and can be generated in the repo? -> Something like: Give me the model requests for the list (Which can be used in actual view lists or in the s-v-selectors).

tsiegleauq:
still sounds like a massive change. I'll make according changes for motion detail and the search-value-selector (today) so we can further discuss this approach

FinnStutzenstein:
Ok, thanks!

FinnStutzenstein:
Should this be merged in the meanwhile (when the one remark is fixed)?

tsiegleauq:
no I don't think so. The restructure to the search value selectors would make the remarks obsolete. Since we are not required to have all updates to the motion detail to be working at this very moment (meaning that you could implement the actions already even without this commit) I can make the changes in a second commit in this pr.

FinnStutzenstein:
Ok, I leave it open then

FinnStutzenstein:
Please rework the SV-Selector in another branch.

PR #28, opened 2020-08-11
Autoupdate for amendments

tsiegleauq:
Adjust the new autoupdate system to the amendment list

### DIFF
--- a/client/src/app/core/repositories/motions/motion-repository.service.ts
+++ b/client/src/app/core/repositories/motions/motion-repository.service.ts
@@ -100,7 +100,9 @@ export class MotionRepositoryService extends BaseIsAgendaItemAndListOfSpeakersCo
         const listFields: (keyof Motion)[] = titleFields.concat([
             'sequential_number',
             'sort_weight',
-            'category_weight'
+            'category_weight',
+            'lead_motion_id', // needed for filtering
+            'amendment_ids'
         ]);
         const blockListFields: (keyof Motion)[] = titleFields.concat(['block_id']);
         const detailFields: (keyof Motion)[] = titleFields.concat([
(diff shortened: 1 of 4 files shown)


tsiegleauq:
@FinnStutzenstein this does not really work too well, but might be due to the amendment filter list

client/src/app/core/repositories/motions/motion-repository.service.ts
```diff
@@ -110,7 +110,7 @@ export class MotionRepositoryService extends BaseIsAgendaItemAndListOfSpeakersCo
             'amendment_paragraphs',
             'modified_final_version'
         ]);
-        const amendmentFields: (keyof Motion)[] = ['amendment_paragraphs'];
+        const amendmentFields: (keyof Motion)[] = listFields.concat(['amendment_paragraphs', "lead_motion_id", "amendment_ids"]);
```
FinnStutzenstein, line 113: You need both id fields, so the client knows, which are amendments ;) This was the "biggest" change.

client/src/app/core/repositories/motions/motion-repository.service.ts
```diff
@@ -175,12 +175,29 @@ export class MotionRepositoryService extends BaseIsAgendaItemAndListOfSpeakersCo
...
 
-    protected createViewModelWithTitles(model: Motion): ViewMotion {
+    protected createViewModel(model: Motion): ViewMotion {
         const viewModel = super.createViewModel(model);
 
         viewModel.getNumberOrTitle = () => this.getNumberOrTitle(viewModel);
         viewModel.getProjectorTitle = () => this.getProjectorTitle(viewModel);
 
+        viewModel.getAmendmentParagraphLines = () => {
+            if (viewModel.lead_motion && viewModel.isParagraphBasedAmendment()) {
```
FinnStutzenstein, line 185: This code was originally a "relation": https://github.com/OpenSlides/OpenSlides/blob/master/client/src/app/core/repositories/motions/motion-repository.service.ts#L326

Since I redone the relations I deleted it and as I just saw I never readded the code again. Here it is ;)

FinnStutzenstein:
How do I get to the amendmentlist in the UI? I didn't find any link to it, so I navigated to `motions/amendments` in the url-bar in my browser...

tsiegleauq:
there is a link in the motion list views dot-menu hidden by various conditions.
as an alternative one might use the direct link from the motion details to the corresponding amendments

```html
<div *ngIf="amendmentsEnabled && (perms.isAllowed('manage') || hasAmendments())">
    <button mat-menu-item routerLink="amendments">
        <mat-icon>color_lens</mat-icon>
        <span>{{ 'Amendments' | translate }}</span>
    </button>
</div>
```

FinnStutzenstein:
Ah, I added a commit reenabling the list menu item.

PR #58, opened 2020-10-19
Adds actions for motions

GabrielInTheWorld:
- Requests in the `motion-repository`.
- Also unifies the creating of agenda-items and creating a function to get information about the creating of an agenda-item (`createAgendaItem(model)`).

PR #62, opened 2020-10-21
Fixed motion detail tag autoupdate

FinnStutzenstein:
Only works in conjunction with #58


FinnStutzenstein:
Closed in favor of #58

PR #64, opened 2020-10-26
Adds actions for mediafiles

GabrielInTheWorld:


client/src/app/core/repositories/mediafiles/mediafile-repository.service.ts
```diff
@@ -106,17 +108,42 @@ export class MediafileRepositoryService extends BaseIsListOfSpeakersContentObjec
...
-        // });
-        throw new Error('TODO');
+        const payload = mediafiles.map(mediafile => ({ id: mediafile.id }));
+        return this.sendBulkActionToBackend(MediafileAction.DELETE, payload);
+    }
+
+    public async createFile(partialMediafile: Partial<Mediafile>): Promise<Identifiable> {
+        const payload: MediafileAction.CreateFilePayload = {
+            meeting_id: this.activeMeetingService.meetingId,
+            file: partialMediafile.url,
```
FinnStutzenstein, line 118: This must be the base64 encoded data (different to OS3!). Please do not use `Partial<Mediafile>` as the parameter type, since `file` is not available in `Mediafile`. Use the payload format directly! Also have a look into the `MediaUploadContentComponent`. There is the call to the `createFile` (btw. this should be called `upload`), which may be adjusted.

PR #69, opened 2020-11-06
[WiP] Adds bulk actions for motions, users and so on

GabrielInTheWorld:
Currently implemented:

- Bulk actions for motions
- Bulk actions for users
- Bulk actions for agenda-items
- Bulk actions for assignments

TODO:
- Refactoring of `motion-import.service.ts`
- Maybe splitting user-repo into `user-repository` and `temporary-user-repository`?

### DIFF
--- a/client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
+++ b/client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
@@ -4,7 +4,7 @@ import { AgendaItemAction } from 'app/core/actions/agenda-item-action';
 import { DEFAULT_FIELDSET, Fieldsets } from 'app/core/core-services/model-request-builder.service';
 import { MeetingSettingsService } from 'app/core/ui-services/meeting-settings.service';
 import { TreeIdNode } from 'app/core/ui-services/tree.service';
-import { AgendaItem } from 'app/shared/models/agenda/agenda-item';
+import { AgendaItem, AgendaItemType } from 'app/shared/models/agenda/agenda-item';
 import { Identifiable } from 'app/shared/models/base/identifiable';
 import { HasAgendaItem, ViewAgendaItem } from 'app/site/agenda/models/view-agenda-item';
 import { BaseViewModel } from 'app/site/base/base-view-model';
@@ -177,5 +177,25 @@ export class AgendaItemRepositoryService extends BaseRepositoryWithActiveMeeting
             }
         });
         return duration;
+    }
+
+    public bulkOpenItems(items: ViewAgendaItem[]): Promise<void> {
+        const payload: AgendaItemAction.UpdatePayload[] = items.map(item => ({ closed: false, id: item.id }));
+        return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
+    }
+
+    public bulkCloseItems(items: ViewAgendaItem[]): Promise<void> {
+        const payload: AgendaItemAction.UpdatePayload[] = items.map(item => ({ closed: true, id: item.id }));
+        return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
+    }
+
+    public bulkSetAgendaType(items: ViewAgendaItem[], agendaType: AgendaItemType): Promise<void> {
+        const payload: AgendaItemAction.UpdatePayload[] = items.map(item => ({ id: item.id, type: agendaType }));
+        return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
+    }
+
+    public bulkRemoveItemsFromAgenda(items: ViewAgendaItem[]): Promise<void> {
+        const payload: AgendaItemAction.DeletePayload[] = items.map(item => ({ id: item.id }));
+        return this.sendBulkActionToBackend(AgendaItemAction.DELETE, payload);
     }
 }
--- a/client/src/app/core/repositories/base-repository.ts
+++ b/client/src/app/core/repositories/base-repository.ts
@@ -336,8 +336,9 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
         });
     }
 
-    protected raiseError = (errorMessage: string | Error) => {
-        this.repositoryServiceCollector.errorService.showError(errorMessage);
+    protected raiseError = (error: string | Error) => {
+        this.repositoryServiceCollector.errorService.showError(error);
+        throw error;
     };
 
     protected async sendActionToBackend(action: string, payload: any): Promise<any> {
--- a/client/src/app/core/repositories/topics/topic-repository.service.ts
+++ b/client/src/app/core/repositories/topics/topic-repository.service.ts
@@ -6,6 +6,7 @@ import { DEFAULT_FIELDSET, Fieldsets } from 'app/core/core-services/model-reques
 import { Identifiable } from 'app/shared/models/base/identifiable';
 import { Topic } from 'app/shared/models/topics/topic';
 import { createAgendaItem } from 'app/shared/utils/create-agenda-item';
+import { ViewAgendaItem } from 'app/site/agenda/models/view-agenda-item';
 import { CreateTopic } from 'app/site/topics/models/create-topic';
 import { ViewTopic } from 'app/site/topics/models/view-topic';
 import { BaseIsAgendaItemAndListOfSpeakersContentObjectRepository } from '../base-is-agenda-item-and-list-of-speakers-content-object-repository';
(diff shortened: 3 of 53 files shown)


client/src/app/core/core-services/action.service.ts
```diff
@@ -24,6 +24,7 @@ export class ActionService {
     }
 
     private async _sendRequest<T>(request: RequestInfo): Promise<T> {
+        console.log('sendRequest', request);
```
GabrielInTheWorld, line 27: Will be removed.

client/src/app/core/ui-services/base-filter-list.service.ts
```diff
@@ -454,13 +454,14 @@ export abstract class BaseFilterListService<V extends BaseViewModel> {
...
     /**
-     * Checks if a given ViewBaseModel passes the filter.
+     * Checks if a given ViewBaseModel passes a filter.
      *
      * @param item Usually a view model
      * @param filter The filter to check
      * @returns true if the item is to be displayed according to the filter
      */
     private checkIncluded(item: V, filter: OsFilter): boolean {
+        console.log('checkIncluded', item, filter);
```
GabrielInTheWorld, line 464: Will be removed, too.

FinnStutzenstein:
> Maybe splitting user-repo into user-repository and temporary-user-repository?

No, since repos are bound to a collection and users and temporary users are the same collection. It might be a bit full in this repo due to the manifold actions, but the advantage is, that users in general are equal - there is no difference in normal handling between a temporary and non-temporary user.

client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
```diff
@@ -16,6 +16,8 @@ export interface AgendaListTitle {
     subtitle?: string;
 }
 
+export type AgendaItemType = 1 | 2 | 3;
```
FinnStutzenstein, line 19: Please move it to the `agenda-item.ts` and maybe use it in `ItemVisibilityChoices` as well as change the type of `number` in `AgendaItem`

client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
```diff
@@ -178,4 +180,24 @@ export class AgendaItemRepositoryService extends BaseRepositoryWithActiveMeeting
...
+        return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
+    }
+
+    public bulkSetAgendaType(items: ViewAgendaItem[], agendaType: AgendaItemType): Promise<void> {
+        const payload: AgendaItemAction.UpdatePayload[] = items.map(item => ({ id: item.id, type: agendaType }));
+        return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
+    }
+
+    public bulkRemoveItemsFromAgenda(items: ViewAgendaItem[]): Promise<void> {
+        const payload: AgendaItemAction.DeletePayload[] = items.map(item => ({ id: item.id }));
```
FinnStutzenstein, line 200: Just a notice. YOu often did not include the `DeletePayload` since it is mostly `Identifiable`. Maybe this is a good reason to include it. If you have time, please add the `DeleteAction` to each model.

client/src/app/core/repositories/base-repository.ts
```diff
@@ -336,8 +336,9 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
         });
     }
 
-    protected raiseError = (errorMessage: string | Error) => {
-        this.repositoryServiceCollector.errorService.showError(errorMessage);
+    protected raiseError = (error: Error) => {
+        this.repositoryServiceCollector.errorService.showError(error);
+        throw error;
```
FinnStutzenstein, line 341: Please do not rethrow here. If you rebase, the send*ActionToBackend Methods are adpoted to intercept, rather to catch, the error: https://github.com/OpenSlides/openslides-client/blob/master/client/src/app/core/repositories/base-repository.ts#L343

client/src/app/core/repositories/topics/topic-repository.service.ts
```diff
@@ -91,14 +85,35 @@ export class TopicRepositoryService extends BaseIsAgendaItemAndListOfSpeakersCon
...
+    public duplicateMultipleTopics(agendaItems: ViewAgendaItem[]): Promise<void> {
+        const payload: TopicAction.CreatePayload[] = agendaItems.map(item =>
+            this.getCreatePayload(this.getDuplicatedTopic(item))
         );
+        return this.sendBulkActionToBackend(TopicAction.CREATE, payload);
+    }
+
+    private getDuplicatedTopic(topicAgendaItem: ViewAgendaItem): CreateTopic {
+        const viewTopic = topicAgendaItem.content_object as ViewTopic;
+        return new CreateTopic({
```
FinnStutzenstein, line 101: Still not a fan of create models. Please use the createAction interface for this. Maybe you are able to completely remove `CreateTopic` (it may be used during importing...).

client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
```diff
@@ -42,6 +45,10 @@ import { hasListOfSpeakers } from '../../models/view-list-of-speakers';
     styleUrls: ['./agenda-item-list.component.scss']
 })
 export class AgendaItemListComponent extends BaseListViewComponent<ViewAgendaItem> implements OnInit {
+    public readonly AGENDA_TYPE_PUBLIC = 1;
```
FinnStutzenstein, line 48: please also move to agenda_item.ts and unify the usage of these constants.

client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
```diff
@@ -275,30 +282,32 @@ export class AgendaItemListComponent extends BaseListViewComponent<ViewAgendaIte
...
      */
-    public async setClosedSelected(closed: boolean): Promise<void> {
+    public async openSelectedItems(): Promise<void> {
         try {
-            for (const item of this.selectedRows) {
-                await this.repo.update({ closed: closed }, item);
-            }
+            await this.repo.bulkOpenItems(this.selectedRows);
+        } catch (e) {
+            this.raiseError(e);
```
FinnStutzenstein, line 300: should not be necessary anymore since it is intercepted in the repos, right?

FinnStutzenstein: Please also check other places where `raiseError` is used.

client/src/app/site/motions/modules/motion-workflow/components/workflow-list/workflow-list.component.ts
```diff
@@ -80,8 +80,7 @@ export class WorkflowListComponent extends BaseListViewComponent<ViewMotionWorkf
                 {
                     idField: 'motion_workflow_ids'
                 }
-            ],
-            fieldset: []
+            ]
```
FinnStutzenstein, line 83: Please undo these changes - the `fieldset: []` clears required fields for the ViewMeeting. If it is not gives, the DEFAULT_FIELDSET will be used, which will currently fetch the `description` of the meeting, but it is not needed here.

client/src/app/site/motions/services/motion-multiselect.service.ts
```diff
@@ -24,6 +31,14 @@ import { ViewMotion } from '../models/view-motion';
     providedIn: 'root'
 })
 export class MotionMultiselectService {
+    private get activeMeetingId(): ActiveMeetingIdService {
```
FinnStutzenstein, line 34: Make and resutn type does not fit. If the serviceCollector has the `activeMeetingIdService`, than there is a missing getter in one of the base classes (the base class, which required the service collector). Please add it there. Same for the action service.

client/src/app/site/motions/services/motion-multiselect.service.ts
```diff
@@ -126,38 +130,27 @@ export class MotionMultiselectService {
...
-        } else {
-            throw new Error(
-                this.translate.instant('You cannot change the recommendation of motions in different workflows!')
+        // hacks custom Displayables from recommendations
+        // TODO: Recommendations should be an own class
+        const title = this.translate.instant('This will set the following recommendation for all selected motions:');
+        const choices: Displayable[] = this.workflowRepo
+            .getWorkflowStatesForMotions(motions)
+            .filter(workflowState => !!workflowState.recommendation_label)
+            .map(workflowState => ({
```
FinnStutzenstein, line 139: Why do you map this, it should not be needed? Recommendation is the name for a state with a recommendation_label; it is not a different model. (Same as the wording `amendment` for a motion with a lead motion assigned). If there are problems, please give me a call or so, so we can speak about it.

FinnStutzenstein (changes requested):

client/src/app/core/repositories/topics/topic-repository.service.ts
GabrielInTheWorld: Yes, I'm not, too. There are many changes, that are not pushed yet. Those "*-create-*"-files will be completely deleted. :)

FinnStutzenstein: Yeah!

PR #74, opened 2020-12-17
Adds a new resource component to get resources with access-token

GabrielInTheWorld:
- Adds a new fullscreen download component to display resources in fullscreen on a new page.
- Also moves the fullscreen-projector to a new directory "fullscreen".

Fixes [openslides-auth-service#60](https://github.com/OpenSlides/openslides-auth-service/issues/60)
Fixes [openslides-media-service#11](https://github.com/OpenSlides/openslides-media-service/issues/11)


client/src/app/fullscreen/download/download-routing.module.ts
```diff
@@ -0,0 +1,17 @@
+import { NgModule } from '@angular/core';
+import { RouterModule, Routes } from '@angular/router';
+
+import { DownloadComponent } from './modules/download/download.component';
+
+const routes: Routes = [
+    {
+        path: 'system/media/get/:id',
```
FinnStutzenstein, line 8: I wonder why this works. DO not overwrite system urls. E.g. If I send you a link to a resource and you click on the link, the Haproxy will do not even load the client, since the routing redirects it to the media service.

What are you current urls for the fullscreen projector and the resource downloads? Maybe we can find a nicer download url

GabrielInTheWorld: It works because the full url is `downloads/system/media/get/:id`, I think.

GabrielInTheWorld: Current urls for
- projector: `projector/...`
- resource downloads: `download/...`

FinnStutzenstein: ah ok, but if it is client controlled, why not just `download/:id`?

GabrielInTheWorld: Because I then can add the whole url of a mediafile to the path. But, I adapted it, now. You can review it.

PR #80, opened 2021-01-08
Hotfix: Updates enum-values

GabrielInTheWorld:


client/src/app/core/repositories/agenda/agenda-item-visibility.ts
```diff
@@ -0,0 +1,5 @@
+export enum AgendaItemVisibility {
```
FinnStutzenstein, line 1: Please move the enums into the model files.

client/src/app/site/agenda/components/agenda-sort/agenda-sort.component.ts
```diff
@@ -155,4 +156,15 @@ export class AgendaSortComponent extends BaseSortTreeComponent<ViewAgendaItem> i
                 return 'visibility_off';
         }
     }
+
+    private getNumericValueForAgendaItemVisibility(visibility: AgendaItemVisibility): number {
```
FinnStutzenstein, line 160: THis seems hacks. Can you adopt the filter to not use numeric values but strings?

GabrielInTheWorld: This would be some more effort. Therefore, I create a new issue for that.

PR #84, opened 2021-01-11
Hotfix: logout of user

GabrielInTheWorld:


client/src/app/core/core-services/auth-guard.service.ts
```diff
@@ -39,15 +40,19 @@ export class AuthGuard implements CanActivate, CanActivateChild {
      *
      * @param route the route the user wants to navigate to
      */
-    public canActivate(route: ActivatedRouteSnapshot): boolean {
+    public canActivate(
```
FinnStutzenstein, line 43: please make the function async und use await down below!

client/src/app/core/core-services/auth-guard.service.ts
```diff
@@ -39,15 +40,19 @@ export class AuthGuard implements CanActivate, CanActivateChild {
...
+    ): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
         const basePerm: Permission | Permission[] = route.data.basePerm;
 
-        if ((this.operator.isAnonymous && this.activeMeeting.guestsEnabled) || this.operator.isAuthenticated) {
-            return this.hasPerms(basePerm);
-        } else {
-            this.router.navigate(['login']);
-            return false;
-        }
+        return this.authService.firstTimeWhoAmI.toPromise().then(() => {
```
FinnStutzenstein, line 48: The OS3 code had the authguard coupled to the operator. There is an operator-loaded event/observable. This is needed so one could query permissions. Please change the waiting to the operator, since you have to wait for it to query perms. It is not clear that waiting for the `firstTimeWhoAmI` solve the issue.

client/src/app/core/core-services/streaming-communication.service.ts
```diff
@@ -126,7 +126,7 @@ export class StreamingCommunicationService {
 
     private close(streamContainer: StreamContainer): void {
         if (this.isStreamOpen(streamContainer)) {
-            this.streams[streamContainer.id].stream.close();
+            this.streams[streamContainer.id].stream?.close();
```
FinnStutzenstein, line 129: Please undo this. Did it raises you errors?

FinnStutzenstein (commented):
Please make another PR and wait for reviews.

PR #102, opened 2021-02-04
Fix list of speakers

### LINKED ISSUE #94: Fix list of speakers
Due to the rebase and massive changes in OS3 (new ListOfSpeakersContentComponent) the LOS does not work. I suspect that the repo etc. is ok and only the components must be adapted and also use the new autoupdate model requests etc.

GabrielInTheWorld:
Fixes #94 

Waits for [openslides-backend#352](https://github.com/OpenSlides/openslides-backend/issues/352)

### DIFF
--- a/client/src/app/core/actions/speaker-action.ts
+++ b/client/src/app/core/actions/speaker-action.ts
@@ -13,6 +13,7 @@ export namespace SpeakerAction {
         list_of_speakers_id: Id;
         user_id: Id;
         marked?: boolean;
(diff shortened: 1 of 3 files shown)


emanuelschuetze:
Tested with current OS4 master: Get error message after adding new users to the los: "User 2 is already on the list of speakers."

emanuelschuetze (approved):

FinnStutzenstein (approved):

Issue #92, opened 2021-01-26
Bulk actions


FinnStutzenstein:
Finish #69

GabrielInTheWorld:
Currently implemented:

Bulk actions for motions
Bulk actions for users
Bulk actions for agenda-items
Bulk actions for assignments
TODO:

Refactoring of motion-import.service.ts

PR #106, opened 2021-02-05
Updates users to match new models.yml

### LINKED ISSUE #85: Adapt changes in the models.yml
From https://github.com/OpenSlides/OpenSlides/pull/5802:

- delete role model and repo
- Add `organisation_management_level` to user
- Change `about_me`, `comment`, `number`, `structure_level`, `vote_weight` to template fields. This requires to add methods to `user.ts` and `view-user.ts`.
- Add fields `default_number` and `default_structure_level`. In the getters of `number` and `structure_level` fall back to the defaults, if there are no specific numbers/structure levels
- watch out for `default_vote_weight`, it may be added in the PR
- adapt all templates and autoupdate queries in the user repo.

### LINKED ISSUE #104: Editing of meta-infos about users is broken (in the users-list)
Distinguish between temporary and non-temporary users...

### LINKED ISSUE #105: Set present of a user
Currently, I cannot set a user to present/non-present

GabrielInTheWorld:
Fixes #85 
Fixes #104 
Fixes #105


emanuelschuetze:
my tests (create new user) failed. As discussed with Joshua we wait for changes in os-backend.

client/src/app/core/repositories/relations.ts
```diff
@@ -126,15 +125,15 @@ export const RELATIONS: Relation[] = [
     }),
     ...makeM2O({
         OViewModel: ViewOrganisation,
-        MViewModel: ViewRole,
+        MViewModel: ViewUser,
         OField: 'roles',
-        MField: 'organisation'
+        MField: 'organisation_management_level'
```
FinnStutzenstein, line 130: This is not a relation - the `user/organisation_management_level` is a string describing his/her role (e.g. the value is `can_manage_organisation`, any of https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/models.yml#L114.

so if `user/organisation_management_level === 'superadmin'` the user is superadmin.

FinnStutzenstein: ...so just remove these relations

client/src/app/site/users/models/view-user.ts
```diff
@@ -150,6 +149,26 @@ export class ViewUser extends BaseProjectableViewModel<User> implements Searchab
     public vote_delegations_from_ids(meetingId?: Id): Id[] {
         return this.user.vote_delegations_from_ids(meetingId || this.getEnsuredActiveMeetingId());
     }
+
+    public vote_weight(meetingId?: Id): number {
+        return this.user.vote_weight(meetingId || this.getEnsuredActiveMeetingId());
+    }
+
+    public number(meetingId?: Id): string {
+        return this.user.number(meetingId || this.getEnsuredActiveMeetingId());
```
FinnStutzenstein, line 158: I think you may not have to use `this.getEnsuredActiveMeetingId` here (it throws an error if I remember correctly) -> If there is not active meeting, you can fall back to the default.

client/src/app/site/users/services/user-import.service.ts
```diff
@@ -36,8 +36,8 @@ export class UserImportService extends BaseImportService<User> {
         'default_password',
         'email',
         'username',
-        'gender',
-        'vote_weight'
```
FinnStutzenstein, line 40: you can query the default values. Adding the template fields  is a todo, right?

GabrielInTheWorld: I didn't change it there, because of #92. The views to import users, motions, etc. will be updated.

emanuelschuetze (approved):

PR #133, opened 2021-03-23
Adds downloads to files and a service for organisation-things

### LINKED ISSUE #90: Download component
I early-merged https://github.com/OpenSlides/openslides-client/pull/74 to be able to rebase the client. @GabrielInTheWorld We talked about this PR. It still needs an overhaul to just download the resource to the clients device instead of displaying it. Also the name `resource` is used twice: Just use it for the model, not the component.

### LINKED ISSUE #126: Implement organisation-settings.service
Implement organisation-settings.service

_Originally posted by @GabrielInTheWorld in https://github.com/OpenSlides/openslides-client/pull/114#r589346959_

GabrielInTheWorld:
Fixes #90 
Fixes #126 

@FinnStutzenstein: I implemented a `resource.service`. You meant, there has to be a function to get a token or should there a function to pass a token and get a resource?
Please give me a feedback.
@FinnStutzenstein, another thing: You implemented a function to print current open streams. I packed them into a map to avoid a huge amount of `console.log`s only to printing streams. How do you find it?


client/src/app/core/translate/translation-parser.ts
```diff
@@ -25,26 +25,26 @@ export class OpenSlidesTranslateParser extends TranslateDefaultParser {
      * Subscribes to the custom translations and watches for updated custom translations.
      */
     public constructor(
-        organisationSettingsService: OrganisationSettingsService,
+        // organisationSettingsService: OrganisationSettingsService, // TODO: Cyclic dependency...
```
FinnStutzenstein, line 28: Does the orga service needs translations? Or where is the cycle?

If it is not easy solvable, create a custom-translation-service just providing a public behaviour subject -> the orga service must actively push in it and this service can read.

client/src/app/core/ui-services/file-export.service.ts
```diff
@@ -23,7 +23,7 @@ export class FileExportService {
             options.type = mimeType;
         }
         const blob = new Blob([file], options);
-        saveAs(blob, filename, { autoBOM: true });
+        saveAs(blob, filename, { autoBom: true });
         // autoBOM = automatic byte-order-mark
```
FinnStutzenstein, line 27: adjust comment

client/src/app/core/ui-services/login-data.service.ts
```diff
@@ -155,32 +155,33 @@ export class LoginDataService {
      */
     private async setup(): Promise<void> {
```
FinnStutzenstein, line 156: I think this service is not needed anymore. It was introduced to OS3 because on the login mask (or for every not-logged in user and no guests allowed), you had no access to the autoupdate, so you could not get the data needed for the login page. We solved this by adding a get-route in the server returning only the login data needed. This services unified the two sources of login data: One source is the get request, but when you are logged in the autoupdate is the second source of data. Also the data was stored in the local storage, so the following startups do not take so long (this is actually a third source used for initializing this service;) )

Now, the autoupdate for the orga and resource is always open, also for not authenticated users. Especially all "login data" is now available through the autoupdate. Also, for the first try, we should not cache this data to make the lifes in the client easier.

I propose just deleting this service and directly use the orgasettingsservice where these pieces of data are needed.

client/src/app/core/ui-services/login-data.service.ts
```diff
@@ -155,32 +155,33 @@ export class LoginDataService {
...
         });
-        this.organisationSettingsService
-            .get<{ path: string; display_name: string }>('logo_web_header')
-            .subscribe(value => {
-                if (value) {
-                    this._logoWebHeader.next(value);
-                    this.storeLoginDataRequests.next();
-                }
-            });
+        // TODO!
```
FinnStutzenstein, line 176: this is a resource now. The token is client-defined, so maybe create a constant somewhere.

client/src/app/fullscreen/download/modules/download/download.component.ts
```diff
@@ -9,30 +12,29 @@ import { OperatorService } from 'app/core/core-services/operator.service';
...
+            const resourceUrl = `/system/media/get/${queryParams.id}`;
+            this.http.get<Blob>(resourceUrl, null, null, null, 'blob').then(response => {
+                this.readBlobData(response, queryParams.name);
+            });
         }
     }
+
+    private readBlobData(data: Blob, fileName: string): void {
+        saveAs(data, fileName);
+        setTimeout(() => window.close(), 100); // file-saver works not synchronously. So, wait some ms.
```
FinnStutzenstein, line 38: is 100 ms good enough?

client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
```diff
@@ -235,7 +235,7 @@ export class AssignmentDetailComponent extends BaseModelContextComponent impleme
         this.tagsObserver = this.tagRepo.getViewModelListBehaviorSubject();
         this.mediafilesObserver = this.mediafileRepo.getViewModelListBehaviorSubject();
 
-        // TODO: why are groups needed here??
+        // Groups are needed, to select entitled groups.
```
FinnStutzenstein, line 238: doesn't they belong into the poll form then?

client/src/app/site/polls/services/poll.service.ts
```diff
@@ -210,10 +206,9 @@ export abstract class PollService {
         protected pollKeyVerbose: PollKeyVerbosePipe,
         protected parsePollNumber: ParsePollNumberPipe
     ) {
-        // organisationSettingsService
-        //     .get<OpenSlidesSettings>('Settings')
-        //     .subscribe(settings => (this.isElectronicVotingEnabled = settings.ENABLE_ELECTRONIC_VOTING));
-        this.isElectronicVotingEnabled = true;
+        _organisationSettingsService
```
FinnStutzenstein, line 209: please remove the underscore

FinnStutzenstein (changes requested):
Looks good!

FinnStutzenstein:
> should there a function to pass a token and get a resource?

This is right.

client/src/app/fullscreen/download/modules/download/download.component.ts
GabrielInTheWorld: I think so. I have tested it with firefox and chrome and I get a download every time I used 100 ms.

client/src/app/core/ui-services/login-data.service.ts
GabrielInTheWorld: Okay

client/src/app/site/assignments/components/assignment-detail/assignment-detail.component.ts
GabrielInTheWorld: Yes :D

FinnStutzenstein (approved):

PR #137, opened 2021-04-01
Shows an error-message to a user

### LINKED ISSUE #135: wrong-password message is missing
There must be a message that the username/password is not correct on the login mask. Also there must be a message if it is correct, but the user is inactive. Both messages must be generated by the auth-service and displayed in the client. The relevant strings are here: https://github.com/OpenSlides/OpenSlides/blob/master/server/openslides/users/views.py#L895

GabrielInTheWorld:
Fixes #135 

Waits for [openslides-auth-service#69](https://github.com/OpenSlides/openslides-auth-service/pull/69)...

### DIFF
--- a/client/src/app/site/login/components/login-mask/login-mask.component.ts
+++ b/client/src/app/site/login/components/login-mask/login-mask.component.ts
@@ -171,9 +171,7 @@ export class LoginMaskComponent extends BaseComponent implements OnInit, OnDestr
             await this.authService.login(this.loginForm.value.username, this.loginForm.value.password);
         } catch (e) {
             // this.overlayService.hideSpinner();
-            this.loginForm.setErrors({
-                notFound: true
-            });
+            this.loginForm.get('password').setErrors({ notFound: true });
             this.loginErrorMsg = e;
         }
         // throw new Error('TODO'); // Ingore SAML for now...
--- a/client/src/app/core/core-services/http.service.ts
+++ b/client/src/app/core/core-services/http.service.ts
@@ -122,7 +122,7 @@ export class HttpService {
         let error = this.translate.instant('Error') + ': ';
         // If the error is a string already, return it.
         if (typeof e === 'string') {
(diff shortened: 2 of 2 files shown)


client/src/app/site/login/components/login-mask/login-mask.component.ts
```diff
@@ -173,7 +173,7 @@ export class LoginMaskComponent extends BaseComponent implements OnInit, OnDestr
             this.loginForm.setErrors({
                 notFound: true
             });
-            this.loginErrorMsg = e;
+            this.matSnackBar.open(e, 'Ok');
```
FinnStutzenstein, line 176: this is weird. why do you open the snackbar? In OS3 no such snackbar opens, rather the error is shown below the password field.

GabrielInTheWorld: Well, it is strange, that there is no error message shown. The code is still there...
That is, why I open a snackbar...

FinnStutzenstein: Please find out, why nothing is shown there

GabrielInTheWorld: Done.

client/src/app/site/login/components/login-mask/login-mask.component.ts
```diff
@@ -171,9 +171,7 @@ export class LoginMaskComponent extends BaseComponent implements OnInit, OnDestr
             await this.authService.login(this.loginForm.value.username, this.loginForm.value.password);
         } catch (e) {
             // this.overlayService.hideSpinner();
-            this.loginForm.setErrors({
-                notFound: true
-            });
+            this.loginForm.get('password').setErrors({ notFound: true });
```
FinnStutzenstein, line 174: Strange, I wonder why it works in OS3 :D

PR #153, opened 2021-04-15
Adds a dialog to manage orga-wide my account

### LINKED ISSUE #149: Orga:"My account"
Make a page/dialog with "my account"-related content.

GabrielInTheWorld:
Fixes #149 

- Adds a user-dialog to manage orga-wide user-related things for my account. Currently, I can only change my password.
- Separates the form to change a password into an own component to reuse it.

### DIFF
--- a/client/src/app/core/core-services/route.guard.ts
+++ b/client/src/app/core/core-services/route.guard.ts
@@ -22,7 +22,7 @@ export class RouteGuard {
         this.router.events.subscribe(event => {
             if (event instanceof NavigationStart) {
                 const paths = event.url.split('/').filter(path => !!path);
-                if (!!Number(paths[0]) && !this.excludedRoutes.includes(paths[0])) {
+                if (!!paths[0] && !Number(paths[0]) && !this.excludedRoutes.includes(paths[0])) {
                     console.warn(`Route "${event.url}" is misleading. Please change.`);
                     this.router.navigate([this.activeMeetingId, ...paths]);
                 }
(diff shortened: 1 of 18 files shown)


client/src/app/core/core-services/route.guard.ts
```diff
@@ -22,7 +22,7 @@ export class RouteGuard {
         this.router.events.subscribe(event => {
             if (event instanceof NavigationStart) {
                 const paths = event.url.split('/').filter(path => !!path);
-                if (!!Number(paths[0]) && !this.excludedRoutes.includes(paths[0])) {
+                if (!!paths[0] && !Number(paths[0]) && !this.excludedRoutes.includes(paths[0])) {
```
GabrielInTheWorld, line 25: I had to change it to this, to catch things like `/1` or `/users`.

client/src/app/management/components/account-dialog/account-dialog.component.html
```diff
@@ -0,0 +1,46 @@
+<h2 mat-dialog-title translate>My account</h2>
```
tsiegleauq, line 1: translate

tsiegleauq (changes requested):

client/src/app/management/components/account-dialog/account-dialog.component.html
tsiegleauq: also "Show profile" was already known

PR #176, opened 2021-04-16
Separates fetching all users for one orga into one service

### LINKED ISSUE #175: Fetch orga-members in every place, they are needed
For example:
- member-list
- committee-edit
- committee-list
- meeting-edit

GabrielInTheWorld:
Fixes #175 

- One service fetch a whole list of all users of one orga.
- Components, which need a list of all users of one orga, fetch them, now.

I don't know, if the naming for "members" is correct...

### DIFF
--- a/client/src/app/management/components/meeting-edit/meeting-edit.component.html
+++ b/client/src/app/management/components/meeting-edit/meeting-edit.component.html
@@ -59,7 +59,7 @@
                 formControlName="guest_ids"
                 [multiple]="true"
                 placeholder="{{ 'Participants' | translate }}"
-                [inputListValues]="members"
+                [inputListValues]="allUsers"
             ></os-search-value-selector>
         </mat-form-field>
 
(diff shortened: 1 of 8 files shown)


client/src/app/management/components/meeting-edit/meeting-edit.component.html
```diff
@@ -59,7 +59,7 @@ <h2 *ngIf="!isCreateView">{{ editMeetingLabel | translate }}</h2>
                 formControlName="guest_ids"
                 [multiple]="true"
                 placeholder="{{ 'Participants' | translate }}"
-                [inputListValues]="members"
+                [inputListValues]="allUsers"
```
tsiegleauq, line 62: Need to be fixed or marked as TODO:
The server does only accept members of the current committee to be set here. At least this is the case during creating, during update the server (wrongly) accepts any ID here.

Anyhow, this list should only contain the members of our committee

client/src/app/management/components/meeting-edit/meeting-edit.component.ts
```diff
@@ -125,29 +129,25 @@ export class MeetingEditComponent extends BaseModelContextComponent implements O
         this.meetingForm.patchValue(patchMeeting);
     }
 
-    private requestUpdates(): void {
+    private async requestUpdates(): Promise<void> {
         /**
-         * TODO: Requires orga members
+         * Requires orga members
          */
+        const userIds = await this.memberService.fetchAllOrgaUsers();
```
tsiegleauq, line 136: as above, memberService prolly needs a `fetchMembersOfCommittee(id)`

tsiegleauq (changes requested):

PR #185, opened 2021-04-19
Fix user update

### LINKED ISSUE #144: Error updating user
When updating a user, the following error occurs: 

`data.vote_delegations_$_from_ids.1 must be array`

The reason is probably that the client gives the following action data:
```
vote_delegations_$_from_ids: {1: {1: []}}
```
Note the additional nested object which should not be there.

Tested on the master branch of the main repo, so just close this if it's already outdated.

GabrielInTheWorld:
Fixes #144 

Depends on #184 

- Makes the building of a vote_delegation_from_ids-object correct
- Receives all information in detail-view for a user

### DIFF
--- a/client/src/app/site/users/components/user-detail/user-detail.component.ts
+++ b/client/src/app/site/users/components/user-detail/user-detail.component.ts
@@ -9,6 +9,7 @@ import { ModelSubscription } from 'app/core/core-services/autoupdate.service';
 import { SpecificStructuredField } from 'app/core/core-services/model-request-builder.service';
 import { OperatorService } from 'app/core/core-services/operator.service';
 import { Permission } from 'app/core/core-services/permission';
+import { Id } from 'app/core/definitions/key-types';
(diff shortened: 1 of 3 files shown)


client/src/app/site/users/components/user-detail/user-detail.component.ts
```diff
@@ -138,18 +141,22 @@ export class UserDetailComponent extends BaseModelContextComponent implements On
             this.newUser = true;
             this.setEditMode(true);
         } else {
-            this.route.params.subscribe(params => {
-                this.loadUserById(Number(params.id));
-            });
+            this.subscriptions.push(
+                this.route.params.subscribe(params => {
+                    this.userId = Number(params.id);
```
FinnStutzenstein, line 146: `Number` is different from `number` -> use `parseInt()` or just `+params.id` (my fav;) )

FinnStutzenstein (approved):

PR #190, opened 2021-04-21
Meeting-specific login -> guests can sign in

### LINKED ISSUE #187: Meeting-specific login-mask

GabrielInTheWorld:
Fixes #187 

- Guests can sign in depending on choosed meeting (specified by its id in the URL)
- Reloading a page as guest is enabled

Some permissions are not loaded after reloading a page...

These changes waits for [openslide-autoupdate-service#214](https://github.com/OpenSlides/openslides-autoupdate-service/issues/214) and [openslides-autoupdate-service#215](https://github.com/OpenSlides/openslides-autoupdate-service/issues/215), because there are some missing fields from autoupdates for anonymous.


client/src/app/core/core-services/active-meeting-id.service.ts
```diff
@@ -32,7 +32,7 @@ export class ActiveMeetingIdService {
             this.currentMeetingId = newMeetingId;
             this.meetingIdSubject.next(newMeetingId);
         }
-        if (Number.isNaN || (newMeetingId && newMeetingId !== this.currentMeetingId)) {
+        if (Number.isNaN(newMeetingId) || (newMeetingId && newMeetingId !== this.currentMeetingId)) {
```
FinnStutzenstein, line 35: It would be nice, if `null` was used instead of NaN. Can you find the place where NaN is generated (I suspect a cast from string to integer). Check *there* if the cast was sucessfull and otherwise return null.

client/src/app/core/core-services/active-meeting-id.service.ts
```diff
@@ -50,7 +50,26 @@ export class ActiveMeetingIdService {
         this.lifecycleService.openslidesShutdowned.subscribe(() => this.shutDown());
     }
 
-    public start(): void {
+    private start(): void {
```
FinnStutzenstein, line 53: move start and shutdown directly to the constructor as lambdas, if they are not used somewhere else.

client/src/app/core/core-services/active-meeting-id.service.ts
```diff
@@ -50,7 +50,26 @@ export class ActiveMeetingIdService {
...
+        this.currentMeetingId = null;
+        this.meetingIdSubject.next(undefined);
+        this.routerSub?.unsubscribe();
+    }
+
+    private initialize(): void {
+        const root = this.router.routerState.snapshot.root;
+        if (root.firstChild) {
+            const params = this.router.routerState.snapshot.root.firstChild.params;
+            this._meetingId = Number(params.meetingId);
```
FinnStutzenstein, line 68: this might be the origin of NaN - check it here and use null instead.

client/src/app/core/core-services/active-meeting-id.service.ts
```diff
@@ -50,7 +50,26 @@ export class ActiveMeetingIdService {
...
 
-    public start(): void {
+    private start(): void {
+        this.initialize();
+        this.initSubscription();
+    }
+
+    private shutDown(): void {
+        this.currentMeetingId = null;
+        this.meetingIdSubject.next(undefined);
```
FinnStutzenstein, line 60: null instead of undefined - this is consistent. And `null` indicates, that the service actively knows, that there is no meeting. Undefined would be like "not yet initialized".

client/src/app/core/core-services/active-meeting.service.ts
```diff
@@ -105,4 +113,19 @@ export class ActiveMeetingService {
             ]
         };
     }
+
+    private cleanModelSubscription(): void {
+        this.hasModelSubscriptionInitiated = false;
```
FinnStutzenstein, line 118: this is somewhat implicit. First, it is more likely a `shouldModelSubscribtionBeInitialized` since you set it to false if something chages. Seconds, it is a sideeffect of this methid - move this line up to the handler of the meeting id subscription, so it is clear, when it is again free to initialize

FinnStutzenstein (approved):
Code looks good, not tested

client/src/app/core/core-services/active-meeting-id.service.ts
GabrielInTheWorld: Why? This makes the functions smaller than if everything is in the constructor...

FinnStutzenstein: but they are only used there

tsiegleauq (changes requested):
Admin does not get apps inside a meeting. Only haopens in this branch, not on master tho.

image

FinnStutzenstein:
@GabrielInTheWorld Please rebase on the current master and https://github.com/OpenSlides/openslides-client/pull/218

The change introduced in https://github.com/OpenSlides/openslides-client/pull/218 should also lead to the meeting-specific login mask.

client/src/app/core/core-services/operator.service.ts
```diff
@@ -345,10 +349,15 @@ export class OperatorService {
      * This behaviour prevents a non-redirect on the startpage.
      */
     private navigateOnLogout(): void {
+        // First, check if a user is at the dashboard
```
FinnStutzenstein, line 352: not only the dashboard, but also the committee list etc. Just make this depending on the active meeting id - if there is none, redirect to `/login`, else to the meeting specific one. With the recent changes in #209 the meeting id should be correct

FinnStutzenstein:
With https://github.com/OpenSlides/openslides-permission-service/pull/183 you should be able to display `meeting/name` in the login mask.

FinnStutzenstein:
- steps:
  1) login and navigate to meeting 1
  2) logout -> redirect to the meeting specific login mask. The name is shown (correct), but in the console I only see 1 open stream for the orga service -> the active meeting service did not start, so when changing the name I do not get an autoupdate
  3) reload -> No name is shown. Two streams are opened (correct) and the `meeting/name` is also send from the server, but not shown in the login mask
- Global login mask: Please show the name of the organization
- Steps:
    1) Go to dashboard, reload
    2) navigate to committees->meetings->meeting 1. Look into the console: 5 streams requested
    3) reload: Now 6 streams are requested with two streams for the operator. This is not correct, the operator must only have 1 active stream

client/src/app/core/core-services/active-meeting.service.ts
```diff
@@ -40,16 +42,43 @@ export class ActiveMeetingService {
...
                 this.setupModelSubscription(id);
             }
         });
+        this.lifecycle.openslidesBooted.subscribe(() => {
+            // Workaround to not fire if this service is initialized.
+            if (!this.isOnStartup) {
+                this.setupModelSubscription(this.meetingId);
+            } else {
+                this.isOnStartup = false;
+            }
```
GabrielInTheWorld, line 67: @FinnStutzenstein Please check again.
There is one remaining thing: I have to write this code to prevent, that the active-meeting's model-subscription is re-opened on startup. If that model-subscription is re-opened on startup, the autoupdate-service does not send any meeting-models.
I don't know if this is a bug (on client-side or on autoupdate-side) - what do you think? Do you know, what I mean?

FinnStutzenstein (commented):
I'll add a commit

PR #201, opened 2021-04-27
Adds organization tags client-side

### LINKED ISSUE #192: Organisation-tag actions
The actions for organisation-tags can be implemented.
See [openslides-backend#675](https://github.com/OpenSlides/openslides-backend/pull/675) for further information.

GabrielInTheWorld:
Fixes #192 

- Implements orga-tag-actions and (view-) models
- Implements a view to manage all orga-tags
- Adds form-fields to add or remove orga-tags to committees or meetings

Depends [openslides-backend#639](https://github.com/OpenSlides/openslides-backend/issues/639)

### DIFF
--- a/client/src/app/management/components/meeting-list/meeting-list.component.html
+++ b/client/src/app/management/components/meeting-list/meeting-list.component.html
@@ -58,6 +58,14 @@
         </div>
     </div>
 
+    <div *pblNgridCellDef="'organisation_tags'; row as meeting" class="cell-slot fill">
+        <os-chip-list *ngIf="meeting.organisation_tags?.length" [model]="meeting.organisation_tags">
+            <ng-template let-tag>
+                {{ tag.name }}
+            </ng-template>
+        </os-chip-list>
+    </div>
+
     <div *pblNgridCellDef="'users'; row as meeting" class="cell-slot fill">
         <a class="detail-link" [routerLink]="meeting.getUrl()" *ngIf="!isMultiSelect"></a>
         <div>
--- a/client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.html
+++ b/client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.html
@@ -1,0 +1,38 @@
+<h2 mat-dialog-title>{{ (isCreateView ? 'Create an organization tag' : 'Edit an organization tag') | translate }}</h2>
+<div mat-dialog-content class="organization-tag-content">
+    <form [formGroup]="organisationTagForm">
+        <mat-form-field appearance="fill">
+            <mat-label>{{ 'Name' | translate }}</mat-label>
+            <input #name matInput formControlName="name" required />
+        </mat-form-field>
+        <mat-form-field appearance="fill">
+            <mat-label>{{ 'Color' | translate }}</mat-label>
+            <input
+                matInput
+                formControlName="color"
+                [required]="isCreateView"
+                [osOnlyNumber]="true"
+                [osOnlyNumberAllowHex]="true"
+                [osOnlyNumberAllowLeadingZero]="true"
+                maxlength="6"
+            />
+            <span matPrefix>#&nbsp;</span>
+            <button
+                mat-icon-button
+                matSuffix
+                (click)="generateColor()"
+                matTooltip="{{ 'Generate a new color' | translate }}"
+            >
+                <mat-icon> sync </mat-icon>
+            </button>
+            <mat-error *ngIf="organisationTagForm.get('color').errors?.pattern">
+                {{ 'You have to enter six hexadecimal digits' | translate }}
+            </mat-error>
+        </mat-form-field>
+        <os-chip *ngIf="name.value" [color]="currentColor">{{ name.value }}</os-chip>
+    </form>
+</div>
+<div mat-dialog-actions>
+    <button mat-raised-button color="primary" (click)="onSaveClicked()">{{ 'Save' | translate }}</button>
+    <button mat-raised-button [mat-dialog-close]="null">{{ 'Cancel' | translate }}</button>
+</div>
--- a/client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.ts
+++ b/client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.ts
@@ -1,0 +1,89 @@
+import { Component, Inject, OnInit } from '@angular/core';
+import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+
+import { HtmlColor } from 'app/core/definitions/key-types';
+import { ComponentServiceCollector } from 'app/core/ui-services/component-service-collector';
+import { ViewOrganisationTag } from 'app/management/models/view-organisation-tag';
+import { BaseComponent } from 'app/site/base/components/base.component';
+
+interface OrganisationTagDialogData {
+    organisationTag?: ViewOrganisationTag;
+    getRandomColor: () => HtmlColor;
+}
+
+@Component({
+    selector: 'os-organisation-tag-dialog',
+    templateUrl: './organisation-tag-dialog.component.html',
+    styleUrls: ['./organisation-tag-dialog.component.scss']
+})
+export class OrganisationTagDialogComponent extends BaseComponent implements OnInit {
+    public get isCreateView(): boolean {
+        return this._isCreateView;
+    }
+
+    public get currentColor(): string {
+        return this._lastValidColor;
+    }
+
+    public organisationTagForm: FormGroup;
+
+    private _isCreateView = false;
+    private _lastValidColor: string;
+
+    public constructor(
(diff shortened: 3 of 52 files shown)


client/src/app/management/components/meeting-list/meeting-list.component.html
```diff
@@ -58,6 +58,16 @@ <h2>
         </div>
     </div>
 
+    <div *pblNgridCellDef="'organisation_tags'; row as meeting" class="cell-slot fill">
+        <div *ngIf="meeting.organisation_tags?.length">
+            <mat-chip-list>
+                <os-chip *ngFor="let tag of meeting.organisation_tags" [color]="tag.color">
```
FinnStutzenstein, line 64: Just a suggestion: Make a `os-tags` component taking in a list (or a single one) of tags, so it is even easier to use.

client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.html
```diff
@@ -0,0 +1,27 @@
+<h2 mat-dialog-title>{{ (isCreateView ? 'Create an organization tag' : 'Edit an organization tag') | translate }}</h2>
+<div mat-dialog-content class="organization-tag-content">
+    <form [formGroup]="organisationTagForm">
+        <mat-form-field appearance="fill">
+            <mat-label>{{ 'Name' | translate }}</mat-label>
+            <input #name matInput formControlName="name" required />
+        </mat-form-field>
+        <mat-form-field appearance="fill">
+            <mat-label>{{ 'Color' | translate }}</mat-label>
+            <input matInput formControlName="color" [required]="isCreateView" />
```
FinnStutzenstein, line 10: Please create a custom validation. Maybe ther is an existing regexvalidator you can use. The regex would be `^[0-9a-f]{6}$` if i'm not mistaken.

FinnStutzenstein: Currently I can enter too many or too less characters and also invalid ones.

client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+            this.updateForm();
+        }
+    }
+
+    public onSaveClicked(): void {
+        const { name, color }: { name: string; color: string } = this.organisationTagForm.value;
+        this.dialogRef.close({ name, color: `#${color}` });
+    }
+
+    public generateColor(): void {
```
FinnStutzenstein, line 55: this code is duplicated

client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private createForm(): void {
+        this.organisationTagForm = this.fb.group({
+            name: ['', Validators.required],
+            color: ['']
+        });
+        this.subscriptions.push(
+            this.organisationTagForm.get('color').valueChanges.subscribe((currentColor: string) => {
+                if (currentColor.length === 3 || currentColor.length === 6) {
```
FinnStutzenstein, line 71: I know github allows 3 digit colors, but we do not.

client/src/app/shared/components/chip/chip.component.html
```diff
@@ -0,0 +1,3 @@
+<mat-basic-chip class="os-chip" [disableRipple]="true">
```
FinnStutzenstein, line 1: ah you created this one newly. Since it is just used for tags, you can add more flexible options as stated above.

Maybe do a very variable input: It can be a string, A ViewOrganizationTag or a list of them or so.

client/src/app/shared/components/chip/chip.component.scss
```diff
@@ -0,0 +1,59 @@
...
+    --os-chip-threshold: 0.5;
+    /**
+     * The threshold at which a darker border will be applied.
+     * Range: decimals from 0 to 1,
+     * recommended 0.8+
+     */
+    --os-chip-border-threshold: 0.8;
+}
+
+.os-chip {
```
FinnStutzenstein, line 23: this is very nice. Didn't expect it to be already implemented. Well done!

client/src/app/shared/components/chip/chip.component.scss
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * 2) sets the border as a 50% darker shade of the base color, ONLY if background color luma is higher than the border threshold.
+     * To achieve this I use the same sub-zero or higher-than-max technique, only this time using the Alpha value from an RGBA declaration.
+     * This results in a border that's either fully transparent or fully opaque
+     */
+    --border-alpha: calc((var(--perceived-lightness) - var(--os-chip-border-threshold)) * 100);
+
+    border-width: 0.2em;
+    border-style: solid;
+    border-color: rgba(
```
FinnStutzenstein, line 53: I do not like having a different border color for bright colors. Please just use the background color. All in all I did only see the tag in the dialog due to autoupdate issues: For me they are a bit too large. They should have the same appearance as the tags used in OS3. Can you equalize the style?

-> This is the same point I meantioned above: Maybe you can use this component for both of them allowing various inputs.

client/src/app/shared/components/chip/chip.component.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+        this._color = color;
+        this.recalcForegroundColor();
+    }
+
+    public get color(): HtmlColor {
+        return this._color;
+    }
+
+    private get red(): string {
```
FinnStutzenstein, line 24: Can you put this into a "color-service" dealing with parsing/normalizing/generating color values?

FinnStutzenstein (changes requested):
- Please provide a random initial color for the dialog when creating a new one.
- I do not get an autoupdate when creating one. The backend does not link them to the orga: See my coment there: https://github.com/OpenSlides/OpenSlides/wiki/organisation_tag.create -> this required you to always send `organisation_id: 1` (hardcoded)

Otherwise, it looks good so far! I could only test the deletion and creation. More to come, if the backend is fixed.

FinnStutzenstein:
Regarding the style and my "it looks too large": Personally, I find the OS3 tags too narrow. Something in the middle would be nice. But ultimatively ask Martin and Emanuel for design decisions like this. For now I would be happy if all labels/tags/... share the same code for the behavior and style, so there is exactly one place to change it afterwards.

client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.ts
GabrielInTheWorld: It is allowed, because HTML-colors can be define with three hex-digits, but okay - I will change this.

client/src/app/shared/components/chip/chip.component.scss
GabrielInTheWorld: Thanks!

client/src/app/management/components/organisation-tag-dialog/organisation-tag-dialog.component.html
GabrielInTheWorld: > Please create a custom validation. Maybe ther is an existing regexvalidator you can use. The regex would be `^[0-9a-f]{6}$` if i'm not mistaken.

Yes, that's right - I will extend this to `^[0-9a-fA-F]{6}$`.

client/src/app/management/components/meeting-list/meeting-list.component.html
GabrielInTheWorld: Hmm, I don't think, that this is a good pattern. Let us discuss this on monday.

GabrielInTheWorld:
https://stackblitz.com/edit/angular-qplyzs?file=src/app/only-number.directive.ts

FinnStutzenstein:
Have a look in the docs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#using_test_on_a_regex_with_the_global_flag and this one: https://stackoverflow.com/questions/11477415/why-does-javascripts-regex-exec-not-always-return-the-same-value#:~:text=A%20JavaScript%20RegExp%20object%20is,manually%2C%20set%20the%20lastIndex%20property.

The RegExp instance is stateful.

GabrielInTheWorld:
> Have a look in the docs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#using_test_on_a_regex_with_the_global_flag and this one: https://stackoverflow.com/questions/11477415/why-does-javascripts-regex-exec-not-always-return-the-same-value#:~:text=A%20JavaScript%20RegExp%20object%20is,manually%2C%20set%20the%20lastIndex%20property.
> 
> The RegExp instance is stateful.

Top!

FinnStutzenstein (approved):
Looks good and tested the orga tag list only. We have to wait for https://github.com/OpenSlides/openslides-backend/issues/699 to test the other update/create views that uses the orga tags. But since it is implemented, we can merge this and adjust bugs there later on.

PR #207, opened 2021-04-29
Add a warning editing motion with amendments

tsiegleauq:
(as in OS3)
Add a warning when editing a motion with existing amendments

### DIFF
--- a/client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
+++ b/client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
@@ -193,6 +193,7 @@ export class MotionDetailComponent extends BaseModelContextComponent implements 
 
     public enterEditMotion(): void {
         this.editMotion = true;
+        this.showMotionEditConflictWarningIfNecessary();
     }
 
     public leaveEditMotion(): void {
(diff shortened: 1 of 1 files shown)


client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
```diff
@@ -503,6 +508,10 @@ export class MotionDetailComponent extends BaseModelContextComponent implements
 
     public enterEditMotion(): void {
         this.editMotion = true;
+
+        if (this.editMotion) {
+            this.showMotionEditConflictWarningIfNecessary();
+        }
```
GabrielInTheWorld, line 514: In which case is this `false`?
I think, it can only be `this.showMotionEditConflictWarningIfNecessary()` (without a condition).

GabrielInTheWorld (changes requested):

client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
tsiegleauq: * you click "edit motion"
* if you happen to enter edit mode 
* `showMotionEditConflictWarningIfNecessary` detects if your motion has any paragraph based amendments
* if does it shows a warning, that editing that motion will lead into inconsistencies

GabrielInTheWorld: Hmm, I meant the condition, you check:
```ts
if (this.editMotion) {
    this.showMotionEditConflictWarningIfNecessary();
}
```
I don't know, if you are referring to this. I think, it can only be:
```ts
this.showMotionEditConflictWarningIfNecessary();
```
It's because two lines above, the flag `editMotion` will set to true (every time, someone enters the edit mode).

GabrielInTheWorld (approved):
See my last comment above.

client/src/app/site/motions/modules/motion-detail/components/motion-detail/motion-detail.component.ts
tsiegleauq: ah yes, have not seen that

tsiegleauq:
rebased. Fixed an issue where a subscription was never removed.

PR #209, opened 2021-05-03
Fixed lifecycle in multiple places

FinnStutzenstein:
- DS was cleared too late
- setting the meeting id was too late
- operator can now be "not ready". Used when changing meetings etc.
- active meeting service did not clean up subscriptions when leaving
meeting
- call to has perms in meeting list
- correctly use undefined instead of null to indicate uninitialized
state and prevent multiple updates
- many more...

### DIFF
--- a/client/src/app/core/repositories/base-repository.ts
+++ b/client/src/app/core/repositories/base-repository.ts
@@ -119,10 +119,6 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
         return this.repositoryServiceCollector.relationManager;
     }
 
-    protected get authService(): AuthService {
-        return this.repositoryServiceCollector.authService;
-    }
-
     public constructor(
         private repositoryServiceCollector: RepositoryServiceCollectorWithoutActiveMeetingService,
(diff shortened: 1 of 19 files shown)


client/src/app/core/repositories/base-repository.ts
```diff
@@ -291,7 +285,7 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
         if (!this.viewModelSubjects[id]) {
             this.viewModelSubjects[id] = new BehaviorSubject<V>(this.viewModelStore[id]);
         }
-        return this.viewModelSubjects[id].pipe(filter(value => !!value));
+        return this.viewModelSubjects[id].asObservable();
```
GabrielInTheWorld, line 288: Why did you change this? Nearly in every repo or component we catch, if a value is existing. Therefore, it can be filtered through the `base-repository`, I think. Or do you need anywhere the value `undefined`?

GabrielInTheWorld (commented):
Nice re-work! I will checkout this branch and test it.

GabrielInTheWorld (approved):

PR #213, opened 2021-05-03
Adds error-handling to streams

GabrielInTheWorld:
As OS3

### DIFF
--- a/client/src/app/core/core-services/http-stream.service.ts
+++ b/client/src/app/core/core-services/http-stream.service.ts
@@ -4,6 +4,7 @@ import { Injectable } from '@angular/core';
 import { HttpOptions } from '../definitions/http-options';
 import { EndpointConfiguration } from './http-stream-endpoint.service';
 import { Stream } from './stream';
+import { CommunicationError, ErrorType } from './stream-utils';
 
 export type Params = HttpParams | { [param: string]: string | string[] };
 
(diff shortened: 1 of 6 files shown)


tsiegleauq (approved):
for these kind of logging we might consider to use something like `console.info` or something else that is a little more suitable for stuff that is not worth logging. besides that it looks fine. Not sure how to test it so merge on your own if you feel confident in these changes

client/src/app/core/core-services/stream.ts
```diff
@@ -47,23 +47,31 @@ export class Stream<T> {
...
+                if (event.type === HEADER_EVENT_TYPE) {
+                    const headerResponse = event as HttpHeaderResponse;
+                    this._statuscode = headerResponse.status;
+                    if (headerResponse.status >= 400) {
+                        this.hasError = true;
+                    }
+                } else if ((event as HttpEvent<string>).type === PROGRESS_EVENT_TYPE) {
+                    this.handleMessage(event as HttpDownloadProgressEvent);
+                } else if ((event as HttpEvent<string>).type === FINISH_EVENT_TYPE) {
+                    this.errorHandler('The stream was closed');
```
FinnStutzenstein, line 65: Please have a look into OS3's improved error handling: `handleError` accepts more parameters. Have a look into the path the error goes and see where the variables are used. I think especially the server/client error type is used to determine whether to retry or not. Also small thinks like: On the first retry, the timeout is fixed to 500ms.

FinnStutzenstein (changes requested):

GabrielInTheWorld:
@FinnStutzenstein I reworked this branch. Please review again and give me a feedback.

client/src/app/core/core-services/http-stream.service.ts
```diff
@@ -14,24 +15,33 @@ export class StreamConnectionError extends Error {
...
-        messageHandler: (message: any, streamId: number) => void,
+        messageHandler: (message: T, streamId: number) => void,
         public params: () => Params,
         public body: () => any,
         public description: string
     ) {
-        this.messageHandler = (message: any) => messageHandler(message, this.id);
+        // this.messageHandler = (message: T) => messageHandler(message, this.id);
+        this.messageHandler = (message: T) => {
+            // {connected: true} is a special message just to trigger the code below
```
FinnStutzenstein, line 37: Please change this to the following code since we do not have connect messages anymore
```
if (this.hasErroredAmount  > 0) {
    console.log(`resetting error amount for ${this.endpoint} since there was a connect message`);                
    this.hasErroredAmount = 0;
}
messageHandler(message, this.id);
```

FinnStutzenstein (approved):
This looks good. Should be a major improvement in the error handling

PR #218, opened 2021-05-06
Redirect to login if unauthed in meeting

### LINKED ISSUE #214: Operator will not be ready if staying on a subroute
If a user stays unauthenticated for example on `/1/motions/5`, the operator will not be ready. So, the user will not be redirected to the login-mask and sees only a white page.

FinnStutzenstein:
fixes #214


GabrielInTheWorld (approved):
In my test, this fix works!

PR #222, opened 2021-05-10
Introduced accessibility for models

FinnStutzenstein:
A view model has now `canAccess` which should determine, whether the
operator can access the model. Currently, only ViewMeeting is affected:
We need to hide those meetings with too less information. This happens
due to the fact, that each meeting is always given to a user (even
anonymous) to provide basic infos, e.g. for login.

These meetings must not be shown in lists and the normal UI. The base
repo only outputs accessible models by default. There are some new
"unsafe" accessors to query unaccessible models. This will be needed by
the per-meeting-login mask.

Additionally moved some misplaced files and worked on the meeting edit.
It is not finished, but wrong parts are removed so far. A good
update-method for the server is missing.


client/src/app/core/repositories/base-repository.ts
```diff
@@ -31,15 +31,15 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
     protected viewModelSubjects: { [modelId: number]: BehaviorSubject<V> } = {};
 
     /**
-     * Observable subject for the whole list. These entries are unsorted and not piped through
-     * auditTime. Just use this internally.
+     * Observable subject for the whole list. These entries are unsorted, not piped through
+     * auditTime ad can contain unaccessible models. Just use this internally.
```
FinnStutzenstein, line 35: typo

PR #227, opened 2021-05-20
Add motion submitters in the agenda subtitles. Fixes #206

### LINKED ISSUE #206: Dynamic agenda subtitles cannot be supported
I noticed something strange while working on https://github.com/OpenSlides/openslides-client/pull/204.

implicit agenda content objects cannot resolve the second level of underlying ViewModels.
https://github.com/OpenSlides/openslides-client/blob/fe0974d0f1490d4259f1ecaf5373c585c9d799b5/client/src/app/site/agenda/components/agenda-list/agenda-list.component.ts#L162-L171
As we see here, the content_object_id will be resolved to their corresponding `title` filedset. This us usually fine, but limits to use native values of the model only.

Since the detail of the ViewModel is hidden in this case (can be Motion, Assignment or Topic) we cannot resolve nested ViewModels anymore, since the fieldset can only access native values of their model.

That means, that we currently could not resolve the (motion) submitter (as user) from the agenda item. The subtitle of motions are their submitter, even in the agenda list.

FinnStutzenstein:
Follow item->motion->submitter->user in the agenda. Adjusted the
modelbuilder to not warn for invali relations and explicitly skip them.
This allows for following some relations for only a subset of models in
a genereic scenario.

### DIFF
--- a/client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.html
+++ b/client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.html
@@ -49,7 +49,7 @@
             </div>
 
             <!-- Subtitle line -->
-            <div class="subtitle ellipsis-overflow" *ngIf="showSubtitle">
+            <div class="subtitle ellipsis-overflow" *ngIf="showSubtitles | async">
                 {{ item.getSubtitle() }}
             </div>
 
--- a/client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
+++ b/client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
@@ -4,6 +4,7 @@ import { ActivatedRoute, Router } from '@angular/router';
 
 import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 import { PblColumnDefinition } from '@pebula/ngrid';
+import { Observable } from 'rxjs';
 
 import { AgendaCsvExportService } from '../../services/agenda-csv-export.service';
 import { AgendaFilterListService } from '../../services/agenda-filter-list.service';
@@ -19,6 +20,7 @@ import {
     MeetingProjectionType,
     MeetingRepositoryService
 } from 'app/core/repositories/management/meeting-repository.service';
+import { SUBMITTER_FOLLOW } from 'app/core/repositories/motions/motion-repository.service';
 import { TopicRepositoryService } from 'app/core/repositories/topics/topic-repository.service';
 import { ComponentServiceCollector } from 'app/core/ui-services/component-service-collector';
 import { DurationService } from 'app/core/ui-services/duration.service';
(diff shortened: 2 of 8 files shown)


client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
```diff
@@ -179,7 +188,7 @@ export class AgendaItemListComponent extends BaseListViewComponent<ViewAgendaIte
                     follow: [
                         {
                             idField: 'content_object_id',
-                            follow: [SPEAKER_BUTTON_FOLLOW],
+                            follow: [SPEAKER_BUTTON_FOLLOW, SUBMITTER_FOLLOW],
```
FinnStutzenstein, line 191: @tsiegleauq I didn't remembered, that I already had this implemented :D It wasn't in use (The warning I removed was never triggered), this is the first case.

client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.html
```diff
@@ -49,7 +49,7 @@ <h2>{{ 'Agenda' | translate }}</h2>
             </div>
 
             <!-- Subtitle line -->
-            <div class="subtitle ellipsis-overflow" *ngIf="showSubtitle">
+            <div class="subtitle ellipsis-overflow" *ngIf="showSubtitles">
```
tsiegleauq, line 52: `*ngIf="showSubtitles | async"`

client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
```diff
@@ -57,7 +58,7 @@ export class AgendaItemListComponent extends BaseListViewComponent<ViewAgendaIte
      */
     public isNumberingAllowed: boolean;
 
-    public showSubtitle: boolean;
+    public showSubtitles: boolean;
```
tsiegleauq, line 61:
```ts
public showSubtitles: Observable<boolean> = this.meetingsSettingsService.get('agenda_show_subtitles');
```

client/src/app/site/agenda/components/agenda-item-list/agenda-item-list.component.ts
```diff
@@ -151,7 +152,7 @@ export class AgendaItemListComponent extends BaseListViewComponent<ViewAgendaIte
                 .subscribe(autoNumbering => (this.isNumberingAllowed = autoNumbering)),
             this.meetingsSettingsService
                 .get('agenda_show_subtitles')
-                .subscribe(showSubtitle => (this.showSubtitle = showSubtitle)),
+                .subscribe(showSubtitles => (this.showSubtitles = showSubtitles)),
```
tsiegleauq, line 155: can be removed if used like above

tsiegleauq (approved):
Works fine. One style remark: Subscribe is usually not necessary when you don't do anything with the result of the subscription.

FinnStutzenstein:
Thanks for the remark. Done.

tsiegleauq (approved):

PR #229, opened 2021-05-26
Adds a 'basic-list-view.component' to detach 'list-view-table.component' from operator

### LINKED ISSUE #221: refactor list-view-table to not be meeting dependend
When I open the committee list, I get one warning with `has perms: Usage outside of meeting!` originating here: https://github.com/OpenSlides/openslides-client/blob/master/client/src/app/shared/components/list-view-table/list-view-table.component.ts#L273

This code is called once during https://github.com/OpenSlides/openslides-client/blob/master/client/src/app/shared/components/list-view-table/list-view-table.component.ts#L426. Since the list view table is now also used outside of meetings, calls to hasPerms and meeting specific things like projector are wrong for non-meetings. Find a way of refactoring the list view to either:
- exclude all meeting related stuff per default (IMO this is not really possible)
- create a new component with just basic stuff and let the current list view table inherit from it and implement meeting related things there. So the base class can be used in e.g. the committee list.

GabrielInTheWorld:
I had to reduplicate the template of the `list-view-table.component` for the `basic-list-view-table.component`, because Angular does not provide a feature like 'template-inheritance'.

Fixes #221.


FinnStutzenstein (approved):

GabrielInTheWorld:
I have reworked this with the use of 'content-projection'. Now, it is a little bit strange because the `list-view-table.component` inherits the `base-list-view-table.component` in the 'code-file' and include it in the 'template-file'.
But, in this way the template is not duplicated and we have only to maintain the template-file of `base-list-view-table.component`.

PR #232, opened 2021-05-28
Fixes tile view

### LINKED ISSUE #216: The tile-view for motions does not work properly
Only the non-category `Unset` is shown. To fix this, the information `parent_id`, `child_ids` and `motion_ids` has to be loaded. Furthermore, a recursively dependency from child_ids to their children has to be resolved.

Depends on #228.

GabrielInTheWorld:
Fixes #216.

Depends on #231.

### DIFF
--- a/client/src/app/site/motions/modules/motion-list/components/motion-list/motion-list.component.ts
+++ b/client/src/app/site/motions/modules/motion-list/components/motion-list/motion-list.component.ts
@@ -324,6 +324,11 @@ export class MotionListComponent extends BaseListViewComponent<ViewMotion> imple
                     ],
                     fieldset: 'list',
                     additionalFields: ['text']
+                },
+                {
+                    idField: 'motion_category_ids',
+                    fieldset: 'list',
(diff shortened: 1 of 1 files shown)


client/src/app/site/motions/modules/motion-list/components/motion-list/motion-list.component.ts
```diff
@@ -324,6 +324,11 @@ export class MotionListComponent extends BaseListViewComponent<ViewMotion> imple
                     ],
                     fieldset: 'list',
                     additionalFields: ['text']
+                },
+                {
+                    idField: 'motion_category_ids',
+                    fieldset: 'list',
+                    follow: ['parent_id', 'child_ids']
```
FinnStutzenstein, line 331: You do not have to follow them, just request them as either additional fields or create a fieldset. This is because all categories are already included in the request itself; the ids are needed by the client for the relations itself.

FinnStutzenstein (approved):

PR #265, opened 2021-06-10
Add some hyphenation

tsiegleauq:
Use hyphenation for motion details poll display.
More lenient display of long poll names using hyphenation


tsiegleauq:
merged cause all changes are trivial

PR #266, opened 2021-06-10
Consequent enum of assignment detail

tsiegleauq:
Also fix all sorting lists

### DIFF
--- a/client/src/app/shared/components/list-of-speakers-content/list-of-speakers-content.component.html
+++ b/client/src/app/shared/components/list-of-speakers-content/list-of-speakers-content.component.html
@@ -106,7 +106,8 @@
                         </div>
                         <!-- Extra line -->
                         <ng-container
-                            *ngTemplateOutlet="speakerSubtitle; context: { speaker: speaker, showcolor: true }"
+                            [ngTemplateOutlet]="speakerSubtitle"
+                            [ngTemplateOutletContext]="{ speaker: speaker, showcolor: true }"
                         ></ng-container>
                     </div>
                     <div class="speaker-controls one-line">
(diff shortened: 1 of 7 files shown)


tsiegleauq:
requires rework of ListOfSpeakers to include PointOfOrder and pro contra

client/src/app/shared/components/list-of-speakers-content/list-of-speakers-content.component.html
```diff
@@ -106,7 +106,8 @@ <h1 class="los-title" *ngIf="!customTitle">
                         </div>
                         <!-- Extra line -->
                         <ng-container
-                            *ngTemplateOutlet="speakerSubtitle; context: { speaker: speaker, showcolor: true }"
+                            [ngTemplateOutlet]="speakerSubtitle"
+                            [ngTemplateOutletContext]="{ speaker: speaker, showcolor: true }"
```
GabrielInTheWorld, line 110: Is there a reason why you transform it from a structural directive to an attribute directive?
Nevertheless, it's okay for me and makes the same.

GabrielInTheWorld (approved):

client/src/app/shared/components/list-of-speakers-content/list-of-speakers-content.component.html
tsiegleauq: Habbit, and it's easier to maintain, since `[ngTemplateOutletContext]` requires only a json that can be slightly easier replaced than in a structural directive. Furthermore, elements can only have one structural directive, thus but as many input and output attributes as required, thus I try to limit the usage to *ngIf, *ngFor or *osPerms

PR #267, opened 2021-06-10
Refine AP Interaction

tsiegleauq:
In auto pilot view:
klicking the projector opens the content object (rather than the
projector detail view)
clicking AP title now opens the content objects as normal link, rather
than having an "open external" icon
clicking the projector name opens the projector detail view
add "readd last speaker" in AP


GabrielInTheWorld (approved):
The changes looks good for me.

PR #269, opened 2021-06-11
Clears meeting-models in related repos

GabrielInTheWorld:
@FinnStutzenstein please review. I have noticed, that the lifecycle-fix (3979734) deletes models in the `datastore.service`, but it does not delete models in the related repos. Therefore, the models was still seen in view-model-lists.

With this PR the described behavior will be prevented.

### DIFF
--- a/client/src/app/core/core-services/autoupdate.service.ts
+++ b/client/src/app/core/core-services/autoupdate.service.ts
@@ -6,6 +6,7 @@ import { CollectionMapperService } from './collection-mapper.service';
 import { CommunicationManagerService } from './communication-manager.service';
 import { DataStoreService, DataStoreUpdateManagerService } from './data-store.service';
 import { HTTPMethod } from '../definitions/http-methods';
+import { Collection } from '../definitions/key-types';
 import { ModelRequestBuilderService, SimplifiedModelRequest } from './model-request-builder.service';
 import { Mutex } from '../promises/mutex';
 
@@ -19,21 +20,28 @@ export interface HasFields {
     fields: Fields;
 }
 
+/**
+ * Whether a relation requests a list of view-models that updates the whole view-model-store of one repo.
+ */
+export interface CanBeFullListUpdate {
+    isFullList?: boolean;
+}
+
 export interface ModelRequest extends HasFields {
     ids: number[];
     collection: string;
 }
 
-export interface GenericRelationFieldDecriptor extends HasFields {
+export interface GenericRelationFieldDecriptor extends HasFields, CanBeFullListUpdate {
     type: 'generic-relation-list' | 'generic-relation';
 }
 
-export interface RelationFieldDescriptor extends HasFields {
+export interface RelationFieldDescriptor extends HasFields, CanBeFullListUpdate {
     type: 'relation-list' | 'relation';
     collection: string;
 }
 
-export interface StructuredFieldDecriptor {
+export interface StructuredFieldDecriptor extends CanBeFullListUpdate {
     type: 'template';
     values?: RelationFieldDescriptor | GenericRelationFieldDecriptor;
 }
@@ -48,6 +56,7 @@ interface DeletedModels {
 interface ChangedModels {
     [collection: string]: BaseModel[];
 }
+
 /**
  * Handles the initial update and automatic updates
  * Incoming objects, usually BaseModels, will be saved in the dataStore (`this.DS`)
@@ -88,15 +97,25 @@ export class AutoupdateService {
      * Which component opens which stream?
      */
     public async simpleRequest(simpleRequest: SimplifiedModelRequest, description: string): Promise<ModelSubscription> {
-        const request = await this.modelRequestBuilder.build(simpleRequest);
+        const {
+            request,
+            collectionsToUpdate
+        }: { request: ModelRequest; collectionsToUpdate: Collection[] } = await this.modelRequestBuilder.build(
+            simpleRequest
+        );
         console.log('autoupdate: new request:', description, simpleRequest, request);
-        return await this.request(request, description);
+        return await this.request(request, description, collectionsToUpdate);
     }
 
-    public async request(request: ModelRequest, description: string): Promise<ModelSubscription> {
+    private async request(
+        request: ModelRequest,
+        description: string,
+        collectionsToUpdate: Collection[]
+    ): Promise<ModelSubscription> {
         const closeFn = await this.communicationManager.connect<AutoupdateModelData>(
             'autoupdate',
-            (message, id) => this.handleAutoupdateWithStupidFormat(message, id),
+            (message, id, isFirstResponse) =>
+                this.handleAutoupdateWithStupidFormat(message, id, { collectionsToUpdate, isFirstResponse }),
             () => [request],
             null,
             description
@@ -104,17 +123,26 @@ export class AutoupdateService {
         return { close: closeFn };
     }
 
-    private async handleAutoupdateWithStupidFormat(autoupdateData: AutoupdateModelData, id: number): Promise<void> {
+    private async handleAutoupdateWithStupidFormat(
+        autoupdateData: AutoupdateModelData,
+        id: number,
+        args: { collectionsToUpdate: Collection[]; isFirstResponse: boolean }
+    ): Promise<void> {
         const modelData = autoupdateFormatToModelData(autoupdateData);
         console.log('autoupdate: from stream', id, modelData, 'raw data:', autoupdateData);
-        await this.handleAutoupdate(modelData);
+        await this.handleAutoupdate(modelData, args.collectionsToUpdate, args.isFirstResponse);
     }
 
-    private async handleAutoupdate(modelData: ModelData): Promise<void> {
+    private async handleAutoupdate(
+        modelData: ModelData,
+        collectionsToUpdate: Collection[],
+        isFirstResponse: boolean
+    ): Promise<void> {
         const unlock = await this.mutex.lock();
 
         const deletedModels: DeletedModels = {};
         const changedModels: ChangedModels = {};
+        const fullListUpdate: Collection[] = isFirstResponse ? collectionsToUpdate : [];
 
         for (const collection of Object.keys(modelData)) {
             for (const id of Object.keys(modelData[collection])) {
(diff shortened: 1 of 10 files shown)


client/src/app/core/core-services/data-store.service.ts
```diff
@@ -278,6 +278,8 @@ export class DataStoreService {
      */
     private readonly clearEvent: EventEmitter<void> = new EventEmitter<void>();
```
FinnStutzenstein, line 279: please merge this event with the clear event: Just one clear event, parametrized per collection.

client/src/app/core/core-services/data-store.service.ts
```diff
@@ -323,15 +329,18 @@ export class DataStoreService {
      * Deletes all models that belong to a meeting.
      */
     public async clearMeetingModels(): Promise<void> {
+        const removedCollections = [];
         for (const collection of Object.keys(this.modelStore)) {
             if (this.modelMapper.isMeetingSpecificCollection(collection)) {
                 this.remove(
```
FinnStutzenstein, line 335: I wonder why the repos do not get notified, the remove should work. I had a look: There is no updateslot running, when calling this from the active meeting service. Please have a look at `handleChangedAndDeletedModels` (`this.DSUpdateManager.getNewUpdateSlot` and `commit`) in the `autoupdate.service.ts`. Updateslots are the way that updates in DS are propagated to repos. `clear` is specially treated. You can decide: 
- Use an update slot and the event is not needed (not tested)
- Use the event, and do not call `remove`. The call can just be removed. Please call `triggerModifiedObservable` at the end of this method (as well as adding it to `clear`).

client/src/app/core/repositories/base-repository.ts
```diff
@@ -144,6 +144,11 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
     public onAfterAppsLoaded(): void {
         this.baseViewModelCtor = this.collectionMapperService.getViewModelConstructor(this.collection);
         this.DS.clearObservable.subscribe(() => this.clear());
+        this.DS.clearMeetingModelsObservable.subscribe(removedCollections => {
```
FinnStutzenstein, line 147: If you use just one observable, you can remove the line above.

FinnStutzenstein (changes requested):

client/src/app/core/core-services/data-store.service.ts
GabrielInTheWorld: I will take the second option. The reason is, that - when moving from a list to the dashboard - a user sees "missing content" in the list.

GabrielInTheWorld:
I updated this and furthermore implemented, that a full-list update will check, if some view-models are destroyed (and the client got no autoupdate).

client/src/app/core/core-services/autoupdate.service.ts
```diff
@@ -159,7 +184,7 @@ export class AutoupdateService {
             await this.DS.addOrUpdate(changedModels[collection]);
         }
 
-        this.DSUpdateManager.commit(updateSlot);
+        this.DSUpdateManager.commit(updateSlot, fullListUpdateForCollection);
```
FinnStutzenstein, line 187: do not pass this to the DS update manager - it does not have the right responsibility. Just get the whole collections here and do a `DS.remove` for these items (`remove` accepts multiple ids for one collection). So you do not have to modify the update manager and repos.

client/src/app/core/core-services/autoupdate.service.ts
```diff
@@ -137,15 +159,18 @@ export class AutoupdateService {
                     }
                 }
             }
+            fullListUpdate[collection] =
+                this.getDescriptorForCollection(originalRequest, collection)?.isFullList && isFirstResponse;
```
FinnStutzenstein, line 163: This is now calculated for every update. Calculate it once when creating the AU-request and just pass the list of collections. Also: Switch around the `&&` since in most cases `isFirstResponse` will be false, so the check is faster.

client/src/app/core/core-services/data-store.service.ts
```diff
@@ -2,6 +2,7 @@ import { EventEmitter, Injectable } from '@angular/core';
 
 import { Observable, Subject } from 'rxjs';
 
+import { FullListUpdate } from './autoupdate.service';
```
FinnStutzenstein, line 5: Undo this all, the DS should not have dependencies to the AU service.

client/src/app/core/definitions/relations.ts
```diff
@@ -26,6 +27,8 @@ export function makeO2O<A extends BaseViewModel, B extends BaseViewModel>(args:
     BField: keyof B & string;
     AIdField?: keyof A & string;
     BIdField?: keyof B & string;
+    AisFullList?: boolean;
```
FinnStutzenstein, line 30: Can *one* to *one* relations be a list?

client/src/app/core/definitions/relations.ts
```diff
@@ -59,7 +64,10 @@ export function makeM2O<O extends BaseViewModel, M extends BaseViewModel>(args:
     OIdField?: keyof O & string;
     MIdField?: keyof M & string;
     order?: keyof M & string;
+    OisFullList?: boolean;
```
FinnStutzenstein, line 67: In this case I think only the M-side is a valid list.

client/src/app/core/repositories/base-repository.ts
```diff
@@ -143,7 +143,15 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
 
     public onAfterAppsLoaded(): void {
         this.baseViewModelCtor = this.collectionMapperService.getViewModelConstructor(this.collection);
-        this.DS.clearObservable.subscribe(() => this.clear());
+        this.DS.clearObservable.subscribe(removedCollections => {
+            if (Array.isArray(removedCollections) && removedCollections.includes(this.collection)) {
```
FinnStutzenstein, line 147: Please join the two cases with `||`

client/src/app/core/repositories/base-repository.ts
```diff
@@ -346,13 +354,32 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
...
         modelIds.forEach(id => {
             this.updateViewModelObservable(id);
         });
     }
 
+    protected doFullListUpdate(modelIds: Id[]): void {
+        const deletedIds = this.getViewModelIdList().difference(modelIds);
+        this.deleteModels(deletedIds);
+        deletedIds.forEach(id => this.updateViewModelObservable(id));
+        this.updateViewModelListSubject(Object.values(this.viewModelStore));
```
FinnStutzenstein, line 371: Please undo this all. Small note: This would not be correct since the wull update is triggered in `commitUpdate` at the end - this would result in the observables triggering two times and the unsafe model list is not updated. This is the reason why it should be handled in the AU service: You do not have to change something here.

client/src/app/core/repositories/relations.ts
```diff
@@ -338,117 +344,136 @@ export const RELATIONS: Relation[] = [
...
     }),
     ...makeM2O({
         OViewModel: ViewMeeting,
         MViewModel: ViewAssignment,
         OField: 'assignments',
-        MField: 'meeting'
+        MField: 'meeting',
+        MisFullList: true
     }),
     ...makeM2O({
```
FinnStutzenstein, line 471: You only use the M-side of M2O and M2M. Please adjust the definition of M2= to just accept `isFullList` (without M) since the O-side is never a list. Also O2O is not used -> remove it.

client/src/app/core/core-services/autoupdate.service.ts
```diff
@@ -19,21 +20,28 @@ export interface HasFields {
...
+    isFullList?: boolean;
+}
+
 export interface ModelRequest extends HasFields {
     ids: number[];
     collection: string;
 }
 
-export interface GenericRelationFieldDecriptor extends HasFields {
+export interface GenericRelationFieldDecriptor extends HasFields, CanBeFullListUpdate {
```
FinnStutzenstein, line 35: The client-internal have nothing to do with the autoupdate service. You are misusing these interfaces to store some information about the data. If I read the code from `getDescriptorForCollection` right, you also just look at the first relations, but nested relations do not work right?

If you want, I can modify the request builder to return a list of collections that can be full-updated.

FinnStutzenstein (changes requested):

client/src/app/core/core-services/autoupdate.service.ts
GabrielInTheWorld: Has the AU-service knowledge about the current view-models in any repo?
I passed these information down to the repos, because I didn't see, that the AU-service have this knowledge and I don't think, that it gets these information.

FinnStutzenstein: No, it only operates on the DS, not the view model layer. The autoupdate system etc. is only on the model layer, the DSUpdateManager is the link to the view model side. First, everything is injected into the DS, so the DS has a correct and consistent data bases, than the commit is triggered and the repos do their stuff to create view models and publish them.

FinnStutzenstein: You can either use `DS.getCollection` or write a similar method to `clearMeetingModels` that takes a collection and clears it (Without triggering the clear observable - just doing a `DS.remove` since this method is tied to the DSUpdateManager.

GabrielInTheWorld:
@FinnStutzenstein Please review again. I looked into the `ModelRequestBuilder` and make already changes to it to have `ModelRequest`s separated from a list of full-updated collections.
I hope, this changes are like your intentions.

client/src/app/core/core-services/model-request-builder.service.ts
```diff
@@ -83,12 +83,53 @@ interface BaseSimplifiedModelRequest {
...
 }
 
 export const DEFAULT_FIELDSET = 'detail';
 
+class ModelRequestObject {
+    public readonly collection: Collection;
+    public readonly ids: Id[] | undefined;
+    public readonly fields: Fields;
+    public readonly simplifiedRequest: BaseSimplifiedModelRequest;
+    public readonly collectionsToUpdate: Collection[] = [];
```
FinnStutzenstein, line 102: I had problems with this kind of initialization before. The array might be static, so every instance of this class might share this array. Please set it in the constructor.

client/src/app/core/core-services/model-request-builder.service.ts
```diff
@@ -210,57 +252,80 @@ export class ModelRequestBuilderService implements OnAfterAppsLoaded {
...
-            descriptor = this.getGenericRelationFieldDescriptor(relation, follow);
+            response = this.getGenericRelationFieldDescriptor(relation, follow);
         } else {
-            descriptor = this.getStructuredFieldDescriptor(relation, follow);
+            response = this.getStructuredFieldDescriptor(relation, follow);
         }
 
-        fields[effectiveIdField] = descriptor;
+        modelRequestObject.fields[effectiveIdField] = response.descriptor;
+        modelRequestObject.collectionsToUpdate.push(...response.collectionsToUpdate);
```
FinnStutzenstein, line 277: Maybe a set as the type for `collectionsToUpdate` would be nice to avoid duplicates

FinnStutzenstein (changes requested):
Looks good!

client/src/app/core/core-services/autoupdate.service.ts
```diff
@@ -88,33 +90,52 @@ export class AutoupdateService {
      * Which component opens which stream?
      */
     public async simpleRequest(simpleRequest: SimplifiedModelRequest, description: string): Promise<ModelSubscription> {
-        const request = await this.modelRequestBuilder.build(simpleRequest);
+        const {
+            request,
+            collectionsToUpdate
+        }: { request: ModelRequest; collectionsToUpdate: Collection[] } = await this.modelRequestBuilder.build(
```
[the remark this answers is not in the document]
GabrielInTheWorld: Yes. Otherwise our tslint-config raises a warning...

client/src/app/core/core-services/model-request-builder.service.ts
GabrielInTheWorld: I don't know, if I understand you correctly, but I don't like to use a static array/set.

FinnStutzenstein (approved):
Not tested, looks very good

PR #293, opened 2021-06-23
Use vote pending state as subject

tsiegleauq:
Fixes an issue where the vote pending state was not updating correctly
when using state changes over the choice service


tsiegleauq:
some synergy, but not directly dependent on, https://github.com/OpenSlides/openslides-client/pull/258

PR #304, opened 2021-06-28
Adjust notify service for icc

tsiegleauq:
Adjust notify service for icc

Enable notify via new ICC integration.
Adjust previously used connection manager to use
new communiction manager

Attach osListenEditing to motion detail to get ICC notifications for
edits

### DIFF
--- a/client/src/app/core/core-services/http.service.ts
+++ b/client/src/app/core/core-services/http.service.ts
@@ -236,4 +236,29 @@ export class HttpService {
         const file = await this.get<Blob>(url, {}, {}, headers, 'blob');
         return await toBase64(file);
     }
+
+    /**
+     * Helper function to check if a given URL was reachable
+     * @param url
+     */
+    public async isServiceOnline(url: string): Promise<boolean> {
+        return new Promise<boolean>((resolve, reject) => {
+            const xhr = new XMLHttpRequest();
+            xhr.open('GET', url);
+            xhr.onload = () => {
+                if (xhr.status >= 200 && xhr.status < 300) {
+                    resolve(true);
+                } else {
+                    reject(false);
+                }
+            };
+            xhr.onerror = () => {
+                reject({
+                    status: xhr.status,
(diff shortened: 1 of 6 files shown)


tsiegleauq:
@FinnStutzenstein please review later

client/src/app/core/core-services/http.service.ts
```diff
@@ -236,4 +236,29 @@ export class HttpService {
         const file = await this.get<Blob>(url, {}, {}, headers, 'blob');
         return await toBase64(file);
     }
+
+    /**
+     * Helper function to check if a given URL was reachable
+     * @param url
+     */
+    public async isServiceOnline(url: string): Promise<boolean> {
```
FinnStutzenstein, line 244: Is this needed? I do not see any reference.

client/src/app/core/core-services/notify.service.ts
```diff
@@ -97,19 +101,81 @@ export class NotifyService {
...
+
+    private async connect(meetingId: number): Promise<void> {
+        if (!meetingId) {
+            throw new Error('Cannot connect to ICC, no meeting ID was provided');
+        }
+
+        this.disconnect();
+
+        const iccMeeting = `${IccPath}?meeting_id=${meetingId}`;
+        this.commManager.registerEndpoint('icc', iccMeeting, iccHealthPath, HTTPMethod.GET);
```
FinnStutzenstein, line 143: Nice, that this worked so easy :)

client/src/app/core/core-services/notify.service.ts
```diff
@@ -180,7 +247,8 @@ export class NotifyService {
         if (channels) {
             notify.to_channels = channels;
         }
-        this.websocketService.send('notify', notify);*/
+
+        await this.httpService.post<unknown>('/system/icc/send', notify);
```
FinnStutzenstein, line 251: You can use `IccPath` from above.

PR #305, opened 2021-06-28
Implements the meeting-import

### LINKED ISSUE #288: Add functionality to import a meeting
We need a functionality in the meeting list to import a meeting from a file. The file upload dialog can probably be recycled for that. There should be "some" slight validation to catch the most obvious stuff (but the server will do it anyways).

`meeting.ts` will require a new field `imported_at: number` - this is a timestamp hints when a meeting was imported. `null` indicated that a meeting was not imported but created manually.

Action:
https://github.com/OpenSlides/OpenSlides/wiki/meeting.import

Depends on the following backend issue:
https://github.com/OpenSlides/openslides-backend/issues/764

GabrielInTheWorld:
- Recycles the <os-media-upload-content />-component to <os-file-upload />-component to have a similar look-&-feel
- Adds a button to import a meeting to the 'meeting-list'
- Deletes the 'create-mediafile'-file because it is not needed anymore

Fixes #288.


tsiegleauq (approved):
Tested, works well enough. Server not ready yet, but does not really matter for this PR

PR #322, opened 2021-07-01
Fixes user-csv-import

GabrielInTheWorld:
- Replaces the templates of all csv-import-components with <basic-import-list />-component
- Moves the logic from base-import-list.component to basic-import-list.component


client/src/app/core/repositories/users/user-repository.service.ts
```diff
@@ -161,9 +155,9 @@ export class UserRepositoryService
      * @param patch An update-payload object or a function to generate a payload. The function gets a user as argument.
      * @param users A list of users, who will be updated.
      */
-    public async bulkUpdate(
-        patch: UserAction.UpdatePayload | ((user: ViewUser) => UserAction.UpdatePayload),
-        users: ViewUser[]
+    public async update(
```
tsiegleauq, line 158: function is not async but returns a real promise

client/src/app/core/repositories/users/user-repository.service.ts
```diff
@@ -441,7 +404,7 @@ export class UserRepositoryService
      * @param field The boolean field to set
      * @param value The value to set this field to.
      */
-    public async bulkSetState(users: ViewUser[], field: UserStateField, value: boolean): Promise<void> {
+    public async setState(field: UserStateField, value: boolean, ...users: ViewUser[]): Promise<void> {
```
tsiegleauq, line 407: function is not async but returns a real promise (no await)

client/src/app/shared/components/basic-import-list/basic-import-list.component.ts
```diff
@@ -0,0 +1,365 @@
...
+import { ImportListFirstTabDirective } from './import-list-first-tab.directive';
+import { ImportListLastTabDirective } from './import-list-last-tab.directive';
+
+@Component({
+    selector: 'os-basic-import-list',
+    templateUrl: './basic-import-list.component.html',
+    styleUrls: ['./basic-import-list.component.scss'],
+    encapsulation: ViewEncapsulation.None
+})
+export class BasicImportListComponent<M extends BaseModel> extends BaseModelContextComponent implements OnInit {
```
tsiegleauq, line 21: name conflicts heavily with `BaseImportListComponent` (directive) `client/src/app/site/base/components/base-import-list.component.ts`

client/src/app/shared/components/basic-import-list/basic-import-list.component.ts
```diff
@@ -0,0 +1,365 @@
...
+
+    /**
+     * Initializes the table
+     */
+    public initTable(): void {
+        const entryObservable = this.importer.getNewEntries();
+        this.subscriptions.push(
+            entryObservable.pipe(distinctUntilChanged(), auditTime(100)).subscribe(newEntries => {
+                this.hasFile = newEntries.length > 0;
+            })
```
tsiegleauq, line 175: `this.hasFile` way exclusively accessed with `this.importer`, this checking for `this.hasFile` seems neglectable (or should be inside an own function with `this.importer`).

Since audit time seems relevant somehow:
```ts
this.hasFile: Observable<boolean> = this.importer.getNewEntries().pipe(
            distinctUntilChanged(),
            auditTime(100),
            map(entries => entries.length > 0)
        );
```
the template can then have `hasFile | async` which might be sufficient. nGrid  is then probably not necessary anymore.

client/src/app/site/users/services/user-import.service.ts
```diff
@@ -48,13 +69,72 @@ export class UserImportService extends BaseImportService<User> {
...
+                gender: 'female',
+                default_vote_weight: 1.5
+            },
+            {
+                last_name: 'Executive Board',
+                is_present_in_meeting_ids: true,
+                username: 'executive',
+                default_vote_weight: 2.5
+            }
+        ];
```
tsiegleauq, line 130: this is merely a lengthy declaration. Better put it into an own file.

client/src/app/site/motions/services/motion-csv-export.service.ts
```diff
@@ -167,19 +179,32 @@ export class MotionCsvExportService {
...
+                category: 'A - Category',
+                submitters: 'Submitter A',
+                reason: 'Penguins can fly'
+            },
+            {
+                number: 'C2',
+                title: 'Title 3',
+                text: 'Text 3'
+            }
+        ];
```
tsiegleauq, line 206: lengthy declaration. Put into own file

tsiegleauq (changes requested):
looks a lot cleaner, however, there are certain things to enhance.

@FinnStutzenstein if i recall that correctly, we "used to" treat bulk actions and single create actions as @GabrielInTheWorld shows here, while working on 3.0. Was there a specific reason to drop it?

FinnStutzenstein:
> if i recall that correctly, we "used to" treat bulk actions and single create actions as @GabrielInTheWorld shows here, while working on 3.0. Was there a specific reason to drop it?

I'm not sure what you mean @GabrielInTheWorld did here (I did not look into all 1000 LOCs ;) ). For the bulk actions now: All actions should accept a list of instances, so you can create multiple objects in one action.

GabrielInTheWorld:
> looks a lot cleaner, however, there are certain things to enhance.
> 
> @FinnStutzenstein if i recall that correctly, we "used to" treat bulk actions and single create actions as @GabrielInTheWorld shows here, while working on 3.0. Was there a specific reason to drop it?

If I recall it correctly, then we have to use different functions for single and bulk actions because there were different endpoints to request to, in OpenSlides 3.x. Now, our action-system has only one endpoint and let us combine multiple actions in one request. This is the reason, why I unified the single and bulk actions for the same intention. They do the same. The only different is the amount of models that are involved.
Do @FinnStutzenstein and my replies answer your question, @tsiegleauq?

tsiegleauq (approved):

PR #341, opened 2021-07-02
Fix autoupdates for projections: Explicitly register for these

FinnStutzenstein:


client/src/app/shared/components/projector/projector.component.ts
```diff
@@ -208,6 +198,37 @@ export class ProjectorComponent extends BaseComponent implements OnDestroy {
...
+                type: projection.type || '',
+                options: projection.options || {}
+            };
+            if (projection.content?.error) {
+                slideData.error = projection.content.error;
+            }
+            return slideData;
+        });
+    }
+
```
tsiegleauq, line 231: This would be way easier (or in fact not necessary at all) if `ViewProjector.currentProjections` was `Observable<ViewProjection[]>`

tsiegleauq (approved):
Works. I think about a way to get this more pretty later

client/src/app/shared/components/projector/projector.component.ts
FinnStutzenstein: I had some thoughts about that already a year ago. If you are interested, we can talk monday or so.

PR #349, opened 2021-07-05
Relation Observables

FinnStutzenstein:


client/src/app/site/base/components/base.component.ts
```diff
@@ -123,7 +125,19 @@ export abstract class BaseComponent implements OnDestroy {
             this.messageSnackBar.dismiss();
         }
 
-        this.cleanSubjects();
+        this.cleanSubscriptions();
+    }
+
+    public updateSubscription(name: string, subscription: Subscription): void {
```
FinnStutzenstein, line 131: This is a solution to an issue I often have: Subscribing to an observable based on a changing fact. E.g. subscribing to an assignment by id from the url -> you have the route observable and inside it's `subscribe` you have to (a) unsubscribe the old assignment subscription (if any) and (b) subscribe to the new assignment. Also you have to clean up on `ngOnDestroy`. This function allows for repeaded one-active-only subscriptions, so doing

```
this.updateSubscription("assignment", <some getter>.subscribe(<some code>));
this.updateSubscription("assignment", <some getter>.subscribe(<some code>));
this.updateSubscription("assignment", <some getter>.subscribe(<some code>));
```

is harmless since it cleans up when subscribing to a new subscription and it is cleaned up in `ngOnDestroy`. So we now have two systems for subscriptions: You can push into `this.subscriptions` (e.g. useful in the CTOR, or when many things are registered just once` and the new one for repeaded-subscriptions.

client/src/app/site/projector/models/view-projector.ts
```diff
@@ -22,6 +24,7 @@ export class ViewProjector extends BaseViewModel<Projector> {
 }
 interface IProjectorRelations {
     current_projections: ViewProjection[];
+    current_projections_as_observable: Observable<ViewProjection[]>;
```
FinnStutzenstein, line 27: This is the new main bit: You can add `_as_observable` to each relation, and it will provide you the Observable. I've only added it here in the type system, but you can add this type everywhere you want, except for generic and structured relations; those are not implemented yet (since not needed yet).

One caveat: imagine I have projector `P` and get the `current_projections` as an observable:
```
P.current_projections_as_observable.subscribe(projections => <do something>);
```
I'll get updates for the list of projections if at least one is changed, BUT: I do not get updates, if the relation itself changes. E.g. if projections `[2, 4]` were in `P.current_projection_ids` and it is updated to e.g. `[2, 4, 8]`, this will not be noticed! Why:

Since the relation itself changes (and not the content), you will get an autoupdate for `P` itself (I assume, that you have an autoupdate for the relation setup, if you want to use the relation...). So you got `P` from somewhere, e.g. `projectorRepo.getViewModelObservable)`, so you have to resubscribe to the relation everytime `P` changes and do not forget to unsubscribe the old one.

client/src/app/shared/components/projector/projector.component.ts
```diff
@@ -177,7 +175,17 @@ export class ProjectorComponent extends BaseComponent implements OnDestroy {
         this._projector = projector;
         if (projector?.id !== this.oldProjectorId) {
             this.oldProjectorId = projector?.id;
-            this.updateProjectionSubscription();
+            this.updateSlides();
+            if (this.projector) {
+                this.updateSubscription(
+                    'projections',
+                    this.projector.current_projections_as_observable.subscribe(projections => {
```
FinnStutzenstein, line 182: This is how it is now used.

FinnStutzenstein:
@tsiegleauq If you have questions, please ask. It is intentionally left as WIP.

client/src/app/site/base/components/base.component.ts
tsiegleauq: Having a route subscription is nice when you never leave your component, i.e going back and forth in motions or only navigating between assignments using the full screen search components.
Otherwise, if you need it synchronously, you could also access the activated route directly and extract the router patameters

client/src/app/shared/components/projector/projector.component.ts
```diff
@@ -177,7 +175,20 @@ export class ProjectorComponent extends BaseComponent implements OnDestroy {
...
+                        return {
+                            collection: collectionFromFqid(projection.content_object_id),
+                            data: projection.content,
+                            stable: !!projection.stable,
+                            type: projection.type || '',
+                            options: projection.options || {},
+                            ...(!!projection.content?.error && { error: projection.content.error })
+                        } as SlideData;
+                    });
+                })
```
tsiegleauq, line 190: I replaced `updateSlides()`, returns an observable now, reduces the logic quite a bit

tsiegleauq (approved):
looks very good to me. Cannot really successfully test it tho, since autoupdate currently appears to has some problems. However, the changes seem quite promising.

client/src/app/shared/components/projector/projector.component.ts
FinnStutzenstein: Nice, I will  try to use `switchMap` to chain a new projector-observable with the `current_projections_as_observable`.

client/src/app/shared/components/projector/projector.component.ts
```diff
@@ -30,20 +30,17 @@ export const PROJECTOR_CONTENT_FOLLOW: Follow = {
     styleUrls: ['./projector.component.scss']
 })
 export class ProjectorComponent extends BaseComponent implements OnDestroy {
+    private readonly projectorSubject = new BehaviorSubject<ViewProjector | null>(null);
+
     @Input()
     public set projector(projector: ViewProjector) {
-        this.setProjector(projector);
+        this.projectorSubject.next(projector);
```
FinnStutzenstein, line 37: @tsiegleauq Best would be to change this input to an observalbe directly; But this is not directly possible (e.g. the projector list); so this might be done later.

client/src/app/shared/components/projector/projector.component.ts
```diff
@@ -171,62 +167,48 @@ export class ProjectorComponent extends BaseComponent implements OnDestroy {
...
-            this.updateScaling();
-        }
-        this.css.projector.color = projector.color;
-        this.css.projector.backgroundColor = projector.background_color;
-        this.css.projector.H1Color = projector.header_h1_color;
-        this.css.headerFooter.color = projector.header_font_color;
-        this.css.headerFooter.backgroundColor = projector.header_background_color;
-        this.updateCSS();
-    }
+        this.slides = this.projectorSubject
```
FinnStutzenstein, line 171: Now with switchMap

tsiegleauq (approved):
amazing!

PR #355, opened 2021-07-07
Adjusts PDF-creation to work

### LINKED ISSUE #203: Create PDF-documents
Currently, the generation of PDF-documents is in general deactivated.
Every action, which will generate a PDF-document, has to be checked.

Seen actions that create a PDF-document:

[Will be fixed in #355]
- [x] Motion-List: (Multiselect) Export as PDF
- [x] Motion-Detail: Export as PDF
- [x] Motion-Detail: Export PersonalNote as PDF
- [x] Agenda-Item-List: Export as PDF
- [x] User-List: Export user-list as PDF
- [x] Assignment-List: Export as PDF
- [x] Assignment-Detail: Export as PDF
- [x] User-Detail: Export as PDF
- [x] User-List: Export credentials as PDF

[Still open]

GabrielInTheWorld:
Fixes #203.

### DIFF
--- a/client/src/app/site/assignments/components/assignment-list/assignment-list.component.ts
+++ b/client/src/app/site/assignments/components/assignment-list/assignment-list.component.ts
@@ -110,7 +110,20 @@ export class AssignmentListComponent extends BaseListViewComponent<ViewAssignmen
             follow: [
                 {
                     idField: 'assignment_ids',
-                    follow: [SPEAKER_BUTTON_FOLLOW],
+                    follow: [
+                        SPEAKER_BUTTON_FOLLOW,
+                        // This makes PDF-export working.
+                        // Should be removed if the AU-service provides a "single-shot"-route.
+                        // See OpenSlides/openslides-autoupdate-service#260.
+                        {
+                            idField: 'candidate_ids',
+                            follow: [{ idField: 'user_id', fieldset: 'shortName' }]
+                        },
+                        {
+                            idField: 'poll_ids'
+                        }
+                        // End of block
+                    ],
                     fieldset: 'list'
                 }
             ],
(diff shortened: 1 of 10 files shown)


tsiegleauq (changes requested):
Missing PDFs:

motion, exportPersonalNote (motion-pdf-export-service)

Errors:
PDF of Assignment list missing candidates and the poll result is not present

I don't understand how exporting the reason works if a reason seems never to be requested by autoupdate before or during PDF generation

client/src/app/site/assignments/components/assignment-list/assignment-list.component.ts
```diff
@@ -110,7 +110,16 @@ export class AssignmentListComponent extends BaseListViewComponent<ViewAssignmen
...
+                    follow: [
+                        SPEAKER_BUTTON_FOLLOW,
+                        {
+                            idField: 'candidate_ids',
+                            follow: [{ idField: 'user_id', fieldset: 'shortName' }]
+                        },
+                        {
+                            idField: 'poll_ids'
+                        }
+                    ],
```
GabrielInTheWorld, line 122: @FinnStutzenstein: These fields are only requested for PDF-generation. Do you know a nicer way to request those fields only on demand?

tsiegleauq (approved):
[ ] - request fields on demand

client/src/app/site/assignments/components/assignment-list/assignment-list.component.ts
FinnStutzenstein: I would restructure this:
- You can create an own request for the PDF. Than it is encapsulated and maybe is put into the pdf-document-provider-service of each pdf.
- The AU service needs a one-shot route: It is like opening the autoupdate, get the initial data and close it afterwards. This could be currently done by the client, but a seperate route by the server would be good. So the server can send the initial data and close the connection afterwards.

GabrielInTheWorld: But how to send a request and wait for the initial data in the client? When does the client know, that the AU-service sent data?

FinnStutzenstein: If the AU service would provide a single-shot route, it is just a normal http request. I do not favor hacking this in the client.

GabrielInTheWorld: Then I will open an issue at the AU-service.

GabrielInTheWorld: Hmm, but I think we need in the client the logic "send AU-request -> wait for response -> close connection". I think, this feature would be more used than such a "single-shot"-route in the AU-service. The reason is, that we have to wait for the response...

GabrielInTheWorld:
I will merge this with the hint, that, when OpenSlides/openslides-autoupdate-service#260 is closed, the mentioned lines of code have to be adjusted.

PR #357, opened 2021-07-07
Recalculates if an operator has already voted

### LINKED ISSUE #336: "open poll" notification panel does not disappear after voting
The poll notification panel does not disappear after the user has send all their open votes

### LINKED ISSUE #337: List of open polls only shows wrong data
The list of open polls shows that a finished poll "has currently open polls",
every poll in the system was shown as open,
the state of the polls does not change.
Url is `/$/polls`

image

might correlate with #336

GabrielInTheWorld:
Fixes #336 
Fixes #337

### DIFF
--- a/client/src/app/shared/models/poll/poll.ts
+++ b/client/src/app/shared/models/poll/poll.ts
@@ -18,9 +18,17 @@ export class Poll extends BaseDecimalModel<Poll> {
     public votesinvalid: number;
     public votescast: number;
     public onehundred_percent_base: PollPercentBase;
+
+    /**
+     * TODO: these fields has to be autocalculated by the AU-service: OpenSlides/openslides-autoupdate-service#262
+     */
     public voted_id: number[];
     public user_has_voted: boolean;
     public user_has_voted_for_delegations: Id[];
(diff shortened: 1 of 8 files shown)


client/src/app/shared/models/poll/poll.ts
```diff
@@ -19,9 +19,17 @@ export class Poll extends BaseDecimalModel<Poll> {
...
+    /**
+     * START is this used anymore?
+     */
     public voted_id: number[];
     public user_has_voted: boolean;
     public user_has_voted_for_delegations: Id[];
+    /**
+     * END
+     */
+
```
GabrielInTheWorld, line 32: @FinnStutzenstein I've come across these lines. My question: When is this used and when are the fields filled? I have seen references for `user_has_voted` and `user_has_voted_for_delegations`, but I think these fields have to be filled by the client because they are client-specific fields, right? 
The field `voted_id` is an array of numbers and nowhere used. For me it is unclear in which cases this field should be used. Do you know any cases?

tsiegleauq (approved):

PR #374, opened 2021-07-12
If a user is superadmin, they can do everything

### LINKED ISSUE #363: superadmin who is not in meeting does not have all permission
If the superadmin was removed from a meeting, they can still access it yet cannot make changes since the permissions seem to be equal to "default". I suppose since the view user was missing from the operator even `this.hasOrganizationPermissions()` was not fired.

GabrielInTheWorld:
Fixes #363

### DIFF
--- a/client/src/app/core/core-services/operator.service.ts
+++ b/client/src/app/core/core-services/operator.service.ts
@@ -228,7 +228,6 @@ export class OperatorService {
 
                 this.OML = user.organization_management_level || null;
                 this.CML = getUserCML(user);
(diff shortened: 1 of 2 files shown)


FinnStutzenstein (changes requested):
`hasPerms` is good. BUt there are more places like `isInGroup`, which must be adjusted, too. E.g. is in group is used to check, whether the user can see a chat group - the superadmin always can see every chat group.

tsiegleauq (approved):

PR #380, opened 2021-07-14
Adds a filtering- and sorting-service for the member-list

### LINKED ISSUE #368: Sorting & filtering in MemberList
- [ ] filter by gender
- [ ] filter by meeting/committee

GabrielInTheWorld:
- Also includes filtering by committees/meetings
- Also makes the OsFilter interface generic

Fixes #368


client/src/app/management/services/member-filter.service.ts
```diff
@@ -0,0 +1,84 @@
...
+                ]
+            },
+            {
+                property: 'gender',
+                label: this.translate.instant('Gender'),
+                options: [
+                    ...genders.map(gender => ({ condition: gender, label: this.translate.instant(gender) })),
+                    { condition: '', label: this.translate.instant('No gender') }
+                ]
+            },
```
tsiegleauq, line 78: Remove.
This cannot be filtered but only sorted. The server currently accepts any string for gender, for which filtering will not work

tsiegleauq (changes requested):

client/src/app/management/services/member-filter.service.ts
GabrielInTheWorld: Then we should adjust the server to accept only an enum of genders to achieve this "feature". @emanuelschuetze mentioned, that we should filter by gender.

GabrielInTheWorld: For the first time, I have removed that lines. When OpenSlides/openslides-backend#847 is resolved these filters can be adjusted accordingly.

tsiegleauq: commenting them out and putting the issue in the comment would have been enough

tsiegleauq (approved):

PR #384, opened 2021-07-15
Update ICC integration

tsiegleauq:
Update notifyPath and publishPath

### DIFF
--- a/client/src/app/core/core-services/notify.service.ts
+++ b/client/src/app/core/core-services/notify.service.ts
@@ -75,12 +75,13 @@ interface ChannelIdResponse {
     channel_id: string;
 }
 
-const IccPath = '/system/icc';
+const notifyPath = '/system/icc/notify';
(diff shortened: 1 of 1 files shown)


client/src/app/core/core-services/notify.service.ts
```diff
@@ -75,12 +75,13 @@ interface ChannelIdResponse {
     channel_id: string;
 }
 
-const IccPath = '/system/icc';
+const notifyPath = '/system/icc/notify';
+const publishPath = `${notifyPath}/publish`;
 /**
  * Does currently not exist
  */
 // const iccHealthPath = `${IccPath}/health`
```
GabrielInTheWorld, line 83: Should this line not be removed?

tsiegleauq: no, just have to be adjusted when the health path exist

PR #385, opened 2021-07-15
First work for motions and poll slides

FinnStutzenstein:
Motions works with https://github.com/OpenSlides/openslides-autoupdate-service/pull/261 together: But CR's and CR's from amendments are missing in the server side. See my comment there.

Polls: Also awaiting change in https://github.com/OpenSlides/openslides-autoupdate-service/pull/252.

Assignments: There is a bug, too; candidates are currently `[Obejct object]` due to wrong server output (https://github.com/OpenSlides/openslides-autoupdate-service/issues/177).

With this PR it should work more than before. But I cannot say, if it is all bug-free since I didn't have the time to test everything (also the server bugs stopped me from more testing). If you want to merge this today it should be fine, but useless without the server changes. Otherwise I (or someone else) can continue, if there are updates for the projector service.


client/src/app/slides/assignment/assignment-slide.component.html
```diff
@@ -1,21 +1,21 @@
-<div *ngIf="data">
+<div *ngIf="data?.data">
     <div class="slidetitle">
         <h1>{{ data.data.title }}</h1>
         <h2>{{ 'Election' | translate }}</h2>
     </div>
 
-    <div *ngIf="data.data && data.data.description" [innerHTML]="data.data.description | trust: 'html'"></div>
```
tsiegleauq, line 7: `*ngIf="data.data?.description"`

tsiegleauq (approved):
tested:
- [x] projecting motions
- [x] motions with CR
- [x] amendments
- [x] motions with accepted amendments
The motion slides seems to ignore the current line-numbering config, not sure if this was only related to the client tho.

what does not work
- motion poll slide
- assignment poll sllide

aproved since I figured they are not really supposed to work for now

tsiegleauq:
sufficient for the time being. Line numbers is an issue and I suppose polls are still worked on.

PR #386, opened 2021-07-15
Fix dynamically loaded choices in settings

jsangmeister:
It seems that often the client does not load the meeting settings and just displays the default. Seems to be a timing issue (happens when I reload on a specific settings page). For other settings, it happens sometimes, for these lists (e.g. default workflow) it happens always, it seems. @tsiegleauq maybe you can check if this is reproducible and what's the reason behind this. Saving works now, anyway.


tsiegleauq:
can confirm, saving with that change. Loading them reliable is then another issue

tsiegleauq (approved):

PR #392, opened 2021-07-16
Remove unused dockerfiles

tsiegleauq:
removes a set of dockerfiles and an nginx.conf that was apparently
unused since os3+


tsiegleauq:
@FinnStutzenstein I suppose these were artifacts from the old day?

FinnStutzenstein (approved):
Yes, maybe I forgot to delete them on the last rebase

PR #409, opened 2021-07-22
Add manager_ids to committee edit

tsiegleauq:
Support manager_ids for the committee objects

### DIFF
--- a/client/src/app/shared/models/event-management/committee.ts
+++ b/client/src/app/shared/models/event-management/committee.ts
@@ -12,6 +12,7 @@ export class Committee extends BaseModel<Committee> {
     public default_meeting_id: Id; // meeting/default_meeting_for_committee_id;
     public template_meeting_id: Id; // meeting/template_meeting_for_committee_id;
     public user_ids: Id[]; // (user/committee_ids)[];
+    public manager_ids: Id[]; // (user/manager_ids)[];
     public forward_to_committee_ids: Id[]; // (committee/receive_forwardings_from_committee_ids)[];
     public receive_forwardings_from_committee_ids: Id[]; // (committee/forward_to_committee_ids)[];
     public organization_id: Id; // organization/committee_ids;
--- a/client/src/app/core/actions/committee-action.ts
+++ b/client/src/app/core/actions/committee-action.ts
@@ -10,6 +10,7 @@ export namespace CommitteeAction {
         description?: UnsafeHtml;
         // For UPDATE: Required permission: OML.can_manage_organization
         user_ids?: Id[];
+        manager_ids?: Id[];
         // For UPDATE: Required permission: CML.can_manage || OML.can_manage_organization
         organization_tag_ids?: Id[];
         // Required permission: OML.can_manage_organization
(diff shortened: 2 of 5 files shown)


tsiegleauq:
committee-edit-components `patchForm` function was executed twice, while at the second call the `user_ids` got altered. I have no idea how to add the `manager_ids` now, adding the `user_ids` alone seems totally oblivious.

image

tsiegleauq:
requires https://github.com/OpenSlides/openslides-backend/pull/860

emanuelschuetze:
tested with current backend-master: committee manager is not set in committee update view (committee create view works).

client/src/app/shared/models/event-management/committee.ts
```diff
@@ -12,6 +12,7 @@ export class Committee extends BaseModel<Committee> {
     public default_meeting_id: Id; // meeting/default_meeting_for_committee_id;
     public template_meeting_id: Id; // meeting/template_meeting_for_committee_id;
     public user_ids: Id[]; // (user/committee_ids)[];
+    public manager_ids: Id[]; // (user/manager_ids)[];
```
GabrielInTheWorld, line 15: This indicates a relation, which is right, but that relation is not registered in the `repositories/relation.ts`-file.
Also there are three things missing: the relation back from users to committees and the relations in the view-*-files of them.

GabrielInTheWorld: A second thing is, that the `manager_ids` has to be requested from the AU-service. So, the field has to be specified in the repo to the fieldsets. (You can follow them, too, to get request a specific fieldset, like only the `shortName`)

GabrielInTheWorld (changes requested):
There are three things I have noticed:

1. Some relations are missing
2. The `manager_ids` have to be requested from the AU-service
3. The managers are not stored in the database
I sent this payload:
```json
[
  {
    "action":"committee.create",
    "data":[
      {
        "name":"Rote Beete #2",
        "organization_id":1,
        "description":"",
        "organization_tag_ids":[
          
        ],
        "user_ids":[
          
        ],
        "manager_ids":[
          1,
          4
        ],
        "forward_to_committee_ids":[
          
        ],
        "receive_forwardings_from_committee_ids":[
          
        ]
      }
    ]
  }
]
```

And this is stored in the database:
```json
{
  "id":3,
  "name":"Rote Beete #2",
  "user_ids":[
    1,
    4
  ],
  "description":"",
  "meta_deleted":false,
  "meta_position":46,
  "organization_id":1,
  "organization_tag_ids":[
    
  ],
  "forward_to_committee_ids":[
    
  ],
  "receive_forwardings_from_committee_ids":[
    
  ]
}
```

client/src/app/shared/models/event-management/committee.ts
FinnStutzenstein: This is not a database field! You cannot add non-database fields here

GabrielInTheWorld: Then we have to change this in the database

tsiegleauq: most likely a design flaw, but we can get around that for now.

tsiegleauq:
extra ping due to high priority @emanuelschuetze @GabrielInTheWorld @FinnStutzenstein

emanuelschuetze (approved):

GabrielInTheWorld (approved):

PR #418, opened 2021-07-27
Adds restrictions to orga.update

### LINKED ISSUE #417: Organization. update
update organization's name has to be possible with OML-right. Currently an error message demand superadmin right

GabrielInTheWorld:
Only superadmins can modify enable_electronic_voting and reset_password_verbose_errors

Fixes #417


emanuelschuetze (approved):

FinnStutzenstein:
> reset_password_verbose_errors

This is a mistake in the action description. Currently only the electronic voting is for the superadmin. the former can be set by orga managers (so it must be in group A instead of B). See the split in the models.yml: https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/models.yml#L69

tsiegleauq (approved):

tsiegleauq (approved):

PR #419, opened 2021-07-27
Fixes setting the presence of users

### LINKED ISSUE #395: user bulk set present broken
No options, error is thrown on OK.

```
Uncaught (in promise): TypeError: Cannot read property '0' of undefined
TypeError: Cannot read property '0' of undefined
    at UserListComponent.<anonymous> (user-list.component.ts:403)
```

image

GabrielInTheWorld:
- Also changes the "setPresent"-function in the user-repo so, that a list of users can be passed

Fixes #395

### DIFF
--- a/client/src/app/site/users/components/user-list/user-list.component.ts
+++ b/client/src/app/site/users/components/user-list/user-list.component.ts
@@ -375,34 +375,16 @@ export class UserListComponent extends BaseListViewComponent<ViewUser> implement
         }
     }
 
-    /**
-     * Handler for bulk setting/unsetting the 'active' attribute.
-     * Uses selectedRows defined via multiSelect mode.
-     */
-    public async setStateSelected(field: UserStateField): Promise<void> {
-        let options: [string, string];
-        let verboseStateName: string;
-        switch (field) {
-            case 'is_active':
-                options = [_('active'), _('inactive')];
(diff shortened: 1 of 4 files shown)


client/src/app/site/users/components/user-list/user-list.component.ts
```diff
@@ -441,7 +423,41 @@ export class UserListComponent extends BaseListViewComponent<ViewUser> implement
...
+                options = [_('active'), _('inactive')];
+                verboseStateName = 'active';
+                break;
+            case 'is_present_in_meetings':
+                options = [_('present'), _('absent')];
+                verboseStateName = 'present';
+                break;
+            case 'is_physical_person':
+                options = [_('natural person'), _('no natural person')]; // switched order: no committee=physical person
+                verboseStateName = 'committee';
```
FinnStutzenstein, line 448: can you also rename `committee` here, so all logical connections to OS3 are gone?

FinnStutzenstein (approved):
Code looks good!

client/src/app/site/users/components/user-list/user-list.component.ts
GabrielInTheWorld: I just removed the `verboseStateName` since it is never used.

PR #425, opened 2021-07-28
Improve Projector

FinnStutzenstein:
- fixed the presentation controls
- hide presentation controls, if no mediafile is projected
- no need for the cd.detect changes with the timer
- use observables for the projector detail component

### DIFF
--- a/client/src/app/shared/components/slide-container/slide-container.component.ts
+++ b/client/src/app/shared/components/slide-container/slide-container.component.ts
@@ -209,6 +209,7 @@ export class SlideContainerComponent extends BaseComponent {
     private setDataForComponent(): void {
         if (this.slideRef && this.slideRef.instance) {
             this.slideRef.instance.data = this.slideData;
+            this.slideRef.changeDetectorRef.detectChanges();
         }
     }
 
--- a/client/src/app/site/projector/components/projector-detail/projector-detail.component.ts
+++ b/client/src/app/site/projector/components/projector-detail/projector-detail.component.ts
@@ -1,14 +1,14 @@
 import { CdkDragDrop, moveItemInArray } from '@angular/cdk/drag-drop';
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component, OnInit } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { ActivatedRoute } from '@angular/router';
 
-import { timer } from 'rxjs';
+import { Observable } from 'rxjs';
+import { map, mergeMap, switchMap } from 'rxjs/operators';
 
 import { ScrollScaleDirection } from 'app/core/actions/projector-action';
 import { ProjectorMessageAction } from 'app/core/actions/projector-message-action';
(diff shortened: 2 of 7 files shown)


client/src/app/shared/components/slide-container/slide-container.component.ts
```diff
@@ -209,6 +209,7 @@ export class SlideContainerComponent extends BaseComponent {
     private setDataForComponent(): void {
         if (this.slideRef && this.slideRef.instance) {
             this.slideRef.instance.data = this.slideData;
+            this.slideRef.changeDetectorRef.detectChanges();
```
tsiegleauq, line 212: don't expose `changeDetectorRef` publicly. It's cleaner to use a function in `slideRef` than to change methods here.

However: This should not be necessary, if all changes to inner Slides are observed already, no?

client/src/app/site/projector/components/projector-detail/projector-detail.component.ts
```diff
@@ -112,30 +113,41 @@ export class ProjectorDetailComponent extends BaseModelContextComponent implemen
...
+                                    }
+                                ]
+                            },
+                            PROJECTOR_CONTENT_FOLLOW
+                        ]
+                    });
                 })
-            );
+            )
+            .subscribe();
```
tsiegleauq, line 140: seems this will never get cleaned, push into subscription list

client/src/app/site/projector/components/projector-detail/projector-detail.component.ts
```diff
@@ -112,30 +113,41 @@ export class ProjectorDetailComponent extends BaseModelContextComponent implemen
...
+
+        this.projectorObservable = projectorId$.pipe(
+            switchMap(projectorId => this.projectorRepo.getViewModelObservable(projectorId))
+        );
+        this.projectorObservable.subscribe(projector => {
+            if (projector) {
+                const title = projector.name;
+                super.setTitle(title);
+                this.projector = projector;
+            }
```
tsiegleauq, line 150: same as above, this sub stays open

tsiegleauq (changes requested):

emanuelschuetze:
Tested: Countdown does not work correctly in projector detail view. Looks like change detection is not correct because not every second the projector is updated.

client/src/app/shared/components/slide-container/slide-container.component.ts
FinnStutzenstein: `slideRef` has the type `ComponentRef` which explicitly exposes the CDRef: https://angular.io/api/core/ComponentRef

I do not know the internals too much, but I guess that in this case the normal change detection is circumvented. As far as I looked into angulars source code, `@Input()` is just a marker and does not trigger the CD when the value is changed.

FinnStutzenstein:
Countdowns should work now (note that the time for countdowns on projectors may be very wrong, see https://github.com/OpenSlides/openslides-autoupdate-service/issues/283 for this issue)

emanuelschuetze (approved):

tsiegleauq (approved):

PR #435, opened 2021-07-29
Normal users can send motions

tsiegleauq:
creates the motion form group by permissions


tsiegleauq:
@FinnStutzenstein @GabrielInTheWorld started to strictly type formGroups, yet I needed to alter some payloads a little. Since the clients sends everything nevertheless it should not matter.

GabrielInTheWorld (approved):

emanuelschuetze (changes requested):
Tested:
* open motion create form as delegate use: `TypeError: Cannot read property 'setValue' of null`
* enter motion title and motion text (reason is empty): SAVE button is disabled (reason required which is wrong)

PR #443, opened 2021-08-02
Add scripts to generate settings defaults & permissions from .yml files

jsangmeister:
What's been done:
- unified naming of permission enum with the respective permissions (for consistency & easier generation) - this generated a lot of the changed lines
- add script to generate `meeting-settings-defaults.ts` where for each setting key the default is stored if one exists
- moved validation of defaults from runtime to generation time to notice errors earlier and prevent non-startable states
- add script to generate `permission-children.ts` where for each permission the list of children (= implicitly given permissions) is stored
- use the above mentioned lists to correctly calculate the actual permissions for the operator
- use the above mentioned lists to correctly display the permission matrix - implicitly given permission are checked and greyed out


jsangmeister:
Needs https://github.com/OpenSlides/OpenSlides/pull/6194 to work correctly. Still WIP, but the existing parts can already be reviewed - permission generation will probably look similar structure-wise.

client/src/app/app/core/repositories/motions/motion-format.service.spec.ts
```diff
@@ -1,16 +0,0 @@
-import { TestBed } from '@angular/core/testing';
```
jsangmeister, line 1: Because of the double `/app/app` I'm guessing this is a rebase relict and deleted it - correct me if I'm wrong

client/package.json
```diff
@@ -31,7 +31,8 @@
         "prettify-check": "prettier --config ./.prettierrc --list-different \"src/{app,environments}/**/*{.ts,.js,.json,.css,.scss}\"",
         "prettify-write": "prettier --config ./.prettierrc --write \"src/{app,environments}/**/*{.ts,.js,.json,.css,.scss}\"",
         "cleanup": "npm run prettify-write; npm run lint-write",
-        "cleanup-win": "npm run prettify-write & npm run lint-write"
+        "cleanup-win": "npm run prettify-write & npm run lint-write",
+        "generate-defaults": "ts-node -s -r tsconfig-paths/register cli/generate-settings-defaults.ts && npm run prettify-write"
```
jsangmeister, line 35: I was not sure whether to automatically run prettify after generation, I'm open to removing it again if someone does not want to have it here

jsangmeister:
Done, ready for review

client/package.json
tsiegleauq: yes, please remove. You could import prettier in your `.ts` files and run it after generating a file directly after file generation only, but right now you would lint and write everything twice if generating both files, which seems unnecessary.

client/src/app/core/core-services/operator.service.ts
```diff
@@ -374,22 +375,27 @@ export class OperatorService {
...
+        const permissionSet = new Set<Permission>();
+        if (this.isAnonymous) {
+            // Anonymous is always in the default group.
+            this.activeMeeting?.default_group?.permissions.forEach(perm => permissionSet.add(perm));
         } else {
-            const permissionSet = new Set<string>();
-            this.DS.getMany(Group, this.groupIds).forEach(group => {
-                group.permissions?.forEach(permission => {
-                    permissionSet.add(permission);
+            if (this.groupIds) {
```
tsiegleauq, line 383: `group_ids(meetingId?: Id): Id[]` (ViewUser)
will in doubt return some array, where `!![]` is true.

Replace with: `this.groupIds?.length`

tsiegleauq (approved):
Code seems good so far. Tested pretty much everything I suppose, working perfectly. Good job. One stylistic remark and reconsider prettify after generation over npm

client/src/app/core/core-services/operator.service.ts
jsangmeister: Sure, although even for `[]` the code should still work since `forEach` will just not execute ;)

jsangmeister:
Also updated cleanup scripts to include the cli folder.

tsiegleauq:
ok. please rebase and merge.

tsiegleauq (approved):

PR #445, opened 2021-08-03
Updates adding and removing groups in meeting.update

### LINKED ISSUE #422: Operators of users without OML.superadmin will be overwritten
Apparently, if an update from the AU-service is received by the client, the client removes the current operator from its memory. Therefore, the operator's user loses they permissions.

This is for example the case if the operator creates a motion without any submitters (the operator's user is then the only submitter).

### LINKED ISSUE #432: After editing a meeting, the operator has lost their permissions
1. Go to a committee and then edit a meeting.
2. Add members to the meeting and save.
3. Go back. 

Current behavior:
No meetings and no committees are seen.

### LINKED ISSUE #440: Clean up adding/removing users and admins to/from a meeting
This is not so beautiful I wanted, but for now it fulfills its purpose and I will clean up it later.

_Originally posted by @GabrielInTheWorld in https://github.com/OpenSlides/openslides-client/pull/434#discussion_r679919806_

GabrielInTheWorld:
- When updating a meeting, groups are removed or added according to the original
group-ids-array of a user instead of mapping them to a user-id
- The operator's OML won't be overwritten if it is "undefined"

Fixes #422 
Fixes #432 
Fixes #440 

It seems, that the operator lost its permissions because new user-data came from an AU-response, which does not include the OML for the operator.

### DIFF
--- a/client/src/app/core/repositories/management/meeting-repository.service.ts
+++ b/client/src/app/core/repositories/management/meeting-repository.service.ts
@@ -13,6 +13,7 @@ import { ViewMeeting } from 'app/management/models/view-meeting';
 import { Identifiable } from 'app/shared/models/base/identifiable';
 import { Meeting } from 'app/shared/models/event-management/meeting';
 import { Projection } from 'app/shared/models/projector/projection';
+import { ViewUser } from 'app/site/users/models/view-user';
 import { BaseRepository } from '../base-repository';
 import { RepositoryServiceCollectorWithoutActiveMeetingService } from '../repository-service-collector-without-active-meeting-service';
 
@@ -26,11 +27,11 @@ export interface ImportMeeting {
     [collection: string]: unknown[];
 }
 
-interface MeetingUserModifiedFields {
-    addedUserIds?: Id[];
-    removedUserIds?: Id[];
-    addedAdminIds?: Id[];
-    removedAdminIds?: Id[];
+export interface MeetingUserModifiedFields {
+    addedUsers?: ViewUser[];
+    removedUsers?: ViewUser[];
+    addedAdmins?: ViewUser[];
+    removedAdmins?: ViewUser[];
 }
 
 @Injectable({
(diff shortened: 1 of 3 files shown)


client/src/app/core/repositories/management/meeting-repository.service.ts
```diff
@@ -169,53 +170,30 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
...
-                userId =>
-                    !meeting.user_ids.includes(userId) &&
-                    !addedUserIds.includes(userId) &&
-                    !removedUserIds.includes(userId)
-            );
-            updateData.push(
-                ...this.getUserGroupUpdatePayload(remainingRemovedAdminIds, meeting.id, meeting.default_group_id)
-            );
-            updateData.push(...this.getUserGroupUpdatePayload(completelyRemovedAdminIds, meeting.id));
+        let userUpdate: { [userId: number]: Id[] } = {};
```
FinnStutzenstein, line 173: Please add a comment, that this is a mapping of user ids to group ids per user.

client/src/app/core/repositories/management/meeting-repository.service.ts
```diff
@@ -249,14 +227,35 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
...
         meetingId: Id,
-        groupId?: Id
-    ): { id: Id; group_$_ids: { [meetingId: number]: Id[] } }[] {
-        return ids.map(id => ({
-            id,
-            group_$_ids: { [meetingId]: groupId ? [groupId] : [] }
-        }));
+        groupId: Id
+    ): { [userId: number]: Id[] } {
+        const nextData = { ...previousData };
```
FinnStutzenstein, line 248: why do you always copy? You can do this as an side effect (so no return)

client/src/app/core/repositories/management/meeting-repository.service.ts
```diff
@@ -249,14 +227,35 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
...
+        const getNextGroupIds = (groupIds: Id[]) => {
```
FinnStutzenstein, line 249: please move this into the arrow fn.

FinnStutzenstein (approved):
Looks good, some cleanup needed.

client/src/app/core/repositories/management/meeting-repository.service.ts
GabrielInTheWorld: Yes, it is because I don't like side effects so much... But I can change it.

GabrielInTheWorld:
I updated the current state: Added a comment and manipulate the user-group-ids-map as a side effect.
I will merge it, now.

PR #446, opened 2021-08-03
Poll Slide

FinnStutzenstein:
Unified the usage of `PollData`: The poll services and
display-components only work with PollData, not ViewPoll. ViewPoll as
well as the data for the projector interfaces with PollData, so PollData
is the only single interface to operate on.

Also cleaned up the assignment/motion-poll service; Since there is just
one poll model in OS4, the data is much more streamedlined, so much code
could be moved into the poll service.

The slide for analog polls works; Maybe something else broke, but this
PR is more of a cleanup-type. Building upon this PR, the bugs/missing
logic can be fixed/implemented in a more clean fashion.

### DIFF
--- a/client/src/app/site/assignments/modules/assignment-poll/components/assignment-poll/assignment-poll.component.ts
+++ b/client/src/app/site/assignments/modules/assignment-poll/components/assignment-poll/assignment-poll.component.ts
@@ -10,14 +10,11 @@ import { ComponentServiceCollector } from 'app/core/ui-services/component-servic
 import { PromptService } from 'app/core/ui-services/prompt.service';
 import { VotingService } from 'app/core/ui-services/voting.service';
 import { VotingPrivacyWarningComponent } from 'app/shared/components/voting-privacy-warning/voting-privacy-warning.component';
-import { PollState } from 'app/shared/models/poll/poll-constants';
-import { ViewPoll } from 'app/shared/models/poll/view-poll';
 import { infoDialogSettings } from 'app/shared/utils/dialog-settings';
 import { ViewAssignment } from 'app/site/assignments/models/view-assignment';
 import { BasePollComponent } from 'app/site/polls/components/base-poll.component';
 import { AssignmentPollDialogService } from '../../services/assignment-poll-dialog.service';
 import { AssignmentPollPdfService } from '../../services/assignment-poll-pdf.service';
-import { AssignmentPollService } from '../../services/assignment-poll.service';
 
 /**
  * Component for a single assignment poll. Used in assignment detail view
@@ -79,7 +76,6 @@ export class AssignmentPollComponent extends BasePollComponent<ViewAssignment> i
         choiceService: ChoiceService,
         repo: PollRepositoryService,
         pollDialog: AssignmentPollDialogService,
-        private pollService: AssignmentPollService,
         private formBuilder: FormBuilder,
         private pdfService: AssignmentPollPdfService,
         private operator: OperatorService,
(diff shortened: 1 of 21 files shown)


client/src/app/shared/models/poll/poll-constants.ts
```diff
@@ -1,5 +1,39 @@
...
+    getContentObjectTitle: () => string | null;
+}
+
+export interface OptionTitle {
+    title: string;
+    subtitle?: string;
+}
+
+export interface OptionData {
+    // content_object?: {};
```
FinnStutzenstein, line 28: todo: remove

client/src/app/site/assignments/modules/assignment-poll/components/assignment-poll/assignment-poll.component.ts
```diff
@@ -109,13 +105,4 @@ export class AssignmentPollComponent extends BasePollComponent<ViewAssignment> i
...
         this.dialog.open(VotingPrivacyWarningComponent, infoDialogSettings);
     }
-
-    protected onAfterUpdatePoll(poll: ViewPoll): void {
-        if (poll.state !== PollState.Finished && poll.state !== PollState.Published) {
-            return;
-        }
-        this.candidatesLabels = this.pollService.getChartLabels(poll);
-        const chartData = this.pollService.generateChartData(poll);
-        this.chartDataSubject.next(chartData);
```
FinnStutzenstein, line 119: the `chartDataSubject` was never used -> removed.

client/src/app/slides/poll/poll-slide.component.ts
```diff
@@ -25,44 +24,108 @@ export class PollSlideComponent extends BaseSlideComponent<PollSlideData> {
...
+
+    public polldata: PollData;
+
+    public constructor(public pollService: PollService, private collectionMapperService: CollectionMapperService) {
         super();
     }
 
     protected setData(value: SlideData<PollSlideData>): void {
         super.setData(value);
+        console.log(value);
```
emanuelschuetze, line 40: remove debug output?

client/src/app/slides/poll/poll-slide.component.ts
```diff
@@ -25,44 +24,108 @@ export class PollSlideComponent extends BaseSlideComponent<PollSlideData> {
...
+            options: data.options.map((option, i) => this.createOptionData(option, i + 1)),
+            global_option: data.global_option
+                ? this.createOptionData(data.global_option)
+                : ({ getOptionTitle: () => ({ title: '' }) } as OptionData)
+        };
+        return poll;
+    }
+
+    private createOptionData(data: SlideOption, weight: number = 1): OptionData {
+        console.log(data);
```
emanuelschuetze, line 113: remove?

client/src/app/shared/models/poll/poll-constants.ts
```diff
@@ -1,5 +1,39 @@
...
+    getOptionTitle: () => OptionTitle;
+    // text?: string;
```
GabrielInTheWorld, line 30: What about this?

client/src/app/shared/models/poll/poll-constants.ts
```diff
@@ -1,5 +1,39 @@
...
+    yes?: number;
+    no?: number;
+    abstain?: number;
+    weight?: number;
+}
```
GabrielInTheWorld, line 35: I know, there are some other interfaces in this file, but I think, that poll-**constants**.ts is not the right place for interfaces. Maybe you could put these interfaces into an own file.

GabrielInTheWorld (approved):
2/3 things, rest looks good!

GabrielInTheWorld:
Wahlergebnisdetails
 I have seen these results in an election.
Maybe you could fix that behavior in this PR or open an own issue for that.

There is also another thing: If there is an election with a minimum number of votes greater than the number of persons, who can be voted for, I cannot send my votes. I don't know if this is a bug...

FinnStutzenstein:
> open an own issue for that.

#451 

> There is also another thing

#452

PR #453, opened 2021-08-04
Mediafiles will be directly requested from the media-service via GET

### LINKED ISSUE #390: Load mediafiles on demand with "normal" image-tag
Currently, a client authenticates itself by a token in an authentication-header in every HTTP-request. The disadvantage here is, that some HTML-tag-attributes (like the `src`-attribute of an ``-tag) do not include the authentication-header in their requests.
We need an approach to deal with this.

GabrielInTheWorld:
With this PR the URL for a mediafile is public, so users can use them to paste it in any text-document.

Fixes #390


FinnStutzenstein (approved):
Looks good

PR #460, opened 2021-08-09
Fixes about me and race conditions in user-detail-view

### LINKED ISSUE #364: "About me" and "vote delegated to/from" do not work in the user-detail-view

GabrielInTheWorld:
- The template-field "about_me_$" is requested
- Prevents a race condition in the user-detail-view: A function to request a user's information was invoked two times independently from each other.

Fixes #364

### DIFF
--- a/client/src/app/site/users/components/user-detail/user-detail.component.ts
+++ b/client/src/app/site/users/components/user-detail/user-detail.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnDestroy } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, combineLatest } from 'rxjs';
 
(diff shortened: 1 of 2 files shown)


client/src/app/site/users/components/user-detail/user-detail.component.ts
```diff
@@ -141,6 +142,11 @@ export class UserDetailComponent extends BaseModelContextComponent implements On
     }
 
     private loadUserById(): void {
+        // Prevent race conditions since this method is invoked two times.
```
FinnStutzenstein, line 145: This is not the solution. Use observables. Some lines above are two calls to this function. You can merge both observables. Also awaiting this.request models can prevent the race condition (do it aditionally)

FinnStutzenstein (changes requested):

FinnStutzenstein (approved):
Looks good, not tested

PR #462, opened 2021-08-10
History list

FinnStutzenstein:
Rework the history list. Works with https://github.com/OpenSlides/openslides-backend/pull/898

I additionally cleaned up the history and time-travel service. The
actual "entering history" still has to be done. This PR only focusses on
the history list as a first step.


client/src/app/core/core-services/time-travel.service.ts
```diff
@@ -26,70 +13,25 @@ interface HistoryData {
...
-    /**
-     * Clears the DataStore and stops the WebSocket connection
-     */
-    private async stopTime(history: History): Promise<void> {
-        // await this.webSocketService.close();
-        await this.DS.set(); // Same as clear, but not persistent.
-        this.historyService.enterHistoryMode(history);
+    private async stopTime(position: Position): Promise<void> {
+        this.historyService.enterHistoryMode(position);
     }
```
GabrielInTheWorld, line 36: I think the `async`-keyword is an artifact, because the function `enterHistoryMode(...)` is not an asynchronous function. So, it can be removed.

client/src/app/core/core-services/time-travel.service.ts
```diff
@@ -26,70 +13,25 @@ interface HistoryData {
...
-     * Leaves the history mode. Just restart OpenSlides:
-     * The active user is checked, a new WS connection established and
-     * all missed auto updates are requested.
+     * Leaves the history mode.
      */
     public async resumeTime(): Promise<void> {
         this.historyService.leaveHistoryMode();
-        await this.DS.set();
-        await this.lifecycleService.reboot();
     }
```
GabrielInTheWorld, line 32: I think, these two functions can be a "synchronous" function (so remove the `async`-keyword) or is there a benefit, when they are asynchronous?

client/src/app/site/history/components/history-list/history-list.component.ts
```diff
@@ -31,75 +32,59 @@ import { ViewUser } from 'app/site/users/models/view-user';
...
-    public modelSelectForm: FormGroup;
+    public motionSelectForm: FormGroup;
 
     /**
-     * The observer for the selected collection, which is currently hardcoded
-     * to motions.
+     * The observer for the all motions
      */
-    public collectionObserver: BehaviorSubject<BaseViewModel[]>;
+    public motions: BehaviorSubject<BaseViewModel[]>;
```
GabrielInTheWorld, line 54: This variable remains exclusively an instance of `BehaviorSubject<ViewMotion[]>`, doesn't it? Then, declare the type of the variable from `BehaviorSubject<BaseViewModel[]>` to `BehaviorSubject<ViewMotion[]>`

GabrielInTheWorld: Why is here a `BehaviorSubject` used?

GabrielInTheWorld (changes requested):
Looks good. Some cleanup is needed, then this PR is fine.

GabrielInTheWorld (approved):
Looks good. Not tested,

PR #466, opened 2021-08-13
Adjusts tree-lists to the new nested modelling from the backend

### LINKED ISSUE #347: Check all tree lists (agenda, category, call list) when the sorting is fixed in the backend

GabrielInTheWorld:
Remakes the flat-nodes as proxies to have full access to the underlying item without access the 'item' previously.

Fixes #347 

@FinnStutzenstein Please review. I have thought about how I can implement the flat tree-like data from the backend and I came up with this solution. I rewrote the existing `FlatNode` to a proxy to have full access to the properties of an underlying model without accessing the `item`-property previously. That avoid creating a third datastructure, which fulfills such functionalities.

### DIFF
--- a/client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
+++ b/client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
@@ -4,7 +4,7 @@ import { AgendaItemAction } from 'app/core/actions/agenda-item-action';
 import { DEFAULT_FIELDSET, Fieldsets } from 'app/core/core-services/model-request-builder.service';
 import { Id } from 'app/core/definitions/key-types';
 import { MeetingSettingsService } from 'app/core/ui-services/meeting-settings.service';
-import { TreeIdNode } from 'app/core/ui-services/tree.service';
+import { TreeIdNode, TreeService } from 'app/core/ui-services/tree.service';
 import { AgendaItem, AgendaItemType } from 'app/shared/models/agenda/agenda-item';
 import { Identifiable } from 'app/shared/models/base/identifiable';
 import { HasAgendaItem, ViewAgendaItem } from 'app/site/agenda/models/view-agenda-item';
@@ -28,11 +28,12 @@ export interface AgendaListTitle {
 export class AgendaItemRepositoryService extends BaseRepositoryWithActiveMeeting<ViewAgendaItem, AgendaItem> {
     public constructor(
         repositoryServiceCollector: RepositoryServiceCollector,
-        private meetingSettingsService: MeetingSettingsService
+        private meetingSettingsService: MeetingSettingsService,
+        private treeService: TreeService
     ) {
         super(repositoryServiceCollector, AgendaItem);
 
-        this.setSortFunction((a, b) => a.weight - b.weight);
+        this.setSortFunction((a, b) => a.tree_weight - b.tree_weight); // leave the sorting as it is
     }
 
     public getFieldsets(): Fieldsets<AgendaItem> {
@@ -193,5 +194,9 @@ export class AgendaItemRepositoryService extends BaseRepositoryWithActiveMeeting
     public bulkSetAgendaType(items: ViewAgendaItem[], agendaType: AgendaItemType): Promise<void> {
         const payload: AgendaItemAction.UpdatePayload[] = items.map(item => ({ id: item.id, type: agendaType }));
         return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
+    }
+
+    protected tapViewModels(viewModels: ViewAgendaItem[]): void {
+        this.treeService.injectFlatNodeInformation(viewModels, 'weight', 'parent_id');
     }
 }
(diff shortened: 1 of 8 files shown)


client/src/app/core/ui-services/tree.service.ts
```diff
@@ -198,25 +208,6 @@ export class TreeService {
...
-        const item = items.find(i => branch.item.id === i.id);
-        if (item) {
-            return [item];
-        } else if (!branch.children) {
-            return [];
-        } else {
-            return [].concat(...branch.children.map(child => this.getTopItemsFromBranch(child, items)));
-        }
-    }
-
```
GabrielInTheWorld, line 219: I have noticed, that this function is not used anymore.

client/src/app/core/repositories/base-repository.ts
```diff
@@ -403,6 +397,10 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
         this.modifiedIdsSubject.next(modelIds);
     }
 
+    protected tapViewModels(viewModels: V[]): V[] {
```
FinnStutzenstein, line 400: Must return void since it is only a tapping.

client/src/app/core/ui-services/tree.service.ts
```diff
@@ -349,29 +340,46 @@ export class TreeService {
...
 
+    private createFlatNode<T extends Identifiable & Displayable>(item: OSTreeNode<T>, level: number): FlatNode<T> {
+        const children = item.children;
+        const node: any = {
+            id: item.id,
+            item: item.item,
+            expandable: !!children,
+            isExpanded: !!children,
+            level: level,
+            isSeen: true
```
FinnStutzenstein, line 351: is this written somewhere? If so, whats about `set` in the Proxy?

client/src/app/shared/overload-js-functions.ts
```diff
@@ -117,6 +123,13 @@ function overloadArrayFunctions(): void {
         },
         enumerable: false
     });
+
+    Object.defineProperty(Array.prototype, 'tap', {
+        value: function <T>(callbackFn: (self: T[]) => T[]): T[] {
+            return callbackFn(this);
+        },
```
FinnStutzenstein, line 130:
```suggestion
        value: function <T>(callbackFn: (self: T[]) => void): T[] {
            callbackFn(this);
            return this;
        },
```

Currently, if you tap callback returns something else, it will replace the list. But you just want tapping without changing the array itself.

FinnStutzenstein (commented):
Looks good so far!

client/src/app/core/repositories/base-repository.ts
GabrielInTheWorld: Hmm... Then I will change the name. I want a function, that modifies the incoming value, because I use it in `AgendaItemRepository.service#200`. Maybe `pipe...`.

client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
```diff
@@ -194,4 +195,8 @@ export class AgendaItemRepositoryService extends BaseRepositoryWithActiveMeeting
         const payload: AgendaItemAction.UpdatePayload[] = items.map(item => ({ id: item.id, type: agendaType }));
         return this.sendBulkActionToBackend(AgendaItemAction.UPDATE, payload);
     }
+
+    protected pipeViewModels(viewModels: ViewAgendaItem[]): ViewAgendaItem[] {
+        return this.treeService.makeFlatTree(viewModels, 'weight', 'parent_id');
```
FinnStutzenstein, line 200: I'm not sure why you need to wrap the models here. You can just set a new property `tree_weight` and `level` according to the sorting. This should be sufficient.

GabrielInTheWorld: Yes, but I can use the `treeService.makeFlatTree`-function like I did it here, if I can modify the original array...

FinnStutzenstein: > if I can modify the original array

GabrielInTheWorld: Yes. I have wanted to use the existing functions.

client/src/app/shared/models/agenda/agenda-item.ts
```diff
@@ -36,6 +36,7 @@ export class AgendaItem extends BaseModel<AgendaItem> {
     public duration: number; // in seconds
     public weight: number;
     public level: number;
+    public position: number;
```
GabrielInTheWorld, line 39: I don't know if this is the right place for this information.

client/src/app/shared/models/agenda/agenda-item.ts
```diff
@@ -36,6 +36,7 @@ export class AgendaItem extends BaseModel<AgendaItem> {
     public duration: number; // in seconds
     public weight: number;
     public level: number;
+    public tree_weight: number;
```
GabrielInTheWorld, line 39: I don't know if this is the right place for this information.

FinnStutzenstein: Ok, please add a comment, that is is now client calculated. Please also set the level in the client and mark `level` as client-calculated here, too.

FinnStutzenstein (approved):

PR #468, opened 2021-08-16
Uses decorators to generic copying data before making an http request

### LINKED ISSUE #124: Generic copy-code for repositories
We need a solution to reduce the amount of explicit copying in the repos. This is too explicit and repetitive.

GabrielInTheWorld:
The data classes have to have initial/default values, so an instance of those classes have their properties.

Fixes #124 

@tsiegleauq and @FinnStutzenstein What do you think?

### DIFF
--- a/client/src/app/core/actions/committee-action.ts
+++ b/client/src/app/core/actions/committee-action.ts
@@ -1,40 +1,46 @@
 import { Identifiable } from 'app/shared/models/base/identifiable';
+import { DataClass } from '../decorators/data-class';
 import { Id, UnsafeHtml } from '../definitions/key-types';
 
-export namespace CommitteeAction {
-    export const CREATE = 'committee.create';
-    export const UPDATE = 'committee.update';
-    export const DELETE = 'committee.delete';
+export class CommitteeAction {
+    public static readonly CREATE = 'committee.create';
+    public static readonly UPDATE = 'committee.update';
+    public static readonly DELETE = 'committee.delete';
+}
 
-    export interface PartialPayload {
-        description?: UnsafeHtml;
-        // For UPDATE: Required permission: OML.can_manage_organization
-        user_ids?: Id[];
-        manager_ids?: Id[];
-        // For UPDATE: Required permission: CML.can_manage || OML.can_manage_organization
-        organization_tag_ids?: Id[];
-        // Required permission: OML.can_manage_organization
-        forward_to_committee_ids?: Id[];
-        receive_forwardings_from_committee_ids?: Id[];
-    }
+class CommitteePartialPayload {
+    description?: UnsafeHtml = '';
+    // For UPDATE: Required permission: OML.can_manage_organization
+    user_ids?: Id[] = [];
+    manager_ids?: Id[] = [];
+    // For UPDATE: Required permission: CML.can_manage || OML.can_manage_organization
+    organization_tag_ids?: Id[] = [];
+    // Required permission: OML.can_manage_organization
+    forward_to_committee_ids?: Id[] = [];
+    receive_forwardings_from_committee_ids?: Id[] = [];
+}
 
-    /**
-     * Required permission: `OML.can_manage_organization`
-     */
-    export interface CreatePayload extends PartialPayload {
-        name: string;
-        organization_id: Id;
-    }
+/**
+ * Required permission: `OML.can_manage_organization`
+ */
+@DataClass(CommitteeAction.CREATE)
+export class CommitteeCreatePayload extends CommitteePartialPayload {
+    name: string = '';
+    organization_id: Id = 0;
+}
 
-    /**
-     * Required permission: `CML.can_manage` || `OML.can_manage_organization`
-     */
-    export interface UpdatePayload extends Identifiable, PartialPayload {
-        // Optional
-        // Required permission: CML.can_manage
-        name?: string;
-        template_meeting_id?: Id;
-    }
+/**
+ * Required permission: `CML.can_manage` || `OML.can_manage_organization`
+ */
+@DataClass(CommitteeAction.UPDATE)
+export class CommitteeUpdatePayload extends CommitteePartialPayload implements Identifiable {
+    id: Id = 0;
+    name?: string = '';
+    template_meeting_id?: number = 0;
+    default_meeting_id?: number = 0;
+}
 
-    export interface DeletePayload extends Identifiable {}
+@DataClass(CommitteeAction.DELETE)
+export class CommitteeDeletePayload implements Identifiable {
+    id: Id = 0;
 }
--- a/client/src/app/core/decorators/data-class.ts
+++ b/client/src/app/core/decorators/data-class.ts
@@ -1,0 +1,49 @@
+import { Type } from '@angular/core';
+
+class DataClassContainer {
+    private static registry: { [descriptorKey: string]: any } = {};
+
+    public static register(descriptor: string, value: Type<any>) {
+        if (!this.registry[descriptor]) {
+            this.registry[descriptor] = new value();
+        }
+    }
+
+    public static get(descriptor: string): any | undefined {
+        return this.registry[descriptor];
+    }
+}
+
+export function DataClass(descriptor: string): any {
+    return (target: Type<any>) => {
+        DataClassContainer.register(descriptor, target);
+    };
+}
+
+export function useDataClass(key: string, propertyIndex: number = 0): any {
+    return (_target: Type<any>, _propertyKey: string, descriptor: TypedPropertyDescriptor<Function>): any => {
+        const dataClass = DataClassContainer.get(key);
+        if (!dataClass) {
+            throw new Error('Undefined data class');
+        }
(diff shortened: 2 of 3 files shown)


client/src/app/core/actions/committee-action.ts
```diff
@@ -1,40 +1,46 @@
...
+    name?: string = '';
+    template_meeting_id?: number = 0;
+    default_meeting_id?: number = 0;
+}
 
-    export interface DeletePayload extends Identifiable {}
+@DataClass(CommitteeAction.DELETE)
+export class CommitteeDeletePayload implements Identifiable {
+    id: Id = 0;
 }
```
GabrielInTheWorld, line 46: I have used here separate classes, because we do not want to use `namespaces` as we discussed, do we?

client/src/app/core/actions/committee-action.ts
```diff
@@ -1,40 +1,46 @@
...
+class CommitteePartialPayload {
+    description?: UnsafeHtml = '';
+    // For UPDATE: Required permission: OML.can_manage_organization
+    user_ids?: Id[] = [];
+    manager_ids?: Id[] = [];
+    // For UPDATE: Required permission: CML.can_manage || OML.can_manage_organization
+    organization_tag_ids?: Id[] = [];
+    // Required permission: OML.can_manage_organization
+    forward_to_committee_ids?: Id[] = [];
+    receive_forwardings_from_committee_ids?: Id[] = [];
```
GabrielInTheWorld, line 20: We could here also use decorators to inject default values, but I think, that will be an extra expense but has no additional value...

client/src/app/core/decorators/data-class.ts
```diff
@@ -0,0 +1,49 @@
...
+    public static get(descriptor: string): any | undefined {
+        return this.registry[descriptor];
+    }
+}
+
+export function DataClass(descriptor: string): any {
+    return (target: Type<any>) => {
+        DataClassContainer.register(descriptor, target);
+    };
+}
```
tsiegleauq, line 21: That name is a little too generic

tsiegleauq (changes requested):

client/src/app/core/decorators/data-class.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export function useDataClass(key: string, propertyIndex: number = 0): any {
+    return (_target: Type<any>, _propertyKey: string, descriptor: TypedPropertyDescriptor<Function>): any => {
+        const dataClass = DataClassContainer.get(key);
+        if (!dataClass) {
```
FinnStutzenstein, line 26: Move this error to the get function and do never return undefined from it.

client/src/app/core/actions/committee-action.ts
FinnStutzenstein: I do not like the defaults. As I can see the actual class is used in data-class.ts line 32 ff.:
```
for (const key in dataClass) {
    result[key] = value[key];
}
```
1. You do nat care about types
2. You do not care about the default values
3. you just need the field names.

Why not just use a list of fields to copy:
```
CommitteePartialPayloadFields = ["description", "user_ids", "manager_ids", ...]
```
This contains all information you need without the unnecessary defaults (and types)

GabrielInTheWorld: I know, that there are then unnecessary default values and types. My reason for it was, that we can "combine" definitions (like an interface) and such fields to copy.
If we use a separate list of fields to indicate which fields have to be "copied", we have two places to maintain - the definitions as well as the list of fields. 
I can use the default values to check if the copied values have the correct types.

FinnStutzenstein: > I can use the default values to check if the copied values have the correct types.

GabrielInTheWorld: > Please add an example then, so we can see the usecase.
>

See the file `committee-repository.service.ts` line 65 to 69: There I replaced the "old" `CommitteeAction.UpdatePayload` with the "new" class `CommitteeUpdatePayload`. 
 
> Currently, there are also two places: In the actions and the model fields.
> 

Well, the models and actions do not contain the same keys.

> I would suggest to add the list of fields into the model classes (e.g. motion.ts) and add a constants to every class (like `FIELDS`). Yes, we have them twice, but then in the same file. But I guess you can use typescript's information to extract every key from the interface. A quick google search: https://stackoverflow.com/questions/43909566/get-keys-of-a-typescript-interface-as-array-of-strings and https://github.com/kimamula/ts-transformer-keys
> 
> We have all fields in each model file (There we need them explicitly once). Based on this list you can add/remove additional fields and build up the list of fields for each action.

The problem is, that `const`s cannot be decorated (as mentioned in the comment above).

client/src/app/shared/models/event-management/meeting.ts
```diff
@@ -20,10 +22,14 @@ export type AgendaItemCreation = 'always' | 'never' | 'default_yes' | 'default_n
...
-export interface Settings {
+const MEETING_NAME = 'meeting';
+export class Settings {
     // Old "general_*" configs
+    @DataClassProperty([], MEETING_NAME)
     name: string;
+    @DataClassProperty([], MEETING_NAME)
     description: string;
+    @DataClassProperty([], MEETING_NAME)
     location: string;
```
GabrielInTheWorld, line 33: As discussed I have thought about another way to achieve those information we need at runtime.
Here is a more "spring-boot-like" approach:
The properties, which can be sent in a specific action to the server, are decorated with such a decorator. To distinguish between different actions and different payloads there are different "namespaces" we define by a string (our actual actions), like `meeting.create` or `meeting.update`.

client/src/app/shared/models/event-management/meeting.ts
```diff
@@ -20,10 +22,14 @@ export type AgendaItemCreation = 'always' | 'never' | 'default_yes' | 'default_n
  */
 export type BallotPaperSelection = 'NUMBER_OF_DELEGATES' | 'NUMBER_OF_ALL_PARTICIPANTS' | 'CUSTOM_NUMBER';
 
-export interface Settings {
+const MEETING_NAME = 'meeting';
+export class Settings {
```
GabrielInTheWorld, line 26: As I mentioned before, only classes and class-members can have decorators. Therefore, I had to "convert" the interface `Settings` into a class.

client/src/app/shared/models/event-management/meeting.ts
```diff
@@ -166,6 +172,22 @@ export interface Settings {
...
+        'description',
+        'start_time',
+        'end_time',
+        'location',
+        'url_name',
+        'enable_anonymous'
+    ]
+})
+@DataClass<Meeting>(MeetingAction.DELETE, { useOnly: ['id'] })
+@DtoClass()
```
GabrielInTheWorld, line 190: About the naming we can discuss later.

GabrielInTheWorld:
I have investigated into the research for auto-generating documentation of our Angular-client. I have investigated several libraries such as `CompoDoc`, `TypeDoc`, `AngularDoc`, `docco` and some other. But I have not found a library for our purposes.

Another thing: I have thought about the auto-generation of our forms. I ended up that we then move the definition for those forms from the file we use a form into an own definition-file, if we declare those definitions in a typescript-file.

client/src/app/shared/models/event-management/meeting.ts
```diff
@@ -20,12 +24,19 @@ export type AgendaItemCreation = 'always' | 'never' | 'default_yes' | 'default_n
...
+    @FormClassControl([MeetingAction.UPDATE])
     name: string;
+    @FormClassControl([MeetingAction.CREATE, MeetingAction.UPDATE])
     description: string;
+    @FormClassControl([MeetingAction.CREATE, MeetingAction.UPDATE])
     location: string;
+    @FormClassControl([MeetingAction.CREATE, MeetingAction.UPDATE])
     start_time: number;
+    @FormClassControl([MeetingAction.CREATE, MeetingAction.UPDATE])
     end_time: number;
```
GabrielInTheWorld, line 40: Another approach we have talked about is to use decorators to describe how to create a form-group.

client/src/app/management/components/meeting-edit/meeting-edit.component.ts
```diff
@@ -198,26 +200,36 @@ export class MeetingEditComponent extends BaseModelContextComponent implements O
...
-            user_ids: [[]],
-            admin_ids: [[], Validators.minLength(1)],
-            organization_tag_ids: [[]]
-        };
+        const action = this.isCreateView ? MeetingAction.CREATE : MeetingAction.UPDATE;
+        const additionalFormControls = { admin_ids: [[], Validators.minLength(1)] };
+        const rawForm = this.formgroupGenerator.generateFormGroup(action, additionalFormControls, {
+            start_time: { defaultValue: currentDate },
+            end_time: { defaultValue: currentDate }
+        });
```
GabrielInTheWorld, line 208: Then we can use those descriptions like this.
What do you mean @tsiegleauq and @FinnStutzenstein ?

PR #469, opened 2021-08-16
Adds highlighting to default meetings

### LINKED ISSUE #327: Set a meeting as "default-meeting" of a committee

GabrielInTheWorld:
Uses #468 
Fixes #327

Waits for OpenSlides/openslides-backend#906


tsiegleauq (changes requested):
Not tested.

the errors are in `/app/src/app/core/decorators/data-class.ts`
```
 32:13  error    The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype           guard-for-in
  32:24  error    'key' is already declared in the upper scope                                                                         @typescript-eslint/no-shadow
  37:36  error    Unexpected space before function parentheses                                                                         space-before-function-paren
```

if you avoid using namespaces now, remove `@typescript-eslint/no-namespace": ["warn"]` from `client/.eslint.json`. Doing so will declare namespaces as errors instead of warnings.

Rest looks fine so far

GabrielInTheWorld:
I have removed the first commit `Uses decorators to generic copying data before making an http request`, since it is not clarified if we use this.

PR #483, opened 2021-08-31
Not found button is not shown if no value is given

### LINKED ISSUE #480: The "not found"-button in the search-repo/-value-selector should only be displayed if a value in the search-input is given

GabrielInTheWorld:
Fixes #480

### DIFF
--- a/client/src/app/shared/directives/not-found-description.directive.ts
+++ b/client/src/app/shared/directives/not-found-description.directive.ts
@@ -1,0 +1,4 @@
+import { Directive } from '@angular/core';
+
+@Directive({ selector: '[osNotFoundDescription]' })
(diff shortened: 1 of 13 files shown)


client/src/app/shared/directives/not-found-description.directive.ts
```diff
@@ -0,0 +1,4 @@
+import { Directive } from '@angular/core';
+
+@Directive({ selector: '[notFoundDescription]' })
```
tsiegleauq, line 3: `3:24  error  The selector should start with one of these prefixes: "os" (https://angular.io/guide/styleguide#style-02-08)  @angular-eslint/directive-selector`

tsiegleauq (approved):

PR #492, opened 2021-09-03
Users can forward to meetings within the same committee

### LINKED ISSUE #487: Cannot update committee: `data.forward_to_committee_ids` must be array
trying to update a committee throws:
`Error: data.forward_to_committee_ids must be array`

Request Object was:
```
{
    "action": "committee.update",
    "data": [
        {
            "id": 4,
            "name": "Neu 123d",
            "description": "",
            "organization_tag_ids": null,
            "user_ids": [
                1
            ],
            "manager_ids": [
                1
            ],
            "forward_to_committee_ids": null,
            "receive_forwardings_from_committee_ids": null
        }
    ]
}
```

GabrielInTheWorld:
Now, users can select the same committee to forward/receive motions from that is currently being edited.

Fixes #487

### DIFF
--- a/client/src/app/management/components/committee-edit/committee-edit.component.html
+++ b/client/src/app/management/components/committee-edit/committee-edit.component.html
@@ -68,8 +68,9 @@
                     formControlName="forward_to_committee_ids"
                     [multiple]="true"
                     [repo]="committeeRepo"
-                    [pipeFn]="getPipeFilterFn()"
                     [showChips]="false"
+                    [lazyLoading]="false"
+                    (selectionChanged)="onSelectionChanged($event)"
                 ></os-search-repo-selector>
             </mat-form-field>
 
@@ -79,8 +80,10 @@
                     formControlName="receive_forwardings_from_committee_ids"
                     [multiple]="true"
                     [repo]="committeeRepo"
-                    [pipeFn]="getPipeFilterFn()"
                     [showChips]="false"
+                    [lazyLoading]="false"
+                    [disableOptionWhenFn]="getDisableOptionWhenFn()"
+                    [tooltipFn]="getTooltipFn()"
                 ></os-search-repo-selector>
             </mat-form-field>
         </ng-container>
--- a/client/src/app/management/components/committee-edit/committee-edit.component.ts
+++ b/client/src/app/management/components/committee-edit/committee-edit.component.ts
@@ -24,9 +24,11 @@ import { Identifiable } from 'app/shared/models/base/identifiable';
 import { Committee } from 'app/shared/models/event-management/committee';
 import { BaseModelContextComponent } from 'app/site/base/components/base-model-context.component';
 import { ViewUser } from 'app/site/users/models/view-user';
+import { OsOptionSelectionChanged } from '../../../shared/components/search-selector/base-search-value-selector/base-search-value-selector.component';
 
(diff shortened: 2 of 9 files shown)


client/src/app/management/components/committee-edit/committee-edit.component.html
```diff
@@ -68,8 +68,9 @@ <h2 *ngIf="!isCreateView">{{ editCommitteeLabel | translate }}</h2>
                     formControlName="forward_to_committee_ids"
                     [multiple]="true"
                     [repo]="committeeRepo"
-                    [pipeFn]="getPipeFilterFn()"
                     [showChips]="false"
+                    [lazyLoading]="false"
```
GabrielInTheWorld, line 72: The committees cannot be lazy loaded, because it will then overwrite the current form, since an incoming committee (from the `getViewModelObservable`-subscription) would update the current form.

emanuelschuetze (approved):

client/src/app/management/components/committee-edit/committee-edit.component.ts
```diff
@@ -93,6 +94,34 @@ export class CommitteeEditComponent extends BaseModelContextComponent implements
...
+    /**
+     * Returns a function to get a tooltip if the option which contains this committee is rendered
+     * in the `receive_forwardings_from_committee_ids`-control
+     *
+     * @returns A function that will return a tooltip
+     */
+    public getTooltipFn(): (value: ViewCommittee) => string | null {
+        return value => {
+            if (value.id === this.committeeId) {
+                return 'You can change this option only in the forwarding section.';
```
tsiegleauq, line 118: this string needs to be marked as translatable

client/src/app/management/components/meeting-preview/meeting-preview.component.ts
```diff
@@ -34,7 +34,7 @@ export class MeetingPreviewComponent {
     }
 
     public get userAmount(): number {
-        return this.meeting?.user_ids.length;
+        return this.meeting?.user_ids?.length;
```
tsiegleauq, line 37: the gettwe now can return null or undefined. Change return value accordingly. Otherwise return 0 if  meeting or user_ids does not exsit

tsiegleauq (changes requested):

client/src/app/management/components/committee-edit/committee-edit.component.ts
```diff
@@ -93,6 +94,34 @@ export class CommitteeEditComponent extends BaseModelContextComponent implements
...
+    /**
+     * Returns a function to get a tooltip if the option which contains this committee is rendered
+     * in the `receive_forwardings_from_committee_ids`-control
+     *
+     * @returns A function that will return a tooltip
+     */
+    public getTooltipFn(): (value: ViewCommittee) => string | null {
+        return value => {
+            if (value.id === this.committeeId) {
+                return this.translate.instant('You can change this option only in the forwarding section.');
```
tsiegleauq, line 118: The translation should happen using the translate pipe in HTML.
To mark something translatable you need to use the marker function.

```ts
import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
```
and

```ts
return _('You can change this option only in the forwarding section.');
```

tsiegleauq (changes requested):

tsiegleauq (approved):

PR #513, opened 2021-09-14
Implements a "single shot" request to the AU-service

### LINKED ISSUE #497: Implement the "Single Shot"-request-action to the AU-service
Since OpenSlides/openslides-autoupdate-service#309 the AU-service provides a "Single Shot"-http-route to get instantly requested data.
The URL is
```
<AU-host>:<AU-port>/system/autoupdate?single=true
```
and expects data in the same format as we send for any other autoupdates.

GabrielInTheWorld:
Fixes #497 

Heavily restructures the existing communication to external services. I think, this is a more reliable and resilient. We have then a direct way to create streams.
Removes also unused files.


client/src/app/core/core-services/http-stream-endpoint.service.ts
```diff
@@ -17,8 +15,19 @@ export class HttpStreamEndpointService {
...
+        name: string,
+        configuration: EndpointConfiguration | string,
+        healthUrl?: string,
+        method: HTTPMethod = HTTPMethod.GET
+    ): void {
+        if (configuration instanceof EndpointConfiguration) {
+            this.endpointConfigurations[name] = configuration;
+        } else {
+            this.endpointConfigurations[name] = { url: configuration, healthUrl, method };
+        }
```
tsiegleauq, line 30: why not simply have
`public registerEndpoint(name: string, configuration: EndpointConfiguration): void` ?
Overloading every possibility seems unnecessary here.

client/src/app/core/core-services/http-stream.ts
```diff
@@ -0,0 +1,381 @@
...
+            this.propagateError(originalText, 'Reported error by server');
+        } else {
+            this.onMessage(parsedContent);
+        }
+    }
+
+    private propagateError(text: string, reason: string, type: ErrorType = this.getErrorTypeFromStatusCode()): void {
+        this.onError(type, text, reason);
+    }
+}
```
tsiegleauq, line 180: should be large enough to be in it's own file

client/src/app/core/core-services/http-stream.ts
```diff
@@ -0,0 +1,381 @@
...
+     * The absolute value will be used to avoid the handling of negative numbers.
+     */
+    reconnectsBeforeClose?: number;
+    /**
+     * A number in ms a stream will wait before it tries to reconnect
+     *
+     * This can also be a function, that will return such a number. Defaults to 2000 ms.
+     */
+    reconnectTimeout?: number | (() => number);
+}
```
tsiegleauq, line 82: should be large enough to be in it's own file

tsiegleauq (approved):
Tested, works fine, seems to be solid. Looks cleaner than before. Some slight cleanup remarks that can be done later. Merging since there are some dependencies for other features

PR #523, opened 2021-09-17
Scrollable sorted jitsi user list

### LINKED ISSUE #404: Cannot scroll jitsi user list
**Describe the bug**
After restructure, the jitsi user list cannot be scrollen

**How to Reproduce**
Requires working jitsi conference and a high amount of users

**Expected behavior**
There should be a scrollbar

**System information**
any

**Additional context**
Add any other information to comprehend your problem

tsiegleauq:
fixes #404
Sorts and scrolls through the jitsi users


client/src/app/site/interaction/components/call/call.component.ts
```diff
@@ -47,10 +48,7 @@ export class CallComponent extends BaseComponent implements OnInit, AfterViewIni
     }
 
     private dominantSpeaker: string;
-    private members = {};
-    public get memberList(): string[] {
-        return Object.keys(this.members);
-    }
+    public members: ConferenceMemberCollection = {};
```
GabrielInTheWorld, line 51: I like, that there are type-information are added. But if the `members`-variable is not changed from the outside, it should be kept private. Please make it accessible from the outside by a `get`-function (if it won't be changed from there).

GabrielInTheWorld (changes requested):
One remark. The rest looks good!

client/src/app/site/interaction/components/call/call.component.ts
tsiegleauq: seems it was not necessary at all since members can be accessed using async pipe

GabrielInTheWorld (approved):
Looks good! Not tested.

PR #603, opened 2021-10-29
Fixes the deletion of models

### LINKED ISSUE #592: Nested models (requested by relation-lists) are not updated
If I delete an agenda_item, the AU-service sends the data `meeting/X/agenda_item_ids: null` (in case there was only one agenda_item), which is correct. But it does not send data for the related objects of the removed agenda_items, anymore.

This is also the case for motions, if the restrictions for a workflow were changed.

GabrielInTheWorld:
Improves the AutoupdateService a little bit to work on class-members instead of passing variables only through various functions...
Now, models will be deleted again...

Fixes #592


tsiegleauq (approved):
Tested, looks good, works fine

PR #604, opened 2021-10-29
Shows a warning if a user is linked to a motion as submitter or an election as candidate or anything else

### LINKED ISSUE #495: Warning when deleting a user if they are assigned to something
If a user is assigned to something (e.g. as candidate in an election, as speaker for a topic or as a submitter for a motion) and they should be deleted, then a warning should raise.
The warning should inform performing users, that the user to delete is linked to other models.

GabrielInTheWorld:
- Implements the "get_user_related_models"-presenter
- Also implements a "PresenterService" for presenters

When a user should be deleted, then a dialog is appearing which shows to which meetings and committees the user is linked to. If the user is linked to something as submitter or candidate (...), their account is only deactivated. If a requesting user (within a meeting) has not the necessary permissions to perform this action, the user will be removed from the meeting instead of being deleted.

Fixes #495

### DIFF
--- a/client/src/app/core/core-services/member.service.ts
+++ b/client/src/app/core/core-services/member.service.ts
@@ -3,15 +3,45 @@ import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 
 import { ViewUser } from 'app/site/users/models/view-user';
-import { HttpService } from './http.service';
 import { Id } from '../definitions/key-types';
 import { SimplifiedModelRequest } from './model-request-builder.service';
 import { UserRepositoryService } from '../repositories/users/user-repository.service';
 import { OperatorService } from './operator.service';
-import { OML } from './organization-permission';
+import { OML, CML } from './organization-permission';
+import { Presenter, PresenterService } from './presenter.service';
+import { Meeting } from 'app/shared/models/event-management/meeting';
+import { MatDialog } from '@angular/material/dialog';
+import { MemberDeleteDialogComponent } from '../../management/components/member-delete-dialog/member-delete-dialog.component';
+import { mediumDialogSettings } from 'app/shared/utils/dialog-settings';
+import { ActiveMeetingService } from './active-meeting.service';
+import { TranslateService } from '@ngx-translate/core';
+import { PromptService } from '../ui-services/prompt.service';
 
-export interface GetUsersRequest {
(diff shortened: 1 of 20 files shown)


GabrielInTheWorld:
I have also updated the orga-settings, so the key `limit_of_users` is requested

client/src/app/core/core-services/member.service.ts
```diff
@@ -74,4 +111,41 @@ export class MemberService {
...
+
+        for (const user of users) {
+            data[user.id].name = user.getFullName();
+        }
+
+        const promptDialog = this.dialog.open(MemberDeleteDialogComponent, { ...mediumDialogSettings, data });
+        if (await promptDialog.afterClosed().toPromise()) {
+            await this.doDelete(users);
+        } else {
+            throw new NoActionError();
```
tsiegleauq, line 128: "No Action" Should not be an error, but is perfectly normal behavior. Don't just throw an error here.

client/src/app/site/users/components/user-detail/user-detail.component.ts
```diff
@@ -298,12 +300,10 @@ export class UserDetailComponent extends BaseModelContextComponent implements On
      * click on the delete user button
      */
     public async deleteUserButton(): Promise<void> {
-        const title = this.translate.instant('Are you sure you want to delete this participant?');
-        const content = this.user.full_name;
-        if (await this.promptService.open(title, content)) {
-            await this.repo.delete(this.user);
+        try {
+            await this.memberService.delete([this.user]);
```
tsiegleauq, line 304: although the error was thrown, after "not" deleting a user the next line was evaluated. Meaning: we would come back into the user list after aborting the delete action, this was surely unexpected. Better to evaluate if the delete request actually returns something and then decide to navigate or stay

tsiegleauq (changes requested):

client/src/app/site/users/components/user-detail/user-detail.component.ts
GabrielInTheWorld: Yeah, I do not like this implementation any more than you do, but I do not like it to return something to continue through the program flow. But, in this case returning a boolean is very likely the best option.

emanuelschuetze (approved):

tsiegleauq (approved):

PR #609, opened 2021-11-01
Show warning if the user has no https

### LINKED ISSUE #536: Show a warning on http
OpenSlides4 does not support http. But it seems to work.

Could you add a warning or a banner, that http is not supported? This can not be done on the server, since the https could have been terminated on a proxy.

See https://github.com/OpenSlides/openslides-manage-service/issues/76

tsiegleauq:
Shows a warning on the login page if the user was not using a https
connection

fixes #536

image

### DIFF
--- a/client/src/app/site/login/components/login-mask/login-mask.component.ts
+++ b/client/src/app/site/login/components/login-mask/login-mask.component.ts
@@ -2,6 +2,7 @@ import { Component, OnDestroy, OnInit } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 import { ActivatedRoute, Router } from '@angular/router';
 
+import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 import { Subscription } from 'rxjs';
 import { filter } from 'rxjs/operators';
 
@@ -18,6 +19,8 @@ import { fadeInAnim } from 'app/shared/animations';
 import { ParentErrorStateMatcher } from 'app/shared/parent-error-state-matcher';
 import { BaseComponent } from 'app/site/base/components/base.component';
 import { BrowserSupportService } from '../../services/browser-support.service';
+
+const HttpWarning = _('Using OpenSlides over HTTP is not supported. Enable HTTPS to continue.');
 
 /**
  * Login mask component.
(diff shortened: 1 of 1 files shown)


client/src/app/site/login/components/login-mask/login-mask.component.ts
```diff
@@ -19,6 +20,8 @@ import { ParentErrorStateMatcher } from 'app/shared/parent-error-state-matcher';
 import { BaseComponent } from 'app/site/base/components/base.component';
 import { BrowserSupportService } from '../../services/browser-support.service';
 
+const HttpWarning = _('Using OpenSlides over a HTTP connection is not supported. Enable HTTPS to continue.');
```
GabrielInTheWorld, line 23: This is a sophistry, but should it not be called a_n_ HTTP connection?

client/src/app/site/login/components/login-mask/login-mask.component.ts
```diff
@@ -191,6 +196,12 @@ export class LoginMaskComponent extends BaseComponent implements OnInit, OnDestr
         this.router.navigate(['./reset-password'], { relativeTo: this.route });
     }
 
+    private raiseHttpWarning(): void {
```
GabrielInTheWorld, line 199: I would expect something like "checkHttpConnection" or "checkForHttp" rather than "raiseHttpWarning". Reading "raise...Warning" suggests that a warning will always be shown - regardless whether a check succeed or not.

GabrielInTheWorld (changes requested):

client/src/app/site/login/components/login-mask/login-mask.component.ts
tsiegleauq: depends on whether you say aitch or haitch.

> An `aitch tee tee pee` connection
> A `haitch tee tee pee` connection

Perhaps I should avoid using a or an ans simply write "Using OpenSlides over HTTP is not supported"

ostcar:
I don't know how to test this but it looks good.

GabrielInTheWorld (approved):

ostcar (approved):
The screenshot looks good

PR #638, opened 2021-11-10
Enhance call restriction

tsiegleauq:
subscribe to autoupdate for getting CLOS updates, so users know if they
are on CLOS to enter automatically
simplify call restriction service
enhance 2nd tab detection to prevent jitsi connections on new tabs

### DIFF
--- a/client/src/app/site/agenda/models/view-list-of-speakers.spec.ts
+++ b/client/src/app/site/agenda/models/view-list-of-speakers.spec.ts
@@ -1,0 +1,104 @@
+import { ListOfSpeakers } from 'app/shared/models/agenda/list-of-speakers';
+import { Speaker } from 'app/shared/models/agenda/speaker';
+
+import { UserListIndexType, ViewListOfSpeakers } from './view-list-of-speakers';
+import { ViewSpeaker } from './view-speaker';
+
+const LosSpeakers: ViewSpeaker[] = [
+    new ViewSpeaker(
+        new Speaker({
+            id: 1,
+            user_id: 1,
+            weight: 4
+        })
+    ),
+    new ViewSpeaker(
+        new Speaker({
+            id: 2,
+            user_id: 2,
+            weight: 3
+        })
+    ),
+    new ViewSpeaker(
+        new Speaker({
+            id: 3,
+            user_id: 3,
+            weight: 2
+        })
+    ),
+    new ViewSpeaker(
+        new Speaker({
+            id: 4,
+            user_id: 4,
+            weight: 1
+        })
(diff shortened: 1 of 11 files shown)


emanuelschuetze (changes requested):
Autoconnect works. But if you open 2nd or 3rd browser tab a 2nd or 3rd livestream is started now. In current client-master this works as expected (only the 1st tab shows the livestream).

Other issues:
* It's not possible to exit from a jitsi conference if autoconnect is enabled and user is still on CLOS.
* In Firefix after _each_ connection to Jitsi you get again this message: "Bitte erlauben Sie jetzt den Zugriff auf Kamera und Mikrofon"

tsiegleauq:
Firefox behavior can only be prevented by disabling the pre checks before
the jitsu connection was established. That's most likely undesired. The
other issues can be addressed

Emanuel Schütze ***@***.***> schrieb am Do., 18. Nov. 2021,
13:16:

tsiegleauq:
> if you open 2nd or 3rd browser tab a 2nd or 3rd livestream is started now. In current client-master this works as expected (only the 1st tab shows the livestream).

I could not reproduce that. The live stream in other tab detection worked as expected and is, as it was before, completely unrelated from the rtc detection. If it was desired that "either" an RTC or live stream was running in all tabs, the fix should be easy, but could bear undesired sideeffects. The "your livestream is already running" could disappear more often when its not intended.
The solution would be, do not load the live stream, if either live stream lock or rtc lock was set.

> It's not possible to exit from a jitsi conference if autoconnect is enabled and user is still on CLOS.

Fixed by havong a new flag "left the meeting manually" in the RTC service. This will work in the current tab and should prevent (all) automatic connections for the current session

> In Firefix after each connection to Jitsi you get again this message: "Bitte erlauben Sie jetzt den Zugriff auf Kamera und Mikrofon"

as explained before, this will most likely be undesired since we would lose the possibility of showing the spinner before an RTC connection opens.

GabrielInTheWorld (approved):
Okay, not tested.

emanuelschuetze (changes requested):
Autoconnect on chromium v95 does not work. Tested with admin and normal user (with fresh localstorage). Firefox works as expected.

tsiegleauq:
> Autoconnect on chromium v95 does not work. Tested with admin and normal user (with fresh localstorage). Firefox works as expected.

cannot reproduce. Make sure you dont have a lock in the indexeddb.

emanuelschuetze (approved):
ok, good enough for now. We discuss the autoconnect issues later.

PR #640, opened 2021-11-11
Use new vote service

tsiegleauq:
Send votes to vote service

Sends the vote objects to the vote service directly
Gets the "have I voted" information from the vote services on request

### DIFF
--- a/client/src/app/shared/models/poll/poll.ts
+++ b/client/src/app/shared/models/poll/poll.ts
@@ -21,14 +21,10 @@ export class Poll extends BaseDecimalModel<Poll> {
     public onehundred_percent_base: PollPercentBase;
 
     /**
-     * TODO: these fields has to be autocalculated by the AU-service: OpenSlides/openslides-autoupdate-service#262
+     * TODO:
+     * Not sure how vote delegations are handled now
      */
-    public voted_id: number[];
-    public user_has_voted: boolean;
     public user_has_voted_for_delegations: Id[];
-    /**
-     * END
-     */
 
     public pollmethod: PollMethod;
 
--- a/client/src/app/shared/models/poll/view-poll.ts
+++ b/client/src/app/shared/models/poll/view-poll.ts
@@ -32,6 +32,7 @@ export class ViewPoll<C extends BaseViewModel<BaseModel> = any>
     }
     public static COLLECTION = Poll.COLLECTION;
     protected _collection = Poll.COLLECTION;
+    public hasVoted = false;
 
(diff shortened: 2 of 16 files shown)


client/src/app/shared/models/poll/poll.ts
```diff
@@ -21,14 +21,17 @@ export class Poll extends BaseDecimalModel<Poll> {
...
     /**
-     * TODO: these fields has to be autocalculated by the AU-service: OpenSlides/openslides-autoupdate-service#262
+     * TODO:
+     * Does currently not exist. Can this be removed?
      */
-    public voted_id: number[];
-    public user_has_voted: boolean;
-    public user_has_voted_for_delegations: Id[];
+    // public voted_id: number[];
+    // public user_has_voted: boolean;
```
GabrielInTheWorld, line 28: I think, this can be removed, if it is not used.

GabrielInTheWorld (approved):
Looks good! Please remove the unused lines.

client/src/app/shared/models/poll/view-poll.ts
```diff
@@ -73,15 +73,15 @@ export class ViewPoll<C extends BaseViewModel<BaseModel> = any>
...
     }
 
-    /**
-     * Is injected by the poll-repo
-     * TODO: Can be removed, when OpenSlides/openslides-autoupdate-service#262 is resolved.
-     */
-    public operatorHasVoted: () => boolean;
+    public set hasVoted(voted: boolean) {
+        this._hasVoted = voted;
+    }
```
GabrielInTheWorld, line 84: If the value of the variable will only be written and read, then write just
```js
public hasVoted = false;
```

client/src/app/site/polls/components/base-poll-vote.component.ts
```diff
@@ -59,17 +65,20 @@ export abstract class BasePollVoteComponent<C extends BaseViewModel = any> exten
             this.voteRequestData[this.user.id] = {
                 value: {}
             } as VotingData;
-            this.alreadyVoted[this.user.id] = this.poll.user_has_voted;
+            this.alreadyVoted[this.user.id] = this.poll.hasVoted;
             this.deliveringVote[this.user.id] = false;
         }
 
+        console.log(`AAA createVotingDataObjects has delegations: `, this.delegations);
```
GabrielInTheWorld, line 72: I think, this could be removed or do you need it for some debug purposes?

GabrielInTheWorld (changes requested):
Two remarks. The rest looks good.

PR #642, opened 2021-11-12
Fixes some bugs

### LINKED ISSUE #627: Summarize issue
- [x] UserDetailComponent: The checkbox "is present" does not reflect the current state
- [x] UserDetailComponent: Disable the "save" button to prevent clicking it, if no fields are changed
- [x] AP: Remove the first `<mat-card />`, if there is no content (title)
- [x] MotionDetailComponent: Remove the selection for categories, if there are no categories in a meeting
- [x] AgendaItemListComponent: Remove agenda_items if a user cannot see them. Attention: If such agenda_items contain children, which the user can see, then the agenda_items should be displayed (at least with a default like `-`)
- [x] CommitteeDetailComponent: `Cannot read property of undefined (reading '2')` -> if the CML of a user is not already available
- [x] General: Prevent clicking a "save" button twice. This is especially for CREATE-views necessary, because then some models could be created twice.

GabrielInTheWorld:
- Removes the title card in the AP, if it is empty
- Prevents double clicking a save action by showing a spinner instead of a save button
- Agenda items will be hidden, if they are not allowed to be seen
- Only users, who can see accounts, can click the account button in a <meeting-preview />-card
- Also prevents an ugly behavior if the CML of a user is not yet loaded

Besides, the global spinner component was renamed to just "spinner component" and the "*osOmlPerms" directive has the same feature (using "then/else") as the "*ngIf" directive.

Fixes #627


emanuelschuetze (approved):

emanuelschuetze:
@tsiegleauq I'll merge it now. As discussed you can review later if you like

PR #644, opened 2021-11-16
Determines, whether a user can be edited or not, now

### LINKED ISSUE #633: Within a meeting, a user with the permission `user.can_manage` is not allowed to change information about another user
Metadata of users (like title, first_name, last_name, gender, ...) can only be changed on orga-level. These fields should be disabled in the UserDetailComponent.
Furthermore, it has to be ensured, that an outgoing request to the backend does not contain such fields, if they were not changed.

GabrielInTheWorld:
- Fixes incorrect editing of users who are out of scope
- Fixes the loading of meetings in the dashboard
- Fixes the 'non-admin-check' in the OperatorService

Fixes #633

There are different ways to test:
- As superadmin, I should be able to edit every user regardless of their scope
- As only meeting-admin, I should be able to edit every user, who is only in the same meeting (not in a committee or elsewhere)
- As only meeting-admin, I should always be able to edit meeting-related information of every user in the same meeting
- I can always edit my own username, my own email (and `about_me` within a meeting)

### DIFF
--- a/client/src/app/core/core-services/member.service.ts
+++ b/client/src/app/core/core-services/member.service.ts
@@ -4,7 +4,6 @@ import { TranslateService } from '@ngx-translate/core';
 import { Meeting } from 'app/shared/models/event-management/meeting';
 import { mediumDialogSettings } from 'app/shared/utils/dialog-settings';
 import { ViewUser } from 'app/site/users/models/view-user';
-import { Observable } from 'rxjs';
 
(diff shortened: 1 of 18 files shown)


emanuelschuetze:
A user manager cannot change the structure level (or number) of a user via user edit form. But the user meta quick edit dialog works.

GabrielInTheWorld:
@emanuelschuetze: Ping

emanuelschuetze (approved):

PR #648, opened 2021-11-17
Some improvements for the committee import

### LINKED ISSUE #647: Improve the import of committees
There are some features, that can be enhanced:

- [ ] Multiple managers for a committee won't work
- [ ] Break up an import if there are too many errors for one import-handler
- [ ] The rows in a preview will be exceeded if there are more than two entries in an array (for an array property)
- [ ] The import of a committee won't work, if its name has more than 256 characters

GabrielInTheWorld:
Fixes #647 

@emanuelschuetze: It is still under "work in progress", but you can already test it and give me a feedback, if anything does not work as you expect it.


emanuelschuetze (approved):
Looks good.

PR #656, opened 2021-11-22
Sorts meetings and adds links to them

### LINKED ISSUE #635: Improve the account dialog
- [x] Sort the meetings of a user in lexicographically order
- [x] Make a link to every user's meeting
- [x] Edit the own profile: On the orga-layer, only the fields `email` and `username` are allowed to change. Switching to a meeting, suddenly all fields are allowed to change... (Fixed by #644)

GabrielInTheWorld:
Fixes #635

### DIFF
--- a/client/src/app/site/common/components/start/start.component.ts
+++ b/client/src/app/site/common/components/start/start.component.ts
@@ -1,5 +1,6 @@
 import { Component, OnInit } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { NavigationEnd, Router } from '@angular/router';
 import { TranslateService } from '@ngx-translate/core';
 import { MeetingAction } from 'app/core/actions/meeting-action';
 import { ActiveMeetingService } from 'app/core/core-services/active-meeting.service';
@@ -7,9 +8,12 @@ import { OperatorService } from 'app/core/core-services/operator.service';
 import { Permission } from 'app/core/core-services/permission';
 import { MeetingRepositoryService } from 'app/core/repositories/management/meeting-repository.service';
 import { ComponentServiceCollector } from 'app/core/ui-services/component-service-collector';
-import { MeetingSettingsService } from 'app/core/ui-services/meeting-settings.service';
 import { ViewMeeting } from 'app/management/models/view-meeting';
 import { BaseModelContextComponent } from 'app/site/base/components/base-model-context.component';
+import { Observable } from 'rxjs';
+import { filter } from 'rxjs/operators';
+
+import { Settings } from '../../../../shared/models/event-management/meeting';
 
 /**
  * The start component. Greeting page for OpenSlides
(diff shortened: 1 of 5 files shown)


tsiegleauq:
> Hint: Currently, we have no routine to re-subscribe to the AU-service. Therefore, if a user stays on the start-page in any meeting and navigates to another meeting (by their account-dialog), they will see an empty page.

I suppose thats partially because if you dont reload a component, neither the constructor not the init will be called again. There would be a routine to renew the subscription, using the `this.localModelSubscriptions[subscriptionName]` from `base-model-context-component`

tsiegleauq (changes requested):
>  start-page in any meeting and navigates to another meeting (by their account-dialog), they will see an empty page.

I suppose this can be (and should) be done using routers. navigationEnd (https://angular.io/api/router/NavigationEnd) and some adjustments to the component.

emanuelschuetze (approved):
Ok for me. Please check the hint of @tsiegleauq

client/src/app/site/common/components/start/start.component.ts
```diff
@@ -57,15 +63,8 @@ export class StartComponent extends BaseModelContextComponent implements OnInit
...
-                this.startContent.welcome_title = welcomeTitle;
-            }),
-            this.meetingSettingsService.get(`welcome_text`).subscribe(welcomeText => {
-                this.startContent.welcome_text = this.translate.instant(welcomeText);
-            }),
-            this.activeMeetingService.meetingObservable.subscribe(meeting => {
-                this.meeting = meeting;
-            })
+            this.router.events.pipe(filter(event => event instanceof NavigationEnd)).subscribe(() => this.refresh()),
+            this.activeMeetingService.meetingObservable.subscribe(meeting => (this.meeting = meeting))
```
tsiegleauq, line 67: they are probably unconcurrent, so you can use `combineLatest` to limit the triggers

client/src/app/site/common/components/start/start.component.ts
```diff
@@ -76,7 +75,23 @@ export class StartComponent extends BaseModelContextComponent implements OnInit
...
+            this.meetingSettingsService.get(`welcome_title`).subscribe(welcomeTitle => {
+                this.startContent.welcome_title = welcomeTitle;
+            })
+        );
+        this.updateSubscription(
+            WELCOME_TEXT_SUBSCRIPTION_NAME,
+            this.meetingSettingsService.get(`welcome_text`).subscribe(welcomeText => {
+                this.startContent.welcome_text = this.translate.instant(welcomeText);
+            })
+        );
```
tsiegleauq, line 94: That might not be necessary. You can simply get the observable from the service and give it to the UI directly using async pipe and translate pipe. Youd also gain a translateable welcome text on the fly

tsiegleauq (changes requested):

client/src/app/site/common/components/start/start.component.ts
GabrielInTheWorld: It works better without `combineLatest`.

tsiegleauq (approved):

PR #659, opened 2021-11-22
Fixes the feature of pro/contra speech and contribution

### LINKED ISSUE #634: Handling of users on list of speakers
Achieve the state as in OS3 for the list of speakers (especially for users without the permission `list_of_speakers.can_manage`).

GabrielInTheWorld:
Fixes #634

Also displays the time-string at a projector in a local dateformat instead of the number of seconds.


tsiegleauq (approved):
ok. Not tested.

emanuelschuetze (approved):

PR #663, opened 2021-11-24
Fix redirection to error page in meetings

### LINKED ISSUE #626: A user who is not allowed to see a page is not redirected to another page or an "error" page
This contains two situations:
- [x] Going manually to `/motions/999` will result in an ugly motion-detail site as well as going to a "forbidden" motion (when the motion's state allows only the motion's submitters and users with the permission `motion.can_manage` to see the motion). These cases should show an "error" page (like "You are not allowed to see this page" or "This page could not be found")
- [ ] As delegate, going manually to `/settings/` within a meeting shows the settings-page although a delegate must not see that page. In such cases a user should be redirected to the startpage (or the first page, they are allowed to see).

tsiegleauq:
Fixes the redirection to the error handling page if a user accesses a
restricted page

fixes #626

### DIFF
--- a/client/src/app/core/core-services/active-meeting-id.service.ts
+++ b/client/src/app/core/core-services/active-meeting-id.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Router, RouterEvent, RoutesRecognized } from '@angular/router';
 import { BehaviorSubject, Observable } from 'rxjs';
-import { filter } from 'rxjs/operators';
+import { distinctUntilChanged, filter } from 'rxjs/operators';
 
 import { DataStoreService } from './data-store.service';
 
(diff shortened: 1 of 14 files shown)


tsiegleauq:
this is mose likely not the end of the line, but prevents some nasty infinity loops when trying to access restricted areas.

* The auth guard checks if OP can access, asks the "hasPerms", that would return false, then would link to the error page
* The error page would call canActivate, and since we are not in a meeting "hasPerms" would fail and link to the error page
* loop.

emanuelschuetze (changes requested):
redirect for /users does not work correctly:

example
https://localhost:8000/5/users redirects to https://localhost:8000/ instead of https://localhost:8000/5

tsiegleauq:
Redirecting to the meeting error page works now in more cases
Accessing non existing routes will forward you to / in the most parts instead of assuming it might be a meeting id.

routes that appear to be a meeting ID are still interpreted as such.

emanuelschuetze (approved):

PR #668, opened 2021-11-26
Adds the 'poll_backend' key to the poll-form

### LINKED ISSUE #655: Set poll/backend field
If I did not miss something, it is currently not possible to set the field [poll/backend](https://github.com/OpenSlides/openslides-backend/blob/e1f83b6807027ffac64134ee30740b8b4558443e/global/meta/models.yml#L2531-L2538)

When creating a poll, there should be a form field to decide, if it is a long running poll or a fast poll.

This feature is urged. It is currently not possible to test the postgres-backend of the vote service. See https://github.com/OpenSlides/openslides-vote-service/pull/66

See also the discussion in https://github.com/OpenSlides/openslides-backend/issues/1037 If we create the field `meeting/poll_default_backend`, then it will be also necessary to adapt the config page.

GabrielInTheWorld:
Fixes #655

Hint: The card for one assignment-poll could be adjusted so, that the poll_backend type can be read.

### DIFF
--- a/client/src/app/core/actions/poll-action.ts
+++ b/client/src/app/core/actions/poll-action.ts
@@ -1,6 +1,7 @@
 import { HasMeetingId } from 'app/shared/models/base/has-meeting-id';
 import { Identifiable } from 'app/shared/models/base/identifiable';
 
+import { PollBackendDurationType } from '../../shared/models/poll/poll-constants';
 import { Decimal, Fqid, Id } from '../definitions/key-types';
 
 export namespace PollAction {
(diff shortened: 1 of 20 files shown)


tsiegleauq (approved):
OK, not tested

GabrielInTheWorld:
Label: "Poll duration"

- [ ] "Short" or "Long" poll
- [ ] Add the label to a poll-detail card

emanuelschuetze (changes requested):
Retested. 
* I cannot change the poll backend in poll update dialog!
* please use in poll detail view the same label as in poll form (short/long poll, not fast/long)
* wish: please add the meta infomation of a motion poll also in the mat-card in motion detail view (as we do it already in assignment poll mat-card in assigment-detail-view)

emanuelschuetze (approved):

PR #676, opened 2021-11-30
Fixes: Implement the chat

### LINKED ISSUE #651: Implement the chat

GabrielInTheWorld:
Fixes #651 

This needs maybe OpenSlides/openslides-autoupdate-service#371

### DIFF
--- a/client/src/app/core/repositories/management/meeting-repository.service.ts
+++ b/client/src/app/core/repositories/management/meeting-repository.service.ts
@@ -63,6 +63,7 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
             `location`,
             `url_name`,
             `enable_anonymous`,
+            `enable_chat`,
             `is_template`,
             `default_group_id`, // needed for adding users
             `jitsi_domain`,
@@ -85,7 +86,7 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
             dashboard: dashboardFields,
             settings: this.meetingSettingsDefinitionProvider
                 .getSettingsKeys()
-                .concat(`jitsi_domain`, `jitsi_room_name`, `jitsi_room_password`),
+                .concat(`jitsi_domain`, `jitsi_room_name`, `jitsi_room_password`, `enable_chat`),
             startPage: startPageFields,
             preview: previewFields
         };
--- a/client/src/app/site/chat/components/chat-group-list/chat-group-list.component.html
+++ b/client/src/app/site/chat/components/chat-group-list/chat-group-list.component.html
@@ -1,0 +1,62 @@
+<os-head-bar [hasMainButton]="hasMainButton" (mainEvent)="createChatGroup()">
+    <!-- Title -->
+    <div class="title-slot">
+        {{ 'Chat' | translate }}
+    </div>
+</os-head-bar>
+
+<mat-card
+    *ngIf="chatGroupsObservable | async as chatGroups"
+    class="os-card chat-group-list-card"
+    [ngClass]="{ 'fit-height': !chatGroups?.length }"
+>
+    <div class="padding-16" *ngIf="!chatGroups?.length">
+        <span>
+            {{ 'No chat groups available.' | translate }}
+        </span>
+    </div>
+    <os-vertical-tab-group
+        *ngIf="chatGroups?.length"
+        #verticalTabGroup
+        [labelHeight]="56"
+        [showAlwaysContentHeader]="true"
+        (contentStateChanged)="onTabContentChanged($event)"
+    >
+        <ng-template osVerticalTabGroupLabelHeader>
+            <div
+                class="flex-vertical-center padding-left-16 padding-right-16 full-width full-height icon fake-disabled"
+            >
+                <mat-icon>groups</mat-icon>
+                <span class="padding-left-8">{{ 'Chat groups' | translate }}</span>
+            </div>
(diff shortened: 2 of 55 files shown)


emanuelschuetze (changes requested):
Tested. Still open:
- [ ] always show usernames for messages
- [x] always show new chat group (for admin) even if admin has not set any read/write groups (-> currently empty new groups are not listed in client!)
- [x] add chat group name also above chat message box
- [x] encrease chat message box window (use width of default mat cards + chatgroup menu) 
- [x] some style changes (as discussed)

client/src/app/core/repositories/management/meeting-repository.service.ts
```diff
@@ -85,7 +86,7 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
             dashboard: dashboardFields,
             settings: this.meetingSettingsDefinitionProvider
                 .getSettingsKeys()
-                .concat(`jitsi_domain`, `jitsi_room_name`, `jitsi_room_password`),
+                .concat(`jitsi_domain`, `jitsi_room_name`, `jitsi_room_password`, `enable_chat`),
```
tsiegleauq, line 89: I don' really get why this is an option. Chat could just be enabled all the time, no?

client/src/app/site/chat/components/chat-group-list/chat-group-list.component.html
```diff
@@ -0,0 +1,53 @@
...
+    <os-vertical-tab-group
+        *ngIf="chatGroups?.length"
+        [labelHeight]="56"
+        [showAlwaysContentHeader]="true"
+        (contentStateChanged)="onTabContentChanged($event)"
+    >
+        <ng-template osVerticalTabGroupLabelHeader>
+            <div
+                class="flex-vertical-center padding-left-16 padding-right-16 full-width full-height icon fake-disabled"
+            >
```
tsiegleauq, line 27: the directive `osVerticalTabGroupLabelHeader` is completely empty and can probably be deleted

client/src/app/site/chat/services/chat-notification.service.ts
```diff
@@ -0,0 +1,109 @@
...
+        return this._chatGroupNotificationsSubject.asObservable();
+    }
+
+    private _chatGroupNotificationsSubject = new BehaviorSubject<NotificationAmountEvent>({});
+    private _chatGroupLastSeenObject: LastSeenTimestampEvent = {};
+    private _openChatGroupIds: Id[] = [];
+    private _lastMessageAmount = 0;
+    private _chatMessages: ViewChatMessage[] = [];
+
+    public constructor(private _storage: StorageService, private _chatMessageRepo: ChatMessageRepositoryService) {
```
tsiegleauq, line 41: avoid prefixing simple variables with `_` unless absolutely necessary

client/src/app/site/chat/services/chat.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    private _canSeeChatSubject = new BehaviorSubject<boolean>(false);
+    private _canSeeSomeChatGroup = false;
+    private _canManage = false;
+    private _isChatEnabled = false;
+
+    public constructor(
+        _chatGroupRepo: ChatGroupRepositoryService,
+        _operator: OperatorService,
+        _meetingSettingService: MeetingSettingsService,
+        _orgaSettingService: OrganizationSettingsService
```
tsiegleauq, line 27: visibility not set

tsiegleauq (changes requested):

emanuelschuetze:
Retested:
 * [ ] linebreak for long group names
 * [ ] group notifcation counter
 * [ ] missing usernames

client/src/app/core/repositories/management/meeting-repository.service.ts
GabrielInTheWorld: This should be discussed with @emanuelschuetze.

client/src/app/site/chat/components/chat-group-list/chat-group-list.component.html
GabrielInTheWorld: The component `VerticalTabGroupComponent` reads the `<ng-template />` by this directive.

tsiegleauq (approved):
Code looks fine, there are some things I would change, but thats not too important and more "refinement".

- The "new chat group" button does not really behave that well. In mobile mode, it would simply disappear when you are "in" a chat group. Thats understandable but extremly unpredictable and tedious to use. Perhaps put it as a "plus" icon in the grey chat box area
image

- the view-port-observer seems to behave a little weird. You can scale down from desktop view to mobile view to have a reduced list, if you scale it up again you keep the "mobile" view.
image

- I suppose the ton of white space is to have the stream open? If this is a direct requirement we should talk about this. You can collapse jitsi and the live stream any time.
image

- the width of the chat-card feels weird since its larger than i.e the typical card for motions or assignments, but smaller than the average list view. Since this view scales pretty well I'd go full with in the normal desktop resolution. We should also go full height to reduce this white bar
 
image

- some slight animations would do wonders to make the chat feel more responsive.

GabrielInTheWorld:
These are good suggestions! Maybe @emanuelschuetze tells what he think about these behaviors. I suggest to make an own issue for that to discuss this.

emanuelschuetze (approved):
Good enough for this pr. later we can improve the chat (ui).

PR #682, opened 2021-12-02
Exclude non natural users from call restrictions

### LINKED ISSUE #632: "Pin" a user in the jitsi room
Currently, when using the jitsi LOS restricion, users who cannot manage speakers would be forces to leave the jitsi room if current-list-of-speakers changed.
It was desired to "pin" certain users so they would stay inside the room although they do not have `manage_speaker`.
This "can" work over client exclusive data but would be entirely stateless, therefore unpredictable. The better approach requires to alter or recycle information from the user model.

**Alternatives:**
- abuse some information that is already there
- just use jitsi externally

tsiegleauq:
Allows non natural users from using restricted jitsi channels as they
please. They would not be kicked nor prevented from entering the call

fixes #632


emanuelschuetze (approved):
Nice feature!

PR #688, opened 2021-12-06
Add invite-to-call to LOS

tsiegleauq:
Adds functions to invite and kick users to a jitsi call via LOS

### DIFF
--- a/client/src/app/core/core-services/notify.service.ts
+++ b/client/src/app/core/core-services/notify.service.ts
@@ -193,7 +193,7 @@ export class NotifyService {
 
     /**
      * Sends a notify message to all open clients with the given users logged in.
-     * @param name The name of th enotify message
+     * @param name The name of the enotify message
      * @param content The payload to send.
      * @param users Multiple user ids.
      */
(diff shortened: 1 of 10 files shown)


tsiegleauq:
following issue needs to be solved:
https://github.com/OpenSlides/openslides-icc-service/issues/30

following PR has to be merged:
https://github.com/OpenSlides/OpenSlides/pull/6305

emanuelschuetze (changes requested):
Tested.
* Admin A who invites user B gets also the invite dialog.
* user B gets in every tab the invite dialog. After accept this all dialogs in other tabs should closed.

tsiegleauq:
> Admin A who invites user B gets also the invite dialog.

Fixed it. Apparently the sender of ICC messages will always receive a copy of their own messages.

> user B gets in every tab the invite dialog. After accept this all dialogs in other tabs should closed.

This might not be easy atm. indexedDB cannot detect changes on another tab, see
https://github.com/cyrilletuzi/angular-async-local-storage/issues/386

There is this browser API ( https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API ), but it seems to fairly new ( https://caniuse.com/?search=Broadcast )

I'll have a look at it.

tsiegleauq:
Now with broadcast.service.ts for real time inter-tab-communication

GabrielInTheWorld (approved):

emanuelschuetze (approved):

PR #692, opened 2021-12-07
Fixes offline banner in firefox

### LINKED ISSUE #652: Offline banner in firefox in idle state
In our demo (https://demo.os4.openslides.com) we can observe, that an offline banner is appearing after ~1min. Until now, we have determines three conditions which have to be true to see this behavior:

- Use FireFox
- Do nothing for at least one minute
- Be in our [demo](https://demo.os4.openslides.com)

The same behavior may occur, using a productive setup instead of a development setup, but this has to be investigated.

GabrielInTheWorld:
I could reproduce the (wrong) behavior in a local prod setup. Before an offline banner is shown, this will check if we are really offline and then, only then, an offline banner is shown.

Fixes #652.


emanuelschuetze (commented):
not tested

PR #709, opened 2021-12-15
Automatically forward users to a unique meeting

tsiegleauq:
If a user has only one active meeting, forward them directly after
login. Dashboard component cannot be accessed

### DIFF
--- a/client/src/app/site/users/models/view-user.ts
+++ b/client/src/app/site/users/models/view-user.ts
@@ -68,6 +68,10 @@ export class ViewUser extends BaseProjectableViewModel<User> implements Searchab
         return DelegationType.Neither;
     }
 
+    public get meeting_ids(): Id[] {
+        return this.user.meeting_ids;
+    }
+
     // Will be set by the repository
     public getName: () => string;
     public getShortName: () => string;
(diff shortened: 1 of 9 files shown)


client/src/app/site/users/models/view-user.ts
```diff
@@ -89,6 +89,21 @@ export class ViewUser extends BaseProjectableViewModel<User> implements Searchab
...
+        const meetingAmount = this.user.is_present_in_meeting_ids?.length || 0;
+        if (meetingAmount === 1) {
+            return this.user.is_present_in_meeting_ids[0];
+        } else if (meetingAmount > 1) {
+            throw new Error(`User has multiple meetings`);
+        } else if (meetingAmount === 0) {
+            throw new Error(`User has no meetings at all`);
+        }
+    }
+
```
GabrielInTheWorld, line 106: I know, this PR is still "work in progress" and perhaps you have that already in mind, but one hint:
`user.is_present_in_meetings` refers only to the meetings a user is present in. This means, there could be meetings, a user has access to, but is currently not as "present" marked.

client/src/app/site/users/models/view-user.ts
```diff
@@ -68,6 +68,10 @@ export class ViewUser extends BaseProjectableViewModel<User> implements Searchab
         return DelegationType.Neither;
     }
 
+    public get meeting_ids(): Id[] {
+        return this.user.meeting_ids;
+    }
+
```
GabrielInTheWorld, line 74: Is it necessary, that a `ViewUser` has this property? Usually, properties of a `User` are treated as properties of a `ViewUser`, too.

GabrielInTheWorld (approved):
Looks good! One question, but this is independent of my approvement.

client/src/app/site/users/models/view-user.ts
tsiegleauq: more or less. The pseudo-automatic transaction from properties from `User` into `ViewUser` has issues that prevent automatic testing quite heavily, thus declaring is implicit seems a lot safer. There might be other solutions on how ViewModels were created.

emanuelschuetze (approved):

PR #712, opened 2021-12-16
Re-Sort settings

tsiegleauq:
Settings are now sorted like they were in OS3


GabrielInTheWorld (approved):

emanuelschuetze (approved):
restructured some more settings

PR #714, opened 2022-01-06
Improves importing of accounts

### LINKED ISSUE #619: Recognize duplicates of accounts and merge them together
If a new account is imported (e.g. by our CSV-importer) or is created, then a routine should recognize if such an account already exists (-> "duplicate").
There is no such unique field, that we can easily use to indicate a "duplicate", because people can have similar names and it is guessing a username is not so easily, too. Therefore, the field "email address" could be a good indicator to recognize duplicates, but maybe not the best. So, we need to find a good indicator.

After we found a good indicator and duplicates, a routine should be implemented, that merge duplicates into the already existing accounts. Maybe, this is a task for the backend.

### LINKED ISSUE #666: Enhance the CSV-import of users by using the new presenter
The backend provides a new presenter to find a user by submitting a first_name, last_name and/or email. If a user was found, the operating user should decide if the found should be unified with the submitted one. This means: Update fields of an existing user.

GabrielInTheWorld:
An account will be merged to an existing one, if one already exists. In a meeting, an existing account will be added to the meeting's default group instead of creating a participant.

Fixes #666 
Fixes #619

### DIFF
--- a/client/src/app/core/core-services/key-transforms.ts
+++ b/client/src/app/core/core-services/key-transforms.ts
@@ -3,6 +3,10 @@ import { Collection, Field, Fqfield, Fqid, Id } from '../definitions/key-types';
 export const KEYSEPERATOR = `/`;
 export const TEMPLATE_FIELD_INDICATOR = `$`;
 
+export function copy<T>(model: T, modelHeaders: (keyof T)[]): T {
+    return modelHeaders.mapToObject(header => ({ [header]: model[header] })) as T;
+}
+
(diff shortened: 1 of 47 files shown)


emanuelschuetze (changes requested):
Please add a second line in summary for user import:
  ... will be created
  ... will be referenced

Start with '1' in preview table.

emanuelschuetze (approved):

emanuelschuetze:
ng lint failed

PR #746, opened 2022-01-12
Show start and end time in pdf doc header

tsiegleauq:
Shows the start and end time in the doc PDF


emanuelschuetze (changes requested):
Use same start/end date -> pdf header shows "12.1.2022 - 12.1.2022" instead of "12.1.2022"!

emanuelschuetze (approved):

PR #785, opened 2022-01-26
Maybe fixes: Being disconnected after some idle time (#784)

### LINKED ISSUE #784: Disconnection after some time in idle state
If a user does nothing for at least 10 - 15 minutes, they don't get updates anymore.
Maybe, this happen, because the AU-service has for a user a ticket, that expires after some time (~10 minutes). If the user's client gets a new ticket, the AU-service doesn't get it. Therefore, the AU-service does not know, that the user is still authenticated and still allowed to get updates.

GabrielInTheWorld:
After refreshing an access token every active connection will be re-established to "ping" the autoupdate system.

Fixes #784

### DIFF
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -98,7 +98,7 @@ export class AppComponent implements OnInit {
         await this.onInitDone;
 
         setTimeout(() => {
-            this.lifecycleService.appLoaded.next();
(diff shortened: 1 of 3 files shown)


GabrielInTheWorld:
Closed in favor of OpenSlides/openslides-autoupdate-service#404

PR #803, opened 2022-02-04
Map the ids of every displayable model to its specified `sequential_number` (#733)

### LINKED ISSUE #733: Request a model (with a detail-view) depending on its sequential number and a meeting's id
This can be accomplished maybe by doing these steps:

1. When in a meeting: Request all displayable models (displayable: with a detail-view) and their sequential numbers
2. List them in a dictionary like "{ [collection: string]: { [ <meeting_id/model.sequential_number> ]: model/id } }"
3. When navigating to a detail-view, we can easily get the id of a model by checking this dictionary
4. A further benefit of this approach would be, that if <meeting_id/model.sequential_number> returns `undefined`, the requested model does not exist.

It has to be waited for OpenSlides/openslides-backend#1130, since only motions have a sequential number at the moment.

GabrielInTheWorld:
Fixes #733

### DIFF
--- a/client/src/app/app.component.ts
+++ b/client/src/app/app.component.ts
@@ -13,6 +13,7 @@ import { OfflineService } from './core/core-services/offline.service';
 import { OpenSlidesService } from './core/core-services/openslides.service';
 import { OpenSlidesStatusService } from './core/core-services/openslides-status.service';
 import { OperatorService } from './core/core-services/operator.service';
+import { SequentialNumberMappingService } from './core/core-services/sequential-number-mapping.service';
 import { ServertimeService } from './core/core-services/servertime.service';
 import { Deferred } from './core/promises/deferred';
 import { CountUsersService } from './core/ui-services/count-users.service';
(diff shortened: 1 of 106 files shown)


emanuelschuetze (changes requested):
error after creating a new motion in a new meeting (there are some other meetings with motions):
`The requested path contains undefined segment at index 2`

emanuelschuetze (changes requested):
Try to create a new amendment for a motion (here: motion/1):
  
`Error: The following models do not belong to meeting 15: ['motion/1']`

There is the wrong motion text shown in the paragraph choice list!

emanuelschuetze (changes requested):
Try to create a new topic item or assignment:
`TypeError: Cannot read properties of undefined (reading 'length')`
-> tinymce editor is not loaded correctly
(motion works)

GabrielInTheWorld:
Waiting for OpenSlides/openslides-backend#1013.

emanuelschuetze (approved):
Ok. Wait for backend.

emanuelschuetze:
Rebased

emanuelschuetze (approved):
Retested

PR #806, opened 2022-02-07
Limit size of error messages in snackBar (#805)

### LINKED ISSUE #805: Error display on faulty import makes client unusable
During a faulty import (malformed meeting), I got a huge error message (> 5000 characters) from the backend.
This was displayed in the popup "error" display.
The "ok" button to dismiss disappeared from screen due to the error being inside a huge "span", resulting in a temporarily unusable client, as the error popup didn't disappear after a while, but kept waiting for cnfirmation.

jkr-os:
Suggested solution to 805: Limiting the error messages in the matSnackbar. I reused the "getLongPreview" function to limit the error message to a size of 1500. In those cases an end user probably just gets the impression "something is completely wrong".

Another point of discussion : The base-import-list.component imports getLongPreview without actually using it. Is this just some leftover debris? getLongPreview seems to be currently orphaned (getShortPreview is being used).

Another approach would be to make this huge error disappear after some timeout (the error service disables the matSnackbar dismiss timeout, and my guess is that is on purpose)


jkr-os:
the code style definition and test error messages could be more explicit and/or informative, but I think I have it the way it is required now.

please review as this is a first commit; also if this behavior is wanted at all.

PR #818, opened 2022-02-09
Fixes: Brings broadcast-channel to webkit browsers (#812)

### LINKED ISSUE #812: WebKit based browsers (like Safari) are unusable caused by the missing Broadcast API
[StackOverflow](https://stackoverflow.com/questions/57700474/how-can-i-use-the-broadcastchannel-api-or-something-similar-in-safari-10)

GabrielInTheWorld:
Fixes #812 

Attention: Try this with caution, because there is a new npm package to install!

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -32,6 +32,7 @@
                 "@tinymce/tinymce-angular": "^4.2.4",
                 "@videojs/http-streaming": "^2.9.1",
                 "acorn": "^8.0.1",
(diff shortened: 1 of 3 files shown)


alveolus:
Works! Tested with Safari 15.2 (17612.3.6.1.6), macOS Monterey 12.1 (21C52)

alveolus (approved):
Works! Tested with Safari 15.2 (17612.3.6.1.6), macOS Monterey 12.1 (21C52)

PR #867, opened 2022-02-23
Fix: Sets vote_weight to 0

### LINKED ISSUE #866: Cannot set a user's vote weight to `0`

GabrielInTheWorld:
Fixes #866


client/src/app/site/users/components/user-detail/user-detail.component.ts
```diff
@@ -115,7 +115,11 @@ export class UserDetailComponent extends BaseModelContextComponent implements On
     public readonly users: BehaviorSubject<ViewUser[]> = new BehaviorSubject<ViewUser[]>([]);
 
     public get showVoteWeight(): boolean {
-        return this.pollService.isElectronicVotingEnabled && this._voteWeightEnabled;
+            return (
```
emanuelschuetze, line 118: looks like this indentation is not correct

emanuelschuetze, line 118: looks like this indentation is not correct

emanuelschuetze (approved):

Issue #808, opened 2022-02-08
Date and filesize of a mediafile is not correctly displayed


GabrielInTheWorld:
I _think_ the date of a mediafile is given in seconds but in the client accepted as ms. The filesize is not in a "human readable" format displayed.

bug0001:
Additions - the problem is present in the file list view, but not in the upload.

Solution

client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.html

            <div class="foreground-icon" *ngIf="!mediafile.is_directory">
                <span>{{ getDateFromTimestamp(mediafile.create_timestamp) }} · {{ mediafile.filesize }}</span>
            </div>

this needs either a pipe (framework terminology) example mediafile.size | h... or h(file.size)

Except we already have that in another part of the application

client/src/app/shared/components/media-upload-content/media-upload-content.component.html

                <mat-icon [inline]="true">data_usage</mat-icon>
                {{ getReadableSize(file.mediafile.size) }}



client/src/app/shared/components/media-upload-content/media-upload-content.component.ts

    /**
     * Converts a file size in byte into human readable format
     *
     * @param bytes file size in bytes
     * @returns a readable file size representation
     */
    public getReadableSize(bytes: number): string {
        if (bytes === 0) {
            return `0 B`;
        }
        const unitLevel = Math.floor(Math.log(bytes) / Math.log(1024));
        bytes = +(bytes / Math.pow(1024, unitLevel)).toFixed(2);
        return `${bytes} ${[`B`, `kB`, `MB`, `GB`, `TB`][unitLevel]}`;
    }

bug0002:
pr #872 solves half the problem

the current view being 

x/x/1970, 2:x:x AM · 35.xx kB

bug0002:
Once this pr is merged the next steps would be :

- Removing the local function from the upload view, using the general pipe there
- Doing the date thing

bug0002:
openslides-client/client/src/app/site/mediafiles/models/view-mediafile.ts

    public get timestamp(): string {
        return this.mediafile.create_timestamp ? this.mediafile.create_timestamp : null;
    }


openslides-client/client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.ts
I'm leaving my debugging notes here :+1: 
    public getDateFromTimestamp(timestamp: string): string {
        return new Date(timestamp).toLocaleString(this.translate.currentLang);
    }






[...]

PR #872, opened 2022-02-24
Human readable file size #808

bug0002:
custom pipe h file size 808

### DIFF
--- a/client/src/app/shared/pipes/hfilesize.pipe.ts
+++ b/client/src/app/shared/pipes/hfilesize.pipe.ts
@@ -2,9 +2,9 @@ import { Pipe, PipeTransform } from '@angular/core';
 
 
 @Pipe({
-    name: `hSize`
+    name: `readableBytes`
 })
-export class HSize implements PipeTransform {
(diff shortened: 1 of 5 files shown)


client/src/app/shared/pipes/hfilesize.pipe.ts
```diff
@@ -0,0 +1,16 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+
+@Pipe({
+    name: `hSize`
```
GabrielInTheWorld, line 5: Please name it more intuitive, something like `readableBytes` or so.

client/src/app/shared/pipes/hfilesize.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+})
+export class HSize implements PipeTransform {
```
GabrielInTheWorld, line 7: Change the name accordingly to the directive-name and suffix it with `Pipe`.

GabrielInTheWorld (changes requested):

GabrielInTheWorld (approved):

PR #879, opened 2022-02-25
Fixes user creating and voting

### LINKED ISSUE #877: Cannot create a participant/an account
- [ ] If the vote_weight is enabled, the view to create a participant crashes
- [ ] Creating an account sends a payload with `default_vote_weight: ".00000"

### LINKED ISSUE #878: After voting as a delegate, vote options do not disappear

GabrielInTheWorld:
- Users can be created even if vote_weight is enabled (Fixes #877)
- After voting vote options will disappear (Fixes #878)


GabrielInTheWorld:
Closed in favor of #881.

PR #883, opened 2022-02-25
Can set voteweight to null

### LINKED ISSUE #882: Cannot set vote_weight to `null`

GabrielInTheWorld:
Fixes #882.

Waits for OpenSlides/openslides-backend#1242.


emanuelschuetze (changes requested):
Set voteweight for user to '0'. Save user form. Reopen user form => field 'voteweight' is not visible!

bastianjoel (changes requested):
I think shared/utils/to-decimal.ts is not needed. The functions are already defined in infrastructure/utils/functions.ts

bastianjoel (approved):

MSoeb (approved):

PR #887, opened 2022-02-28
date (#808)

bug0002:


client/src/app/site/mediafiles/components/mediafile-list/mediafile-list.component.ts
```diff
@@ -280,7 +280,8 @@ export class MediafileListComponent extends BaseListViewComponent<ViewMediafile>
     }
 
     public getDateFromTimestamp(timestamp: string): string {
-        return new Date(timestamp).toLocaleString(this.translate.currentLang);
+        const timestamptoms = Number(timestamp) * 1000;
+        return new Date(timestamptoms).toLocaleString(this.translate.currentLang);
     }
```
GabrielInTheWorld, line 285: As we discussed, here are multiple things to do:

1. Change the type of `timestamp` from `string` to `number`
2. Change the type of `mediafile.create_timestamp` from `string` to `number`
3. Write this in one line: `return new Date(timestamp * 1000).toLocaleString(this.translate.currentLang);`

You can check, if there are multiple places, where a timestamp is given and multiplied with `1000`. Then, you could convert this block to a pipe and add this pipe to the HTML-template.

bug0002:
To sum up : a solution was implemented based on the legacy code (with string input). This was reverted.

An architecture change was then requested above.

in mediafile-list.component.ts :

public getDateFromTimestamp(timestamp: string): string {

in view-mediafile.ts :

public get timestamp(): string {

in openslides-client/client/src/app/shared/models/mediafiles/mediafile.ts :

public create_timestamp: string;

I've made relevant changes (to Typescript number).

No further instances of "timestamp: string" signatures were found. Fails to compile.

PR #928, opened 2022-03-10
Remove archive banner (#796)

### LINKED ISSUE #796: After signing out, the banner "Archived" is displayed

bug0002:
Fixes #796

### DIFF
--- a/client/src/app/core/ui-services/banner.service.ts
+++ b/client/src/app/core/ui-services/banner.service.ts
@@ -82,4 +82,8 @@ export class BannerService {
             this.activeBanners.next(newBanners);
         }
     }
+
+    public clear(): void {
+        this.activeBanners.next([]);
+    }
 }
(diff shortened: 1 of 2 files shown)


GabrielInTheWorld (changes requested):
The code looks good!
Two remarks:
1. Change the title, so the number of the related issue is not shown or in paranthesis linked (by a `#`)
2. Write in the description something like "Fixes #...", so everyone knows what this PR will solve

emanuelschuetze (changes requested):
Tested: archive banner is still visible after logout.

bug0002:
First review request should be done. Second : can confirm the issue...

bug0002:
Debugging, keeping notes :

    public clear(): void {
        console.log('clear');

so that the message "clear" is displayed when the clear function is called

and

    public addBanner(toAdd: BannerDefinition): void {
        console.log('addBanner', toAdd);

When reproducing the issue, it becomes obvious that addBanner is called after clear (on devtools console), resulting in above.

Further :

addBanner appears in two files, banner.service.ts (definition), and active-meeting.service.ts where the setupActiveMeeting function is relevant

PR #948, opened 2022-03-17
Some fixes for motions, motion blocks and working with agenda items

### LINKED ISSUE #944: Working with motion blocks does not work
- [x] Creating a motion block also creates a random default motion, attached to the motion block
- [x] Assigning a motion to a motion block does not work
- [x] To a motion block assigned motions are not displayed in the detail view of a motion block (in the list view is displayed, that there are motions assigned)

### LINKED ISSUE #945: When creating a motion, it cannot be added to the agenda
- [x] Reloading on the startpage -> go to motions -> create a motion -> there are no parent elements for the "add-to-agenda"-menu
- [x] There is sometimes no dropdown-menu to choose a parent element for the agenda
- [x] After a parent element was chosen, the label does not float above the chosen element. So, it seems, that no element was chosen.

### LINKED ISSUE #946: A motion block cannot be added as a child to an agenda item
- Edit metadata of a motion block
- Mark it as "add to agenda"
- Try to select a parent element

### LINKED ISSUE #949: A motion is not removed from a motion block within its detail view
When on the detail view of a motion block and trying to remove a motion, the motion won't disappear.

GabrielInTheWorld:
- Fixes working with motion blocks (Fixes #944)
- Fixes working with agenda items (Fixes #945)
- A motion block can be added to the agenda as a child item of another agenda item (Fixes #946)
- Fixes removing a motion from a block (Fixes #949)

### DIFF
--- a/client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
+++ b/client/src/app/core/repositories/agenda/agenda-item-repository.service.ts
@@ -1,6 +1,10 @@
 import { Injectable } from '@angular/core';
 import { AgendaItemAction } from 'app/core/actions/agenda-item-action';
-import { DEFAULT_FIELDSET, Fieldsets } from 'app/core/core-services/model-request-builder.service';
+import {
+    DEFAULT_FIELDSET,
+    Fieldsets,
+    SimplifiedModelRequest
+} from 'app/core/core-services/model-request-builder.service';
 import { Id } from 'app/core/definitions/key-types';
 import { MeetingSettingsService } from 'app/core/ui-services/meeting-settings.service';
 import { TreeIdNode, TreeService } from 'app/core/ui-services/tree.service';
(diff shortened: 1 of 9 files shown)


emanuelschuetze (changes requested):
Tested:
* Create new motion block and add to agenda. Then go to agenda list and try to remove the motoin block from agenda -> "Error: data.id must be integer or null"
* Create new motion block and add a motion to this block. Then go to motion block detail view and try to remove the motion from the block -> Nothing happens. Reload required

GabrielInTheWorld:
> Tested:
> 
> * Create new motion block and add to agenda. Then go to agenda list and try to remove the motoin block from agenda -> "Error: data.id must be integer or null"

Okay, I will check this.

> * Create new motion block and add a motion to this block. Then go to motion block detail view and try to remove the motion from the block -> Nothing happens. Reload required

This is #949.

emanuelschuetze (approved):

PR #972, opened 2022-03-29
Fixes some bugs

### LINKED ISSUE #969: Autopilot hangs, if submitter list is empty/`null`

### LINKED ISSUE #970: PDF exporting of motions do not work because no text is stored

### LINKED ISSUE #971: Participated vote rights do not work

### LINKED ISSUE #974: There are sometimes groups with the same name

GabrielInTheWorld:
- Autopilot does not hanging when there aren't any submitters (Fixes #969)
- Exporting a motion-list gets the text of every motion (Fixes #970)
- Delegations for polls work (Fixes #971)
- In the selection form for entitled groups every group will appear only once (Fixes #974)


emanuelschuetze (changes requested):
pdf list export does not contain voting results (motion and assignment)

emanuelschuetze (approved):

PR #986, opened 2022-04-11
Implemented a create wizard to add participants to a meeting (#667)

### LINKED ISSUE #667: Enhance the updating and creating of users
Both use the backend-actions for creating or updating the users. The client calls the presenter with proposed username/email-tuples and decides or let the user decide to call user.create or user.update. This can be used in various scenarios like creating new users, adding users to a meeting or give them CML-rights. With the payload the client has to send a scope (Organization, Committee or Meeting) and in case of the latter 2 an id to check permissions.

GabrielInTheWorld:
Fixes #667

### DIFF
--- a/client/src/app/core/core-services/presenters/get-user-scope-presenter.service.ts
+++ b/client/src/app/core/core-services/presenters/get-user-scope-presenter.service.ts
@@ -13,11 +13,13 @@ interface PresenterPayload {
     user_ids: Id[];
 }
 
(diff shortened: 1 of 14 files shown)


GabrielInTheWorld:
Still todo: Unify create and edit template to the new wizard. The old UserDetailComponent should only be a "view" template.

emanuelschuetze (approved):
Looks good now!

PR #988, opened 2022-04-11
Fixes creating of change-recommendations (#987)

### LINKED ISSUE #987: Broken start and end line numbers for change recommendation creation

GabrielInTheWorld:
Fixes #987.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .DS_STORE
 
-/dist/
+dist/
 /bazel-out
 /integration/bazel/bazel-*
 .angular/cache
--- a/client/src/app/core/repositories/motions/motion-change-recommendation-repository.service.ts
+++ b/client/src/app/core/repositories/motions/motion-change-recommendation-repository.service.ts
@@ -147,7 +147,7 @@ export class MotionChangeRecommendationRepositoryService extends BaseRepositoryW
 
         const changeReco: Partial<MotionChangeRecommendationAction.CreatePayload> = {};
         changeReco.line_from = lineRange.from;
-        changeReco.line_to = lineRange.to;
+        changeReco.line_to = lineRange.to - 1;
         changeReco.type = ModificationType.TYPE_REPLACEMENT;
         changeReco.text = this.diffService.extractMotionLineRange(motionText, lineRange, false, lineLength, null);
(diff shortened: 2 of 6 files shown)


jsangmeister:
Since this is in "Review in progress": All places evaluating `line_to` of change recommendations must be changed to, they assume that it is exclusive as well. Otherwise the diff view will probaby be broken.

emanuelschuetze (changes requested):
Does not work for a single line (e.g. try to cretae a cr for line 2 -> editor is empty -> enter something -> line 2 exists 2 times...)

GabrielInTheWorld:
@CatoTH: Currently, the linenumber algorithm expects the format [from, to) for changed linenumbers. The backend, however, detects collisions of linenumbers, if there is already one change recommendation for a linenumber, which will be changed by another change recommendation. This is especially the case, if a change recommendation should be created and regardless whether a linenumber is excluded in the linenumber algorithm. 
Therefore, we have the following options:
- Either we adjust every time we get a change recommendation or send an update/create of a change recommendation the `to` field
- Or we adjust the linenumber algorithm to handle `[from, to]`

CatoTH:
In principle, I don't have very strong opinions about this. The reason for the implementation as it is: it actually refers to the line numbers, not the lines. That is, to change line 3, the algorithm extracts all in between line number 3 and line number 4, which is line 3. One could change this, you would just need to make sure that the "+1" is done whenever the line extraction and diff algorithms are applied.

GabrielInTheWorld:
> One could change this, you would just need to make sure that the "+1" is done whenever the line extraction and diff algorithms are applied.

Then, could you do this please?

jsangmeister:
> * Or we adjust the linenumber algorithm to handle `[from, to]`

This was also what I had in mind when changing the backend and what we talked about @GabrielInTheWorld - we should keep this consistent.

GabrielInTheWorld:
> > * Or we adjust the linenumber algorithm to handle `[from, to]`
> 
> This was also what I had in mind when changing the backend and what we talked about @GabrielInTheWorld - we should keep this consistent.

Yes, this also my preferred choice.

emanuelschuetze:
the new PR #1012 contains the commit of this PR. -> close this PR.

PR #993, opened 2022-04-14
Adjust PDF export and add configuration options

m-schieder:
* Makes the line height and the page margins configurable. 
* Adjust margins to save space.

The default values are not set after the first start, so all values have to be set via the GUI first.  I haven't figured out how to solve this yet.

The following changes must be made in the [backend](https://github.com/OpenSlides/openslides-backend/pull/1300).


client/src/app/core/repositories/management/meeting-settings-definition.ts
```diff
@@ -117,12 +117,9 @@ export const meetingSettings: SettingsGroup[] = [
...
                         label: _(`Standard font size in PDF`),
-                        type: `choice`,
-                        choices: {
-                            10: 10,
-                            11: 11,
-                            12: 12
-                        }
+                        type: `integer`,
+                        validators: [Validators.min(10), Validators.max(12)],
+                        helpText: _(`Available sizes are 10 ,11 and 12.`)
```
emanuelschuetze, line 122: Please check if this needs a database migration in backend.

emanuelschuetze (approved):
One remark. Rest looks and works good.

client/src/app/site/motions/services/motion-pdf.service.ts
```diff
@@ -551,7 +552,8 @@ export class MotionPdfService {
         motion: ViewMotion,
         lineLength: number,
         lnMode: LineNumberingMode,
-        crMode: ChangeRecoMode
+        crMode: ChangeRecoMode,
+        lineHeight: number
```
GabrielInTheWorld, line 556: Please pass an object or an interface, which holds the data.

client/src/app/core/pdf-services/pdf-document.service.ts
```diff
@@ -116,6 +116,17 @@ export class PdfDocumentService {
         return url.split(`/`).pop();
     }
 
+    /**
+     * Calculates millimeters in points for pdfmake.
+     *
+     * @param millimeters
+     * @return points
+     */
+    private mmToPoints(mm: number): number {
```
GabrielInTheWorld, line 125: When does it be invoked?

GabrielInTheWorld (changes requested):

client/src/app/core/repositories/management/meeting-settings-definition.ts
jsangmeister: No migration needed since the actual values did not change.

client/src/app/core/pdf-services/pdf-document.service.ts
GabrielInTheWorld: Ok, didn't see this.

GabrielInTheWorld (approved):

PR #994, opened 2022-04-16
Some improvements for motion states

### LINKED ISSUE #981: Motions should only be forwarded if their state allows this

### LINKED ISSUE #982: The `created_at` timestamp is configurable by motions' states

GabrielInTheWorld:
- "Created timestamp" depends on a motion state (Fixes #982)
- A motion can only be forwarded if it is in the correct motion state (Fixes #981)

### DIFF
--- a/client/src/app/core/repositories/motions/motion-state-repository.service.ts
+++ b/client/src/app/core/repositories/motions/motion-state-repository.service.ts
@@ -35,6 +35,8 @@ export class MotionStateRepositoryService extends BaseRepositoryWithActiveMeetin
             `allow_support`,
             `allow_create_poll`,
             `allow_submitter_edit`,
+            `allow_motion_forwarding`,
+            `set_created_timestamp`,
             `set_number`,
             `show_state_extension_field`,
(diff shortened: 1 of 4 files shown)


emanuelschuetze (changes requested):
* timestamp is not recreated/updated if timestamp already exists (e.g. set two checkboxes for two states -> first time works, second state not)
* forwording state check is not working -> forwarding is possible in each state

jsangmeister:
> * timestamp is not recreated/updated if timestamp already exists (e.g. set two checkboxes for two states -> first time works, second state not)

This is intended and what we discussed - only the first state should set the timestamp, or else it becomes meaningless and more of an "updated" timestamp.

jsangmeister:
> * forwording state check is not working -> forwarding is possible in each state

This is not yet implemented in the backend and will follow with the whole rework in https://github.com/OpenSlides/openslides-backend/pull/1296. The client should, however, probably hide the button if forwarding is not possible.

emanuelschuetze (approved):

PR #1005, opened 2022-04-20
Add flow text option to PDF export

m-schieder:


client/src/app/site/motions/services/motion-pdf-catalog.service.ts
```diff
@@ -53,10 +53,20 @@ export class MotionPdfCatalogService {
         const motionDocList = [];
         const printToc = exportInfo.pdfOptions.includes(`toc`);
         const enforcePageBreaks = exportInfo.pdfOptions.includes(`addBreaks`);
+        const contentFlowText = exportInfo.content.includes(`flowText`);
```
GabrielInTheWorld, line 56: Use here the const from above, too.

client/src/app/site/motions/services/motion-pdf.service.ts
```diff
@@ -103,14 +103,18 @@ export class MotionPdfService {
...
+            const title = this.createTitle(motion, crMode, lineLength);
+            const sequential =
+                infoToExport?.includes(`id`) ?? this.meetingSettingsService.instant(`motions_show_sequential_number`);
+            const subtitle = this.createSubtitle(motion, sequential);
 
-        motionPdfContent = [title, subtitle];
+            motionPdfContent = [title, subtitle];
+        } else {
+            motionPdfContent = [];
+        }
```
GabrielInTheWorld, line 115: The `else`-block is not necessary, since the `motionPdfContent` is initialized by `[ ]`.

client/src/app/site/motions/services/motion-pdf.service.ts
```diff
@@ -71,7 +71,7 @@ export class MotionPdfService {
      *                         this selection will be ignored and all comments exported
      * @returns doc def for the motion
      */
-    public motionToDocDef(motion: ViewMotion, exportInfo?: MotionExportInfo): object {
+    public motionToDocDef(motion: ViewMotion, exportInfo?: MotionExportInfo, flowText: boolean = false): object {
```
GabrielInTheWorld, line 74: Perhaps, the name `flowText` is misleading: You use it to en-/disable the use of a title/subtitle. Therefore, name it something like `useTitle` or something else (maybe, this should also be done in the files above).

client/src/app/site/motions/services/motion-pdf-catalog.service.ts
```diff
@@ -53,10 +53,20 @@ export class MotionPdfCatalogService {
...
-                const motionDocDef: any = this.motionPdfService.motionToDocDef(motions[motionIndex], exportInfo);
+                if (motionIndex === 0) {
+                    flowText = false;
+                }
+
+                const motionDocDef: any = this.motionPdfService.motionToDocDef(
+                    motions[motionIndex],
+                    exportInfo,
+                    flowText
+                );
```
GabrielInTheWorld, line 69: The variable `flowText` does not seem to be necessary: Just pass `contentFlowText || motionIndex !== 0`.

emanuelschuetze (approved):

PR #1009, opened 2022-04-21
Pdf export revise

m-schieder:
These are still things that can be improved. These are the things with high priority for now.

There is still a bug in there. If 'Table of Contents' and 'Force Page Breaks' are active and then disabled by 'Continuous Text' the buttons remain disabled. When the export dialog is called again, these buttons can be selected again.

### DIFF
--- a/client/src/app/site/motions/modules/shared-motion/motion-export-dialog/motion-export-dialog.component.ts
+++ b/client/src/app/site/motions/modules/shared-motion/motion-export-dialog/motion-export-dialog.component.ts
@@ -66,7 +66,7 @@ export class MotionExportDialogComponent implements OnInit {
     private defaults: MotionExportInfo = {
         format: ExportFileFormat.PDF,
         content: [`text`, `reason`],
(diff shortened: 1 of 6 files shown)


client/src/app/site/motions/modules/shared-motion/motion-export-dialog/motion-export-dialog.component.ts
```diff
@@ -103,9 +103,15 @@ export class MotionExportDialogComponent implements OnInit {
...
      */
-    @ViewChild(`toc`, { static: true })
+    @ViewChild(PDF_OPTIONS.Toc)
     public tocButton: MatButtonToggle;
 
+    /**
+     * To deactivate the addBreaks button.
+     */
+    @ViewChild(PDF_OPTIONS.AddBreaks)
+    public addBreaksButton: MatButtonToggle;
```
GabrielInTheWorld, line 113: Good solution!

GabrielInTheWorld (approved):

emanuelschuetze (approved):

PR #1021, opened 2022-04-28
Add wrapper for TinyMCE editor

jsangmeister:
When visiting a page with a tinymce editor, the following error message appears:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
```
The behaviour is described and analyzed e.g. in https://github.com/tinymce/tinymce-angular/issues/245 and seems to be a reoccuring TinyMCE-with-angular problem (see also https://github.com/tinymce/tinymce-angular/issues/76 for more explanation). A described workaround is to only initialize the editor after `ngAfterViewInit` has been called, so I tried to wrap the editor in our own component, which works well except for the `fromControlName`, which still throws an error since angular is missing a `formGroup`. Maybe @GabrielInTheWorld can look into how to fix this?


client/src/app/site/motions/modules/motion-detail/components/motion-comment/motion-comment.component.html
```diff
@@ -12,13 +12,11 @@
...
         <form [formGroup]="commentForm" *ngIf="isCommentEdited">
             <!-- The HTML Editor -->
-            <editor
+            <os-editor
                 [id]="section.id"
-                formControlName="comment"
+                _formControlName="comment"
                 [init]="tinyMceSettings"
-                (onInit)="onInitTinyMce($event)"
-                (onRemove)="onLeaveTinyMce($event)"
```
jsangmeister, line 20: These seem to have been a fragment from a previous restructure and even provoke errors since the functions do not exist anymore, so I removed them.

GabrielInTheWorld:
Closed in favor of #1024.

Issue #777, opened 2022-01-25
Cannot set null values in poll


GabrielInTheWorld:
- [ ] Remove a value
- [ ] Send a not entered value

luisa-beerboom:
Have investigated the issue and found out that the values for the options in clients' update and create requests will always contain `undefined` if the field is empty upon submit.
If one creates an empty poll in the current version the backend will set all empty fields to `VOTE_UNDEFINED`, however upon edit such fields are set to 0 for the options.
This _could_ be caught in the client, but it would probably be cleaner to solve this in the backend
(i.e. have the backend handle update-requests for analog polls by setting all vote-fields, that are passed as `undefined`, to -2(=VOTE_UNDOCUMENTED))
Would probably need @jsangmeister s and @GabrielInTheWorld s opinion on that.

jsangmeister:
Sure, the backend can implement this.

luisa-beerboom:
Yes, thank you

PR #1047, opened 2022-05-02
#957 no empty poll result values

### LINKED ISSUE #957: In poll results, `0` values are always shown
They should be hidden.

luisa-beerboom:
Closes #957 
For analog polls: Hid undocumented vote results in motion poll, same for the values of votescast, votesinvalid and votesvalid in the assignment poll.
In order for this to work properly the backend needs to set the votescast, votesvalid and votesinvalid values to VOTE_UNDOCUMENTED (i.e. -2) for analog polls, if no value is sent over by the client. (This has already been discussed with @jsangmeister and @r-peschke )

Also noticed that the percentages for poll results used negative vote-values (like VOTE_UNDOCUMENTED) in their calculations (which caused frankly impossible percentages like 200% to be calculated) and fixed that as well.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/services/poll.service/poll.service.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/services/poll.service/poll.service.ts
@@ -195,13 +195,13 @@ export abstract class PollService {
                 totalByBase = this.sumOptionsYNA(option);
                 break;
             case PollPercentBase.Valid:
-                totalByBase = poll.votesvalid >0 ? poll.votesvalid : 0;
+                totalByBase = poll.votesvalid > 0 ? poll.votesvalid : 0;
                 break;
             case PollPercentBase.Entitled:
                 totalByBase = poll.entitled_users_at_stop?.length || 0;
                 break;
             case PollPercentBase.Cast:
-                totalByBase = poll.votescast >0 ? poll.votescast : 0;
+                totalByBase = poll.votescast > 0 ? poll.votescast : 0;
                 break;
             default:
                 break;
--- a/client/src/app/site/pages/meetings/modules/poll/services/poll.service/poll.service.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/services/poll.service/poll.service.ts
@@ -195,13 +195,13 @@ export abstract class PollService {
                 totalByBase = this.sumOptionsYNA(option);
                 break;
             case PollPercentBase.Valid:
-                totalByBase = poll.votesvalid;
+                totalByBase = poll.votesvalid >0 ? poll.votesvalid : 0;
                 break;
             case PollPercentBase.Entitled:
                 totalByBase = poll.entitled_users_at_stop?.length || 0;
                 break;
             case PollPercentBase.Cast:
-                totalByBase = poll.votescast;
+                totalByBase = poll.votescast >0 ? poll.votescast : 0;
                 break;
             default:
                 break;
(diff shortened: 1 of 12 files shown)


client/src/app/site/polls/services/poll.service.ts
```diff
@@ -267,11 +267,14 @@ export class PollService {
...
-        return (option?.yes ?? 0) + (option?.no ?? 0);
+        const yes = option?.yes ?? 0;
+        const no = option?.no ?? 0;
+        return (yes >= 0 ? yes : 0) + (no >= 0 ? no : 0);
     }
 
     private sumOptionsYNA(option: OptionData): number {
-        return this.sumOptionsYN(option) + (option?.abstain ?? 0);
+        const abstain = option?.abstain ?? 0;
+        return this.sumOptionsYN(option) + (abstain >= 0 ? abstain : 0);
```
GabrielInTheWorld, line 277: Why did you change it?

client/src/app/site/motions/modules/motion-poll/motion-poll-dialog/motion-poll-dialog.component.ts
```diff
@@ -5,7 +5,7 @@ import { TranslateService } from '@ngx-translate/core';
 import { PollDialogData } from 'app/core/ui-services/base-poll-dialog.service';
 import { ComponentServiceCollector } from 'app/core/ui-services/component-service-collector';
 import { BaseModel } from 'app/shared/models/base/base-model';
-import { VoteValue } from 'app/shared/models/poll/poll-constants';
+import { VOTE_UNDOCUMENTED,VoteValue } from 'app/shared/models/poll/poll-constants';
```
GabrielInTheWorld, line 8: Will this be used?

GabrielInTheWorld (changes requested):

client/src/app/site/polls/services/poll.service.ts
luisa-beerboom: Because I found out that if values were set to VOTE_UNDOCUMENTED (i.e. set to -2), they would be added onto the sum, even if they really shouldn't.
The result was that the sum would be (sum of documented votes - 2*number of undocumented options).
Since percentage values are calculated on these functions this would warp them (sometimes to the point where certain options would have values like 200%).
To counteract this I changed the function to not add on values that are smaller than 0

emanuelschuetze (changes requested):
It is not possible to remove a vote value, e.g.
set votes valid to "100" -> save -> now try to remove the value "100" (leave field empty) and save
=> "100" still visible. expected behavior:  empty value (= -2) should be saved.

luisa-beerboom:
> It is not possible to remove a vote value, e.g. set votes valid to "100" -> save -> now try to remove the value "100" (leave field empty) and save => "100" still visible. expected behavior: empty value (= -2) should be saved.

This is currently being fixed in the backend ( see comments for issue #777 )

jsangmeister:
Done in https://github.com/OpenSlides/openslides-backend/pull/1322

luisa-beerboom:
Status report: Currently works for YNA-polls, for other polls there is a bug in the backend that currently makes it impossible to update an analog poll

GabrielInTheWorld (approved):

luisa-beerboom:
> Status report: Currently works for YNA-polls, for other polls there is a bug in the backend that currently makes it impossible to update an analog poll

Working again thanks to OpenSlides/openslides-backend#1325

emanuelschuetze (approved):

PR #1050, opened 2022-05-03
#1037 assignment polls sort by vote number

luisa-beerboom:
Closes #1037 
Changed table-data-generation to treat undocumented optional values as if they were 0.

PR #1076, opened 2022-05-06
Assignment poll results now sorted correctly

### LINKED ISSUE #1037: Assignment polls: The result of a poll should be sorted by their number of votes

luisa-beerboom:
Closes #1037 
Changed table-data-generation to treat undocumented optional values as if they were 0.

### DIFF
--- a/client/src/app/infrastructure/utils/functions.ts
+++ b/client/src/app/infrastructure/utils/functions.ts
@@ -153,7 +153,7 @@ export function mmToPoints(mm: number, dense: number = 72): number {
  * @param b the number that needs to be compared with a
  * @returns 0 if they are equal, a negative value if a>b, else a positive value
  */
-export function compareValues(a: number, b: number): number {
+export function compareNumber(a: number, b: number): number {
     if (!b) {
         if (!a) {
             return 0;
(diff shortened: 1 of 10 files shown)


luisa-beerboom:
Basically same content as #1050 but requested changes were applied and it was rebased to current client

emanuelschuetze (approved):

client/src/app/infrastructure/utils/functions.ts
```diff
@@ -146,3 +146,22 @@ export function mmToPoints(mm: number, dense: number = 72): number {
...
     return inches * dense;
 }
+
+/**
+ * Compares two numbers.
+ * @param a the first number
+ * @param b the number that needs to be compared with a
+ * @returns 0 if they are equal, a negative value if a>b, else a positive value
+ */
+export function compareValues(a: number, b: number): number {
```
GabrielInTheWorld, line 156: If you only compare numbers, please name it something like `compareNumber` or add this function to the `Number` namespace (my preference). To do the second one, you should check the file `overload-js-functions.ts`.

GabrielInTheWorld (changes requested):

GabrielInTheWorld (approved):

PR #1078, opened 2022-05-06
Re-activates the history mode

### LINKED ISSUE #533: Reactivate history mode
The following PR temporarily disabled history mode/time travel service.
When the backend/system is ready, reactivate it again

#529

GabrielInTheWorld:
Fixes #533


GabrielInTheWorld:
@ostcar: The connection to the openslides-autoupdate-service does not work as expected. The client sends the request:
Url: `https://localhost:8000/system/autoupdate?position=53`

Body:
```
[
    {
        "collection": "motion",
        "ids": [
            5
        ],
        "fields": {
            "sequential_number": null,
            "title": null,
            "number": null,
            "created": null,
            "forwarded": null,
            "sort_weight": null,
            "start_line_number": null,
            "category_weight": null,
            "lead_motion_id": null,
            "amendment_ids": null,
            "submitter_ids": null,
            "reason": null,
            "recommendation_id": null,
            "tag_ids": null,
            "personal_note_ids": null,
            "block_id": null,
            "category_id": null,
            "comment_ids": null,
            "modified_final_version": null,
            "state_extension": null,
            "recommendation_extension": null,
            "list_of_speakers_id": null,
            "agenda_item_id": null,
            "amendment_paragraph_$": {
                "type": "template"
            },
            "all_origin_ids": null,
            "derived_motion_ids": null,
            "poll_ids": null,
            "origin_id": null,
            "sort_parent_id": null,
            "state_id": null,
            "workflow_id": null,
            "text": null,
            "change_recommendation_ids": null,
            "id": null
        }
    }
]
```

However, it receives nothing. What am I missing?

ostcar:
The request seems correct. I also tried it locally and it worked.

The permission checks are a bit special. Can you check the same request with the superadmin and see if it works?

The user that uses the "position" request either needs the organization management permission or he has to be in the admin group of a meeting: https://github.com/OpenSlides/openslides-backend/blob/e44c1c45ff208e9f420c6cb650606881478bcc28/global/meta/models.yml#L1576-L1580

emanuelschuetze:
I tested this branch with superadmin. It does not work: all autoupdate and au-helth requests failed...

ostcar:
Please test with the performance tool, if the direct request works. If not, please paste the request and the response.

ostcar:
When I check our your branch, then I can confirm, that the request "does not return". But it seems to be a problem from the client.

When I copy the request as cURL and send it direly, then it works (you also have to add the "-k" option to skip ssl validation).

The error message that is shown from Firefox is "NS BINDING ABORTED". In chome the error message is "(canceled)". It seems, the client is canceling the request, after it was send but before a response is returned.

ostcar:
I found the error in the client. The stream is not started. You have to set the attribute `shouldStartImmediately` in the `HttpStreamOptions` to `true`.

emanuelschuetze (approved):
ok, we can fix some history/wording issues later

GabrielInTheWorld:
> I found the error in the client. The stream is not started. You have to set the attribute `shouldStartImmediately` in the `HttpStreamOptions` to `true`.

Thanks for your help! The issue was not the property `shouldStartImmediately` in the `HttpStreamOptions`. That property did not affect the sending of the history requests. I fixed this in another way.

PR #1108, opened 2022-05-12
Updated translations

emanuelschuetze:


emanuelschuetze:
@GabrielInTheWorld can you check why the motion state.name is not translated (in motion list and motion detail view)?

emanuelschuetze:
It looks like some of your last translation fixes are not rebased correctly by your rework branch. E.g. the translation of input values in meeting config is missing again...

Issue #1112, opened 2022-05-13
Check in which component the `translate.instant` function can be replaced by the marker (also seen as `_`)


GabrielInTheWorld:
I think the function `translate.instant` (provided by the `TranslateService`) is only in services and pipes required. In components, the marker function should be used and sufficient.

jsangmeister:
Looked through the components using `translate.instant` and it seems the only uses are for titles, dialog titles and content, snackbars, errors, dynamically created strings and other places where an instant translation is necessary, so I'm closing this.

Issue #1085, opened 2022-05-09
Left-over fragments from restructure


jsangmeister:
There are many places where code is still commented out and it is not clear whether these lines are no longer needed or provide functionality which must be worked back into the rest of the code. Examples:
- https://github.com/OpenSlides/openslides-client/blob/9f3920145a19551588da6cb3d33e776c82983e07/client/src/app/site/pages/meetings/pages/mediafiles/view-models/view-mediafile.ts#L30-L43
- https://github.com/OpenSlides/openslides-client/blob/d2b6c1955c74972335039cf00511074abe94bed3/client/src/app/site/services/operator.service.ts#L640-L645
- https://github.com/OpenSlides/openslides-client/blob/9f3920145a19551588da6cb3d33e776c82983e07/client/src/app/site/base/base.component.ts#L172
- https://github.com/OpenSlides/openslides-client/blob/9f3920145a19551588da6cb3d33e776c82983e07/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-comment/motion-comment.component.ts#L109
- https://github.com/OpenSlides/openslides-client/blob/9f3920145a19551588da6cb3d33e776c82983e07/client/src/app/site/pages/organization/pages/accounts/pages/account-import/components/account-import-list/account-import-list.component.ts#L34-L35
- https://github.com/OpenSlides/openslides-client/blob/9f3920145a19551588da6cb3d33e776c82983e07/client/src/app/site/pages/organization/pages/accounts/services/common/account-controller.service.ts#L69-L95
@GabrielInTheWorld you should systematically search for these places and either create issues for them or delete them completely if they are no longer needed before you leave and no one knows where they come from or what they should do. The restructure made it pretty hard to research these things with `git blame`...

jsangmeister:
Not all places were fixed by the linked PR. Also, as stated above, I suspect there may be many more commented-out lines, since these are just examples I found while restructuring the DI.

jsangmeister:
Can do this while coming across these places in other PRs. Closing this for now

PR #1117, opened 2022-05-16
Simplify DI for base classes & add error service back

### LINKED ISSUE #1081: Action errors are no longer displayed to the user
Regression from #989: Previously, if an action request returned with an error code, the error message was displayed to the user via a snack bar. Now, if I try to change my password for example and enter the wrong old password, the request fails silently. I suspect that this affects all requests and not only the password change.

### LINKED ISSUE #1225: Enhance the DI
Many components and services are extended from a base directive. The base directive usally injects services, which have to be injected in the extended component or service. The DI can be enhanced as shown in #1081.

jsangmeister:
fixes #1225


client/src/app/site/base/base-controller.ts
```diff
@@ -1,23 +1,20 @@
...
+    public translate: TranslateService;
 
-    public constructor(
-        protected controllerServiceCollector: ControllerServiceCollectorService,
-        protected baseModelCtor: ModelConstructor<M>,
-        protected repo: BaseRepository<V, M>
-    ) {}
+    public constructor(protected baseModelCtor: ModelConstructor<M>, protected repo: BaseRepository<V, M>) {
+        const injector = AppInjector.getInjector();
+        this.translate = injector.get(TranslateService);
```
GabrielInTheWorld, line 16: Here we have to check, if this is sufficient to get the translations extracted from the TypeScript files. To check this, run `npm run extract` and check, if there are some changes. This might be related to #1112.

client/src/app/site/base/base-import.service/base-import.service.ts
```diff
@@ -217,13 +214,19 @@ export abstract class BaseImportService<MainModel extends Identifiable> implemen
...
 
     /**
      * Constructor. Creates a fileReader to subscribe to it for incoming parsed
      * strings
      */
-    public constructor(private importServiceCollector: ImportServiceCollectorService) {
+    public constructor() {
+        const injector = AppInjector.getInjector();
+        this._papa = injector.get(Papa);
+        this.translate = injector.get(TranslateService);
```
GabrielInTheWorld, line 228: This is the same as for `BaseControllerService`.

client/src/app/site/pages/meetings/modules/poll/components/poll-form/poll-form.component.ts
```diff
@@ -146,7 +146,7 @@ export class PollFormComponent extends BaseUiComponent implements OnInit {
             //         this.data.max_votes_amount = assignment.open_posts;
             //     }
             //     if (!this.data.pollmethod) {
-            //         this.data.pollmethod = this.meetingSettingService.instant(`assignment_poll_default_method`)!;
+            //         this.data.pollmethod = this.meetingSettingsService.instant(`assignment_poll_default_method`)!;
             //     }
             // }
```
GabrielInTheWorld, line 151: I know, its something from #1085, but I think, you can remove this ;-)

client/tsconfig.json
```diff
@@ -5,7 +5,7 @@
         "baseUrl": "./",
         "declaration": false,
         "downlevelIteration": true,
-        "emitDecoratorMetadata": true,
+        "emitDecoratorMetadata": false,
```
GabrielInTheWorld, line 8: Why did you change this?

client/src/app/openslides-main-module/services/app-injector.service.ts
```diff
@@ -0,0 +1,13 @@
...
+export class AppInjector {
+    private static injector: Injector;
+
+    static setInjector(injector: Injector) {
+        AppInjector.injector = injector;
+    }
+
+    static getInjector(): Injector {
+        return AppInjector.injector;
+    }
```
GabrielInTheWorld, line 12: In every file, you use the `Injector`, you write something like
```js
    const injector = AppInjector.getInjector();
    const service = injector.get(ServiceToken);
```

Could you provide some methods directly, which invokes the methods provided by the `Injector`?
Something like:

```suggestion
    public static get<Token, Service>(token: Token): Service {
        return AppInjector.injector.get(token);
    }
```

client/src/app/openslides-main-module/services/app-injector.service.ts
```diff
@@ -0,0 +1,13 @@
+import { Injector } from '@angular/core';
+
+export class AppInjector {
+    private static injector: Injector;
+
+    static setInjector(injector: Injector) {
```
GabrielInTheWorld, line 6: Please do not forget the visibility ;-)

client/src/app/openslides-main-module/services/app-load.service.ts
```diff
@@ -54,12 +55,19 @@ const appConfigs: AppConfig[] = [
...
+
     public constructor(
         private injector: Injector,
-        private modelMapper: CollectionMapperService,
-        private mainMenuService: MainMenuService,
-        private fallbackRoutesService: FallbackRoutesService
-    ) {}
+    ) {
+        // Circumventing the standard CI here to set the injector before any other services are loaded
+        AppInjector.setInjector(injector);
```
GabrielInTheWorld, line 66: Hmm, this seems a little bit hacky, doesn't it? Why don't you set the injector as in your [example](https://devblogs.microsoft.com/premier-developer/angular-how-to-simplify-components-with-typescript-inheritance/)?

```suggestion
// in main.ts
platformBrowserDynamic()
    .bootstrapModule(OpenSlidesMainModule)
    .then(moduleRef => AppInjector.setInjector(moduleRef.injector))
    .catch(err => console.error(err));
```

GabrielInTheWorld (changes requested):
Overall, it looks very cool! There are some questions or remarks to resolve.

client/src/app/openslides-main-module/services/app-load.service.ts
jsangmeister: Because this way, angular loads some components before the bootstrap promise resolves, which fails because the injector is not present yet. This was the best place I found to set the injector before anything else is loaded.

client/tsconfig.json
jsangmeister: I came across the same error as in #1174 and read that this might fix it, since setting this option to `true` is no longer needed since angular 8. But it didn't help, so I'll revert it for now.

client/src/app/openslides-main-module/services/app-injector.service.ts
jsangmeister: Good call, changed it accordingly.

client/src/app/site/base/base-controller.ts
jsangmeister: You are right, this destroys the extractability of strings. We have multiple options now as I see it:
- use another extraction tool (probably not practical)
- fork https://github.com/biesbjerg/ngx-translate-extract and implement a fix for this (maybe some more work, but probably the cleanest solution - unfortunately, the repo seems to be no longer maintained, which means on the other hand that we do not need to update the fork anytime soon)
- explicitly override the `translate` variable in all subclasses/use the traditional DI for the injection of the `TranslateService` (also some work, but very redundant one, but the solution is not as clean as the fork one)

PR #1120, opened 2022-05-16
Fix wrong sequential number in pdf export

### LINKED ISSUE #1111: Instead of the real `sequential_number` of a motion its `id` is printed to PDF
1. Create a motion in a meeting
2. Create another motion in another meeting
The second motion should have the sequential number `1`, but its id is at least `2`.
-> Go to its detail view and print a PDF from it. In the PDF you should see a `Sequential number 2`...

m-schieder:
fixes #1111


emanuelschuetze (approved):
Works as expected (after clear site data)

PR #1164, opened 2022-05-20
Re-introduce eslint rules & organize imports

### LINKED ISSUE #1090: Re-introduce eslint rules
After merging #989, the eslint rules are resetted to their default ones. We should re-enable the previous ones and apply additional rules.

jsangmeister:
fixes #1090 

I just copied the old rules. If you want to extend them somehow, we can also do that.

Because I also reintroduced the sorting plugin, a lot of files are changed. The sorting order seems to differ a little bit from the current one (higher imports before lower ones instead of purely alphabetical) - if this poses a problem, we'd have to change the sorting plugin to get more configuration options.


GabrielInTheWorld (approved):
Looks good! We can append further rules after merging this.

PR #1200, opened 2022-06-01
Implemented PermissionGuard

### LINKED ISSUE #1178: Implement the `PermissionGuard`
Currently, there is the `PermissionGuard`, which implements the interface `CanLoad` to check, if a module can be loaded by the operator. However, at the moment, it returns always `true`. We should overwrite the method `canLoad` to achieve the following things:
- Load only modules which an operator can really access
- If an operator can access only one meeting, they should be redirected in the `PermissionGuard` instead of the `AuthGuard`

luisa-beerboom:
Closes #1178

### DIFF
--- a/client/src/app/site/guards/permission.guard.ts
+++ b/client/src/app/site/guards/permission.guard.ts
@@ -1,28 +1,49 @@
 import { Injectable } from '@angular/core';
-import { CanLoad, Route } from '@angular/router';
+import { CanLoad, Route, Router, UrlSegment } from '@angular/router';
 
-import { OperatorService } from '../services/operator.service';
+import { AuthCheckService } from '../services/auth-check.service';
+import { OpenSlidesRouterService } from '../services/openslides-router.service';
+import { RerouteService } from '../services/reroute.service';
 
 @Injectable({
     providedIn: `root`
 })
 export class PermissionGuard implements CanLoad {
-    public constructor(private operator: OperatorService) {}
+    public constructor(
+        private router: Router,
+        private reroute: RerouteService,
+        private osRouter: OpenSlidesRouterService,
+        private authCheck: AuthCheckService
+    ) {}
 
-    public async canLoad(route: Route): Promise<boolean> {
+    public async canLoad(route: Route, segments: UrlSegment[]): Promise<boolean> {
+        if (this.osRouter.isOrganizationUrl(this.getCurrentNavigationUrl())) {
+            if (!(await this.authCheck.isAuthorizedToSeeOrganization())) {
+                this.reroute.forwardToOnlyMeeting(this.getCurrentNavigationUrl() === `/info` ? [`info`] : []);
+                return false;
+            }
+        } else if (!(await this.authCheck.isInMeeting(this.getCurrentNavigationUrl()))) {
+            this.reroute.handleForbiddenRoute(route.data, segments);
+        }
+        if (route.data) {
+            if (!(await this.authCheck.isAuthenticated())) {
+                this.reroute.toLogin();
+                return false;
+            } else if (!(await this.authCheck.isAuthorized(route.data))) {
+                this.reroute.handleForbiddenRoute(route.data, segments);
+                return false;
+            }
+        }
+        this.authCheck.lastSuccessfulUrl = this.getCurrentNavigationUrl();
         return true;
-        // await this.operator.ready;
-        // const meetingPermissions: Permission[] = route.data?.['meetingPermissions'];
-        // const omlPermissions: OML[] = route.data?.['omlPermissions'];
-        // console.log(`canLoad`, meetingPermissions, omlPermissions);
-        // if (meetingPermissions?.length) {
-        //     console.log(`operator canload:`, this.operator.hasPerms(...meetingPermissions));
-        //     return this.operator.hasPerms(...meetingPermissions);
-        // }
-        // if (omlPermissions?.length) {
-        //     console.log(`operator canLoad:`, this.operator.hasOrganizationPermissions(...omlPermissions));
-        //     return this.operator.hasOrganizationPermissions(...omlPermissions);
-        // }
-        // return true;
+    }
+
+    /**
+     * Workaround for getting the current url because canLoad's 'segments' parameter currently does not supply the whole route.
+     * Can be replaced as soon as segments is guaranteed to contain the whole url
+     * @returns the current navigation's url
+     */
+    public getCurrentNavigationUrl(): string {
+        return this.router.getCurrentNavigation().extractedUrl.toString();
     }
 }
--- a/client/src/app/site/guards/permission.guard.ts
+++ b/client/src/app/site/guards/permission.guard.ts
@@ -22,7 +22,7 @@ export class PermissionGuard implements CanLoad {
                 this.reroute.forwardToOnlyMeeting(this.getCurrentNavigationUrl() === `/info` ? [`info`] : []);
                 return false;
             }
-        } else if (!(await this.authCheck.isInMeeting(this.getCurrentNavigationUrl()))) {
+        } else if (!(await this.authCheck.hasAccessToMeeting(this.getCurrentNavigationUrl()))) {
             this.reroute.handleForbiddenRoute(route.data, segments);
         }
         if (route.data) {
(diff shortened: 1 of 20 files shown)


GabrielInTheWorld (changes requested):

client/src/app/site/guards/permission.guard.ts
```diff
@@ -1,28 +1,42 @@
...
+        if (segments.length === 0) {
+            await this.operator.ready;
+            try {
+                return !this.operator.knowsMultipleMeetings;
+            } catch (e) {
+                return true;
+            }
+        }
+        return false;
     }
```
GabrielInTheWorld, line 41: Can the implementation of `isOnForbiddenRoute` in the `PermissionGuard` and the implementation of `hasStateMeaning` in the `AuthGuard` be unified?

client/src/app/site/services/auth-check.service.ts
```diff
@@ -0,0 +1,120 @@
...
+     * If the user requested a route without direct meaning, forward to their meaningful meeting
+     */
+    public forwardToOnlyMeeting(): void {
+        try {
+            const meetingId = this.operator.onlyMeeting;
+            this.router.navigate([meetingId]);
+        } catch (e) {
+            throw new Error(`Error when trying to forward to only meeting: ${e?.message}`);
+        }
+    }
```
GabrielInTheWorld, line 119: I would expect, that the `Auth**Check**Service` checks only if a user is, for example, authorized or authenticated (or both). These functions should be implemented at another place.

GabrielInTheWorld, line 119: I would expect, that the `Auth**Check**Service` checks only if a user is, for example, authorized or authenticated (or both). These functions should be implemented at another place.

GabrielInTheWorld (changes requested):

client/src/app/site/services/reroute.service.ts
```diff
@@ -0,0 +1,79 @@
...
+                }
+            });
+        } else {
+            this.router.navigate([`error`], {
+                queryParams: {
+                    error: `Authentication Error`,
+                    msg: routeDataArray
+                }
+            });
+        }
```
GabrielInTheWorld, line 61: Please unify these code blocks.

client/src/app/site/services/auth-check.service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { ActiveMeetingService } from '../pages/meetings/services/active-meeting.service';
+import { MeetingSettingsService } from '../pages/meetings/services/meeting-settings.service';
+import { OperatorService } from './operator.service';
+import { RerouteService } from './reroute.service';
+
+export enum AuthorizationStatus {
+    Authorized,
+    NoPerms,
+    NotLoggedIn
```
GabrielInTheWorld, line 15: Please use an own enumeration `AuthenticationStatus` (this could be in another service, like the `AuthService`).
I like opposites more: `Authenticated` and `Unauthenticated`.

client/src/app/site/services/reroute.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { FallbackRoutesService } from './fallback-routes.service';
+import { OpenSlidesRouterService } from './openslides-router.service';
+import { OperatorService } from './operator.service';
+
+@Injectable({
+    providedIn: `root`
+})
+export class RerouteService {
+    constructor(
```
GabrielInTheWorld, line 13: Don't forget the visibility modifier.

client/src/app/site/services/auth-check.service.ts
```diff
@@ -0,0 +1,75 @@
...
+})
+export class AuthCheckService {
+    public constructor(
+        private reroute: RerouteService,
+        private operator: OperatorService,
+        private activeMeeting: ActiveMeetingService,
+        private meetingSettingService: MeetingSettingsService
+    ) {}
+
+    public async getAuthStatus(routeData: Data): Promise<AuthorizationStatus> {
```
GabrielInTheWorld, line 29:
```suggestion
    public async getAuthorizationStatus(routeData: Data): Promise<AuthorizationStatus> {
```

...to distinguish from an authentication status.

GabrielInTheWorld (changes requested):
We should think about an `AuthorizationService` and should rename the `AuthService` to `AuthenticationService`.

client/src/app/site/guards/permission.guard.ts
```diff
@@ -1,28 +1,42 @@
...
-        // }
-        // if (omlPermissions?.length) {
-        //     console.log(`operator canLoad:`, this.operator.hasOrganizationPermissions(...omlPermissions));
-        //     return this.operator.hasOrganizationPermissions(...omlPermissions);
-        // }
-        // return true;
+    }
+
+    /**
+     * Workaround for getting the current url bc canLoad's 'segments' parameter currently does not supply the whole route.
```
GabrielInTheWorld, line 35:
```suggestion
     * Workaround for getting the current url because canLoad's 'segments' parameter currently does not supply the whole route.
```

client/src/app/site/pages/error/components/error-main/error-main.component.ts
```diff
@@ -1,20 +1,37 @@
...
+    }
+
+    private getReturnUrl(): string | number {
+        if (!this.authService.isAuthenticated()) {
+            return `login`;
+        } else if (this._meetingId) {
+            return this._meetingId;
         }
+        return ``;
+    }
```
GabrielInTheWorld, line 32: Please order private functions under public and protected functions

client/src/app/site/services/auth-check.service.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (typeof info === `string`) {
+            const segments = info.split(`/`);
+            meetingIdString = segments.length > 1 ? segments[1] : segments[0];
+        }
+        if (Number.isNaN(Number(meetingIdString))) {
+            return false;
+        }
+        await this.operator.ready;
+        return this.operator.isInMeeting(Number(meetingIdString));
+    }
```
GabrielInTheWorld, line 92: I think, these functions are better placed in the `OpenSlidesRouterService`.

client/src/app/site/services/auth-check.service.ts
```diff
@@ -0,0 +1,93 @@
...
+        const toCheck = Array.isArray(omlPerm) ? omlPerm : [omlPerm];
+        return result && this.operator.hasOrganizationPermissions(...toCheck);
+    }
+
+    private isMeetingSettingEnabled(meetingSetting?: keyof Settings): boolean {
+        if (!meetingSetting) {
+            return true;
+        }
+        return this.meetingSettingService.instant(meetingSetting) as boolean;
+    }
```
GabrielInTheWorld, line 57: Please order private functions under public and protected ones.

GabrielInTheWorld (changes requested):
Some last changes

emanuelschuetze (changes requested):
Cannot navigate as superadmin into a meeting:
Get empty page with "Authorization Error" and "Back to OpenSlides" button.

client/src/app/site/services/openslides-router.service.ts
```diff
@@ -67,6 +67,25 @@ export class OpenSlidesRouterService {
         }
     }
 
+    /**
+     * Returns true if an url is not meeting-specific
+     */
+    public isOrganizationUrl(url: string): boolean {
+        const urlSegments = url.slice(1).split(`/`);
+        return !urlSegments[0] || Number.isNaN(Number(urlSegments[0]));
+    }
```
jsangmeister, line 76: Please unify with the `auth-check.service`

jsangmeister (changes requested):
looks good!

jsangmeister (approved):
not tested

emanuelschuetze (changes requested):
Routing does not work correctly if 'can_see_frontpage' is disabled:
- create two meetings (1 and 2)
- create new delegate user
- remove permission 'can_see_frontpage' for meeting 2
- as delegate user: go from calender to meeting 2 -> Authorization Error ("Required permissions to view this page: meeting.can_see_autopilot") => wront
- try to open https://localhost:8000/2/ -> redirects to meeting _1_ (https://localhost:8000/1/autopilot)

emanuelschuetze (approved):
works good!

PR #1210, opened 2022-06-03
Changed how vote-permission-errors are determined

### LINKED ISSUE #1206: Vote delegation problem

luisa-beerboom:
in order to fix vote delegation bug.
Closes #1206

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/services/voting.service/voting.service.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/services/voting.service/voting.service.ts
@@ -5,6 +5,7 @@ import { ViewPoll } from 'src/app/site/pages/meetings/pages/polls';
 import { ViewUser } from 'src/app/site/pages/meetings/view-models/view-user';
 import { OperatorService } from 'src/app/site/services/operator.service';
 
+import { ActiveMeetingService } from '../../../../services/active-meeting.service';
 import { PollServiceModule } from '../poll-service.module';
 
 export enum VotingError {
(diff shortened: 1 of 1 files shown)


emanuelschuetze (changes requested):
Some tests with the usecase user A delegates to B. The main issue works now. But some minor subissues:
* If B is NOT present the message "You do not have the permission to vote." is irritating. "You have to be present to vote." would be helpful to understand that vote for B requires present state.
* start voting (B can vote for A works) -> now remove for A the vote delegation as admin => B still see the voting buttons (expected: should be removed)

luisa-beerboom:
> Some tests with the usecase user A delegates to B. The main issue works now. But some minor subissues:
> 
>     * If B is NOT present the message "You do not have the permission to vote." is irritating. "You have to be present to vote." would be helpful to understand that vote for B requires present state.
> 
>     * start voting (B can vote for A works) -> now remove for A the vote delegation as admin => B still see the voting buttons (expected: should be removed)

fixed the first point,
relocated the second point into new issue #1231

emanuelschuetze (approved):

client/src/app/site/pages/meetings/modules/poll/base/base-poll-vote.component.ts
```diff
@@ -97,6 +97,11 @@ export abstract class BasePollVoteComponent<C extends BaseViewModel = any> exten
         return this.votingService.getVotePermissionErrorVerbose(this.poll, user) || ``;
     }
 
+    public getVotingErrorFromName(errorName: string) {
+        console.info(`Cannot vote because:`, this.votingService.getVotePermissionErrorVerboseFromName(errorName));
```
emanuelschuetze, line 101: this looks like for debuggin only?

PR #1226, opened 2022-06-08
Made calendar filter for enterable meetings

### LINKED ISSUE #1201: A template meeting is seen by every committee manager in the dashboard
Those should only be visible when a committee manager creates a new meeting.
A committee detail view has a check to let a user only enter a meeting, if they have necessary permissions for that.
We should use the same check in the dashboard to filter every meeting, which a user cannot enter.

luisa-beerboom:
also fixed template selector sorting

Closes #1201


emanuelschuetze (changes requested):
mark a meeting as template meeting.
open new meeting form.
=> duplicate from field shows meeting in unorderd/ungrouped form...

emanuelschuetze (approved):

PR #1227, opened 2022-06-08
Refactors list-component and projectable-list-component

### LINKED ISSUE #1217: Cannot use the action `selectall` for multiselection in the `ProjectableListComponent`

### LINKED ISSUE #1221: search filter in motion/amendment list view does not work
open motion list
enter e.g. "Max" (to find all motoins with submitter "Max Mustermann")
=> nothing found => 0 of x motions

GabrielInTheWorld:
Introduces a base component for list components and a presentation component for list components.
- Also injects a scrolling-table-component by its dedicated service (Fixes #1217)
- Also saves a search query (Fixes #1221)


emanuelschuetze (changes requested):
* search filter string (in user list) has to be lowercase letters ("Administrator" does not match; same for "Niedersachsen" in structure level)
* search filter (in motion list) for motion identifier does not work for strings which contains letters (e.g. "L01" does not work; but "01" works)

emanuelschuetze (approved):

PR #1292, opened 2022-06-17
Implemented autoupdate decompression in the client

luisa-beerboom:
The autoupdate service recently got a feature where it could compress the data it sends out (see OpenSlides/openslides-autoupdate-service#507 ), this pull request updates the client to utilize that feature.

### DIFF
--- a/client/src/app/gateways/http-stream/http-stream.ts
+++ b/client/src/app/gateways/http-stream/http-stream.ts
@@ -1,4 +1,5 @@
 import { HttpDownloadProgressEvent, HttpErrorResponse, HttpEvent, HttpHeaderResponse } from '@angular/common/http';
+import * as fzstd from 'fzstd';
 import { firstValueFrom, Observable, Subscription } from 'rxjs';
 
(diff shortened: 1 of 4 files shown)


GabrielInTheWorld (approved):
One remark, rest looks good.

client/src/app/gateways/http-stream/http-stream.ts
```diff
@@ -228,7 +246,7 @@ export class HttpStream<T> {
         this._shouldReconnectOnFailure = config.shouldReconnectOnFailure ?? false;
         this._reconnectTimeout = config.reconnectTimeout ?? 2000;
 
-        this.buildMessageParser();
+        this.buildMessageParser(this.endpoint.url);
```
GabrielInTheWorld, line 249: Why is this necessary although an endpoint can be accessed as member of an http stream?

GabrielInTheWorld (changes requested):

GabrielInTheWorld (approved):

PR #1343, opened 2022-06-29
Reworked motion edit mediafile dialog

### LINKED ISSUE #1336: Problems with filesystem-related dialogs
- [x] #1337
- [x] Motion edit: Directory list only has base folder (should have all folders)
   - [x] Folders are shown
   - [x] File is moved into selected folder
- [ ] #1342
- [x] Motion edit: Access groups should be selectable

luisa-beerboom:
Closes #1336

### DIFF
--- a/client/src/app/ui/modules/media-upload-content/components/media-upload-content/media-upload-content.component.ts
+++ b/client/src/app/ui/modules/media-upload-content/components/media-upload-content/media-upload-content.component.ts
@@ -1,10 +1,11 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { UntypedFormBuilder, UntypedFormGroup } from '@angular/forms';
 import { TranslateService } from '@ngx-translate/core';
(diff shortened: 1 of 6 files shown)


client/src/app/ui/modules/media-upload-content/components/media-upload-content/media-upload-content.component.ts
```diff
@@ -82,7 +87,7 @@ export class MediaUploadContentComponent extends BaseUiComponent implements OnIn
     }
 
     public getDirectoryTitle(): string {
-        return this.currentDirectory?.title || ``;
+        return this.repo.getViewModel(this.directoryId)?.title || this.currentDirectory?.title || ``;
```
jsangmeister, line 90: Can this component be supplied with an id or a view model or are both needed? If the former, maybe it can be unified. If the latter, the id should be removed, since the view model also contains it.

jsangmeister (changes requested):

jsangmeister (approved):

emanuelschuetze (approved):

PR #1357, opened 2022-07-07
Fix mediafile upload form

### LINKED ISSUE #1342: Motion edit: If selected directory is the base folder the placeholder will behave as if the field is empty
i.e. in the file-upload-dialog: The name of the current value ("Base folder") will be shown in the selector's output-field, with the placeholder ("Parent directory") right behind it, when the field is not focused. This makes it look as if there are two names layered on top of each other.
This also happens in other places (f.e. in the move file dialog of the file-list)

This probably has something to do with the fact, that the selector is holding the base-folder not as a normal object, but as a noneItem (which has the value 0).
Perhaps change the way the selector handles noneItems, or make the affected views handle the selection of the base folder without using the noneItem-functionality.

jsangmeister:
fixes #1342 

also fixes another bug where the page would just redirect to the previous location after an upload. This would lead to unwanted behaviour if one was e.g. directed by a direct link to the upload page.

### DIFF
--- a/client/src/app/ui/modules/media-upload-content/components/media-upload-content/media-upload-content.component.html
+++ b/client/src/app/ui/modules/media-upload-content/components/media-upload-content/media-upload-content.component.html
@@ -2,7 +2,7 @@
     [addFileFn]="getAddFileFn()"
     [uploadFileFn]="getUploadFileFn()"
     [parallel]="parallel"
-    (uploadSucceeded)="uploadSuccessEvent.emit($event)"
+    (uploadSucceeded)="uploadSucceeded($event)"
     (uploadFailured)="errorEvent.emit($event)"
 >
     <ng-template #additionalContent>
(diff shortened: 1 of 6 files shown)


client/src/app/ui/modules/media-upload-content/components/media-upload-content/media-upload-content.component.html
```diff
@@ -19,13 +19,6 @@
                 ></os-list-search-selector>
             </mat-form-field>
         </div>
-
-        <div>
-            <span>{{ 'Upload to:' | translate }}</span>
-            &nbsp;
-            <i *ngIf="!directoryId">{{ 'Base folder' | translate }}</i>
-            <i *ngIf="directoryId">{{ getDirectoryTitle() }}</i>
-        </div>
```
jsangmeister, line 28: Also did some cleanup: This div (incorrectly) mirrored the selection directly above it, so I removed it and everything which was only there for it (e.g. `currentDirectory`).

luisa-beerboom (approved):
Looks good to me so far, does this work?

emanuelschuetze (approved):

PR #1361, opened 2022-07-11
Fix logout on auth error

jsangmeister:
Minor fix which forced retries on errors because the wrong property was accessed. Test in conjunction with https://github.com/OpenSlides/openslides-auth-service/pull/130 for best results.

### DIFF
--- a/client/src/app/gateways/http-stream/http-stream.service.ts
+++ b/client/src/app/gateways/http-stream/http-stream.service.ts
@@ -93,7 +93,7 @@ export class HttpStreamService {
     }
 
     private shouldReconnect({ error }: ShouldReconnectContext): boolean | Promise<boolean> {
-        if (error.error.type === PossibleStreamError.AUTH) {
+        if (error.error?.type === PossibleStreamError.AUTH) {
             this.authService.logout();
             return false;
         }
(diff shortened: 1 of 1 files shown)


client/src/app/gateways/http-stream/http-stream.service.ts
```diff
@@ -93,7 +93,7 @@ export class HttpStreamService {
     }
 
     private shouldReconnect({ error }: ShouldReconnectContext): boolean | Promise<boolean> {
-        if (error.type === PossibleStreamError.AUTH) {
+        if (error.error.type === PossibleStreamError.AUTH) {
```
luisa-beerboom, line 96: See http-stream.ts `parseCommunicationError` (around ln. 396): I believe the return of this function's result is fed into the `shouldReconnect` function as a parameter. The return has the type `CommunicationError` ( `{type: string, msg: text}` ).
I ofc haven't tested this, but I have a suspicion that your solution in connection with this may cause a TypeError, since by calling error.error you'd be receiving a null value as output.

maybe do `error.error?.type` or smth

luisa-beerboom (changes requested):
See http-stream.ts ln

client/src/app/gateways/http-stream/http-stream.service.ts
jsangmeister: Good catch! I even intended to do that but seems like I forgot

luisa-beerboom (approved):
Looks good to me

emanuelschuetze (approved):

PR #1364, opened 2022-07-11
Filtered empty users out of account list

### LINKED ISSUE #1344: Relics in account/user list after deletion of a user
After you delete a user in the account list, the user is removed, but an empty row is appended at the end of the list with only the text "Inactive" on the right side.

luisa-beerboom:
Closes #1344 

Participant list should automatically update upon any changes in the meetings participants.
Account list should update upon any changes in the **_known_** users. As soon as the account list view is first loaded, the creation of any users will not be registered, unless any of the other user subscriptions register a new user or the page is reloaded. This will not be fixed in this PR, as it is currently impossible to fix this without sending a constant stream of presenter requests to the backend.


jsangmeister (changes requested):
This looks like fighting symptoms - the "invalid" user entry is still somewhere in the data, but is just not shown. I think it would be better to fix the problem at the source and remove the user from the data completely (although I'm not sure where to do this).

emanuelschuetze:
Same problem for user list view

luisa-beerboom:
Currently got this to the point, where the participant list works again, but the account list is still broken.

luisa-beerboom:
This should fix all the updating problems that came with user deletion

client/src/app/site/services/view-model-store-update.service.ts
```diff
@@ -58,24 +58,38 @@ export class ViewModelStoreUpdateService {
...
         const update: BaseModel<any>[] = [];
+        const toDelete: number[] = [];
         for (const id of Object.keys(modelData[collection])) {
             const model = modelData[collection][+id];
-            // Important: our model system needs to have an id in the model, even if it is partial
-            model[`id`] = +id;
-            const basemodel = this.mapObjectToBaseModel(collection, model);
-            if (basemodel) {
-                update.push(basemodel);
+            if (this.isEmpty(model)) {
```
jsangmeister, line 77: Better check only for the id. If the id is null, then the model was deleted, if not then not. All other cases cannot happen and are prevented by the autoupdate service.

client/src/app/site/pages/organization/pages/accounts/components/account-main/account-main.component.ts
```diff
@@ -30,16 +31,29 @@ export class AccountMainComponent extends BaseModelRequestHandlerComponent {
...
     }
 
     protected override async onBeforeModelRequests(): Promise<void> {
         this._accountIds = await this.controller.fetchAccountIds();
+        this.subscriptions.push(
+            this.userController.getViewModelListObservable().subscribe(async users => {
+                const userIds = users.map(user => user.id);
+                if (
+                    userIds.length !== this._accountIds.length ||
+                    !(userIds.sort().join(`,`) === this._accountIds.sort().join(`,`))
```
jsangmeister, line 47: `userIds.every((userId, idx) => userId === this._accountIds)` seems to be better because it breaks as soon as a difference is found (and does not unnecessarily convert it to strings). Also, saving `this._accountIds` in its sorted order saves time because the array does not have to be sorted newly on every update. Since the users per meeting can potentially be very large, it may be worth to save as much runtime as we can.

jsangmeister (changes requested):

client/src/app/site/pages/organization/pages/accounts/components/account-main/account-main.component.ts
luisa-beerboom: Meinst du `userIds.every((userId, idx) => userId === this._accountIds[idx])`?

jsangmeister: yes, my bad

jsangmeister (changes requested):
remark about checking for the id is still open

MSoeb:
Fix don't work. Needs improvement. 

After deleting a user in the accounts menu, the user is still visible in the list. Also the user details could be opened by clicking on it. But if I try to delete it again, an error message appears. Fehler: Datastore Error. - Now we have Schrödinger's User ;)

luisa-beerboom:
I think it should all work now

client/src/app/site/pages/organization/pages/accounts/components/account-main/account-main.component.ts
```diff
@@ -61,6 +64,14 @@ export class AccountMainComponent extends BaseModelRequestHandlerComponent {
         );
     }
 
+    private deleteOldModels(): void {
+        const oldModelIds = this.userController
+            .getViewModelList()
+            .map(user => user.id)
+            .filter(id => !this.accountIds.includes(id));
+        this.userRepo.deleteModels(oldModelIds);
```
jsangmeister, line 72: The component should not be responsible for the deletion of models in the repo. Why does the approach not work without this?

jsangmeister (changes requested):

client/src/app/site/pages/organization/pages/accounts/components/account-main/account-main.component.ts
luisa-beerboom: It generally works without this, but there was one problem (which this was intended to solve):

The participant list subscribes to the participants via their parent meeting.
If I delete a participant through said list, it therefore gets reported that the participant is not in the meeting anymore, but not that he's been deleted.
If the account list (which subscribes by id) is already loaded this isn't a problem, if it isn't the user stays as a relic in the local storage and will be shown in the account list as soon as it is loaded.

This was meant to catch that by checking for the existence of any relic users before initializing the account subscription and internally deleting the models.
I agree that doing this in this place is probably wrong though, I'll try to divert this functionality to the repo and access it in a safer manner.

jsangmeister (approved):
looks better!

MSoeb:
Needs improvement. 

Good: At meeting level, everything works fine.

On organization level in the Accounts menu there are several errors: 

- First, users deleted via the three-dot menu behind their name or via multiple selection are still visible. Only by reloading the account is no longer visible and clickable. 

- Secondly, opening the account list leads to a heavy CPU load if the console is active and the console displays permanent autoupdates or, after deleting a user, it displays the error message "ERROR Error: No active meeting to query the structured relation! After a short time, the console closes itself.

luisa-beerboom:
@MSoeb 
I can't replicate any of your problems in my branch. 
Are you doing something special? Like visiting some other page in between, or coming from another place?
Is there something special about the instance? Like does it have a large amount of users or smth?

I won't be able to tell what the problem is, unless I know exactly what you're doing

luisa-beerboom:
@MSoeb I rewrote it a bit, hopefully it's better now

MSoeb:
needs improvement. 

On account menu deleted users still exists until reload. Additionally the console is spammed with autoupdates messages when the account menu is open.

luisa-beerboom:
@MSoeb , @m-schieder Please try this new version. I can't guarantee anything, but I think it should be better now.

MSoeb:
got an 503 if i try to open the instance. could you please check this, @m-schieder ?

MSoeb:
need improvement. fix works partially. 
If deleting users via multiple selection is done more than once, error messages are displayed. The error occurs in Account and Participant menu. Deletion is then only possible if the account menu/participant menu is closed and reopened. 

**Reproduction:**
1. open _Account_ menu
2. click on three-dot-menu, choose _Multiselection_
3. choose at least 1 one user
4. click again on the three-dot-menu and choose _Delete_. Confirm deletion
5. First wave of deletion works fine
6. choose again at least 1 user, open three-dot-menu and choose Delete. Confirm deletion -> error message appears in console and at the bottom of the page.

Imporant: Don't leave the Account/participant menu during the steps. 

Screenshot_20220826_093837

luisa-beerboom:
@rrenkert @emanuelschuetze 
In my last commits I changed the `scrolling-table.component.ts` to fix the deletion bugs mentioned by @MSoeb  (and then also the `account-main.component.ts` as part of a rebase (there were some conflicts concerning the recent hotfix where I increased the amount of users that would be loaded))

So: Since I made changes in the scrolling-table.component it would probably be a good idea to see, whether the other places where it is used still work (the file-list, import-lists and the other view-lists)

rrenkert (approved):
Looks good and I cant find side-effects on other list-views (tested with tags and file-list).

emanuelschuetze (approved):

PR #1423, opened 2022-07-20
Implemented client-side action worker logic

### LINKED ISSUE #1414: Implement handling for 202-timeouts with thread watcher
See OpenSlides/openslides-backend#1355

luisa-beerboom:
Closes #1414 

TODO: 
- [x] thrown errors still incomplete
- [x] there's a circular dependency error
- [x] Make it automatically open a dialog with options "stop listening" and "continue waiting" or smth.
  - [x] If the worker couldn't be written
  - [x] If the worker is slow/might be dead (worker.timestamp too long ago -> worker.hasPassedInactivityThreshold)
  - [x] Name the reason why the Dialog showed up 
  - [x] Make this dialog appear by default but reserve the option to turn it off (and handle this kind of thing manually in the component that sent the action)
- [x] test if this works with OpenSlides/openslides-backend#1137 branch and debug

### DIFF
--- a/client/src/app/gateways/action-worker-watch/action-worker-watch.service.ts
+++ b/client/src/app/gateways/action-worker-watch/action-worker-watch.service.ts
@@ -4,6 +4,7 @@ import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 import { BehaviorSubject, filter, firstValueFrom } from 'rxjs';
 import { Id, Ids } from 'src/app/domain/definitions/key-types';
 import { ActionWorkerState } from 'src/app/domain/models/action-worker/action-worker';
+import { idFromFqid } from 'src/app/infrastructure/utils/transform-functions';
 import { WaitForActionReason, waitForActionReason } from 'src/app/site/modules/wait-for-action-dialog/definitions';
 import { WaitForActionDialogService } from 'src/app/site/modules/wait-for-action-dialog/services/wait-for-action-dialog.service';
 import { ModelRequestService } from 'src/app/site/services/model-request.service';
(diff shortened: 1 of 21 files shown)


luisa-beerboom:
Needs current autoupdate-service, backend PR#1355 and datastore PR#207 to work

Have tested it a bit, there are some things where I have no idea how to, but here's some points:
- By setting `THREAD_WATCH_TIMEOUT` to an insanely low value in `action_worker.py` one can force the 202 responses to always be sent by the backend (which will call the new code).
- By setting the check-value of `isSlow()` from 2 to an equally insanely low value in `view-action-worker.ts`, one may be able to catch a glimpse of the wait-dialog (but unless you somehow manage to send a request that takes really long, it probably won't stay long enough for the content to load, considering that it closes, when the action succeeds (I think I have seen the content once though))

luisa-beerboom:
Rebased to current main

emanuelschuetze (changes requested):
- (wrong pull request, needs retest)

client/src/app/gateways/action-worker-watch/action-worker-watch.service.ts
```diff
@@ -0,0 +1,228 @@
...
+        private modelRequestService: ModelRequestService,
+        private dialogService: WaitForActionDialogService
+    ) {
+        this.actionWorkerRepo.getViewModelListObservable().subscribe(workers => this._workerSubject.next(workers));
+    }
+
+    public async watch<T>(originalResponse: HttpResponse<T>, watchActivity: boolean): Promise<HttpResponse<T>> {
+        const actionName = originalResponse.body[`results`][0][0][`name`];
+        const fqid: string = originalResponse.body[`results`][0][0][`fqid`];
+        const id = Number(fqid.split(`/`).slice(-1)[0]);
```
bastianjoel, line 56: You should use idFromFqid here.

client/src/app/gateways/action-worker-watch/action-worker-watch.service.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * Removes a worker from the currentWorkerIds.
+     * May be called from outside this service, to force the client to stop watching a running worker.
+     * If this worker is currently watched, the running watch command will throw an error "Client has stopped watching worker: <Action name>".
+     */
+    public async unsubscribeFromWorker(workerId: Id): Promise<void> {
+        this._currentWorkerIds = this._currentWorkerIds.filter(id => id !== workerId);
+        await this.refreshAutoupdateSubscription([workerId]);
```
bastianjoel, line 129: It might be a better idea to not update the au subscription on unsubscribe but only mark it for deletion, because it reopens the connection. For example in case multiple workers get resolved quickly one after another this would stress the au service unnecessarily. 

The subscriptions marked for deletion could be processed when a new worker is subscribed or all workers got resolved.

bastianjoel (approved):

r-peschke (commented):
What do you think about aborting with a message, if there passes more than one minute without a new timestamp from backend? I'll write a memory checker, because otherwise the action_worker keeps on state "running" forever.
The rest seems okay for me:-)

bastianjoel (approved):

r-peschke (changes requested):
The last timestamp seems to be not set:
Last confirmed activity on action: 20.1.1970, 08:21:37

Would be nice to fix and thereafter merge it:-)

bastianjoel (approved):

luisa-beerboom:
> The last timestamp seems to be not set: Last confirmed activity on action: 20.1.1970, 08:21:37
> 
> Would be nice to fix and thereafter merge it:-)

Should be fine now

r-peschke (approved):

r-peschke:
There are failing tests
If this is fixed you can merge it. The backend part is merged since some days.

PR #1429, opened 2022-07-21
Update history to new format

jsangmeister:


emanuelschuetze (changes requested):
the {arg1} placeholder is not replaced. Example string after state changes:
"Status gesetzt auf {arg1}"

MSoeb:
Fix does not work. After opening a request in the history, errors are permanently displayed in the console, CPU load increases and details about the opened motion are not displayed in the client.

jsangmeister:
@MSoeb did you test in conjunction with https://github.com/OpenSlides/openslides-backend/pull/1356? Also, you have to clear all existing informations before trying it out, which I explained in the backend PR. Executing `UPDATE positions SET information=NULL` in the postgres DB suffices - maybe @peb-adr can assist you in this as well.

MSoeb:
> @MSoeb did you test in conjunction with [OpenSlides/openslides-backend#1356](https://github.com/OpenSlides/openslides-backend/pull/1356)? Also, you have to clear all existing informations before trying it out, which I explained in the backend PR. Executing `UPDATE positions SET information=NULL` in the postgres DB suffices - maybe @peb-adr can assist you in this as well.

No, I do not assume that I have tested this in the configuration. I did not know that. I discuss everything else with @m-schieder or @peb-adr

MSoeb:
needs improvement. errors still appear. talked with @jsangmeister about it.

MSoeb:
I had been talking to Magnus about this PR. He mentioned that with this PR in combination with https://github.com/OpenSlides/openslides-backend/pull/1356 the existing chronicle will no longer be available. This is problematic because the chronicle is already in use and existing data will be lost. We should urgently avoid this. The optimal solution would be one that preserves the existing data. We need to discuss the implementation before publishing. @emanuelschuetze @jsangmeister @r-peschke

emanuelschuetze (changes requested):
Open the history of an old motion (created on main branch with some changes) you get a type error - the history table is empty:
```
TypeError: position.information is not iterable
    at HistoryListComponent.parseInformation (history-list.component.ts:212:43)
```

jsangmeister:
> Open the history of an old motion (created on main branch with some changes) you get a type error - the history table is empty:
> 
> ```
> TypeError: position.information is not iterable
>     at HistoryListComponent.parseInformation (history-list.component.ts:212:43)
> ```

Did you clear the position information as described above? Also, did you test in conjunction with the backend branch? When I try to reproduce it including the clearing (create motion and change status on current main - checkout this branch and the backend history branch - clear position information - go into history of motion), I don't get any error. When I update the state again, the entry is correctly appended to the history.

emanuelschuetze (approved):
Tested (with `UPDATE positions SET information=NULL` in postgres)

PR #1435, opened 2022-07-21
Made BaseFilterListService parse loaded filters into new ones

### LINKED ISSUE #1348: BaseFilterListService loading old definitions from storage causing slight problems when developing
In dev environment:
1. Open any list that has filters: (filter definitions will be stored in local storage)
2. Change the settings for that lists filters in the code of the related filter service
3. After the code has re-compiled you will find, that the lists filters do not apply your changes and are still working just as before (because the filter definitions from before were found in the storage and have been used instead of the new ones)

I spent about 3 hours on a problem that should have taken me 30 min to solve, just because I didn't realize this.

luisa-beerboom:
Closes #1348

### DIFF
--- a/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
+++ b/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
@@ -389,19 +389,19 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements 
     private parseFilters(oldFilters: OsFilter<V>[]): OsFilter<V>[] {
         const newFilterDefs = this.getFilterDefinitions();
         newFilterDefs.forEach(definition => {
-            const oldDefinition = oldFilters.filter(old => old.property === definition.property);
-            if (!oldDefinition.length) {
+            const oldDefinition = oldFilters.find(old => old.property === definition.property);
+            if (!oldDefinition) {
                 return;
             }
-            oldDefinition[0].options.forEach(option => {
+            oldDefinition.options.forEach(option => {
                 if (typeof option === `string`) {
                     return;
                 }
                 const newOption = (
                     definition.options.filter(newOpt => typeof newOpt !== `string`) as OsFilterOption[]
-                ).filter(newOpt => newOpt.label === option.label);
-                if (newOption.length) {
-                    newOption[0].isActive = option.isActive;
+                ).find(newOpt => newOpt.label === option.label);
+                if (newOption) {
+                    newOption.isActive = option.isActive;
                 }
             });
         });
--- a/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
+++ b/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
@@ -389,7 +389,7 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements 
     private parseFilters(oldFilters: OsFilter<V>[]): OsFilter<V>[] {
         const newFilterDefs = this.getFilterDefinitions();
         newFilterDefs.forEach(definition => {
-            const oldDefinition = oldFilters ? oldFilters.find(old => old.property === definition.property) : null;
+            const oldDefinition = oldFilters?.find(old => old.property === definition.property);
             if (!oldDefinition) {
                 return;
             }
--- a/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
+++ b/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
@@ -389,7 +389,7 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements 
     private parseFilters(oldFilters: OsFilter<V>[]): OsFilter<V>[] {
         const newFilterDefs = this.getFilterDefinitions();
         newFilterDefs.forEach(definition => {
-            const oldDefinition = oldFilters.find(old => old.property === definition.property);
+            const oldDefinition = oldFilters ? oldFilters.find(old => old.property === definition.property) : null;
             if (!oldDefinition) {
                 return;
             }
--- a/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
+++ b/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
@@ -144,7 +144,7 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements 
         const storedFilter: OsFilter<V>[] | null = await this.activeFiltersStore.load<V>(this.storageKey);
 
         if (storedFilter && this.isOsFilter(storedFilter)) {
-            this.filterDefinitions = storedFilter;
+            this.filterDefinitions = this.parseFilters(storedFilter);
             this.activeFiltersToStack();
         } else {
             this.filterDefinitions = this.getFilterDefinitions();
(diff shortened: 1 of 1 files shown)


client/src/app/site/base/base-filter.service/base-filter-list.service.ts
```diff
@@ -382,6 +382,32 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements
...
+
+    private parseFilters(oldFilters: OsFilter<V>[]): OsFilter<V>[] {
+        const newFilterDefs = this.getFilterDefinitions();
+        newFilterDefs.forEach(definition => {
+            const oldDefinition = oldFilters.filter(old => old.property === definition.property);
+            if (!oldDefinition.length) {
+                return;
+            }
+            oldDefinition[0].options.forEach(option => {
+                if (typeof option === `string`) {
```
jsangmeister, line 397: why do you filter out strings?

luisa-beerboom: The point of this exercise is to get the old isActive values and parse them into the new filter settings from the code. If the option is just a string there is no isActive value to pull out, so they're ignored.

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
```diff
@@ -382,6 +382,32 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements
...
     protected abstract getFilterDefinitions(): OsFilter<V>[];
 
+    private async loadFilters(): Promise<OsFilter<V>[] | null> {
+        return this.parseFilters(await this.activeFiltersStore.load<V>(this.storageKey));
+    }
+
+    private parseFilters(oldFilters: OsFilter<V>[]): OsFilter<V>[] {
+        const newFilterDefs = this.getFilterDefinitions();
+        newFilterDefs.forEach(definition => {
+            const oldDefinition = oldFilters.filter(old => old.property === definition.property);
```
jsangmeister, line 392: `find` would be better suited here, since you ignore all except the first item anyway

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
jsangmeister: Ah, I see. Is the possibility to provide just a string ever used anywhere? If not, maybe we should just remove it.

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
```diff
@@ -382,6 +382,32 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements
...
+            if (!oldDefinition.length) {
+                return;
+            }
+            oldDefinition[0].options.forEach(option => {
+                if (typeof option === `string`) {
+                    return;
+                }
+                const newOption = (
+                    definition.options.filter(newOpt => typeof newOpt !== `string`) as OsFilterOption[]
+                ).filter(newOpt => newOpt.label === option.label);
```
jsangmeister, line 402: Also use `find` here. Also, can you merge the two calls together into one filter like `typeof newOpt !== ``string`` && newOpt.label === option.label`?

jsangmeister (changes requested):

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
luisa-beerboom: I think it is not, I'll look into it

luisa-beerboom: There are indeed some places, where it is used, so I'll leave this in

jsangmeister: ok

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
luisa-beerboom: I tried putting these into the same function, but, even with the check being literally part of the same statement, typescript complains that `newOpt.label` doesn't work if `newOpt` is a string

jsangmeister: Maybe if you use `typeof newOpt === OsFilterOption` instead? But if not, it's no problem.

jsangmeister (approved):

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
luisa-beerboom: I think `typeof` only works with basic types

MSoeb:
Fix didn't work. By opening a meeting, every menu point with a list displays an error in the console. 

Repro:
1. Open a meeting
2. open console
3. navigate to participants, elections, motions...and you will see some error messages.

luisa-beerboom:
I fixed the error messages that I found. If there are any more, could you perhaps give examples of which error messages you're getting and where?

m-schieder (approved):

peb-adr:
@luisa-beerboom 
We're still getting an error in the browser console.
```
main.692390e49836c05b.js:1 ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'find' of null
TypeError: Cannot read property 'find' of null
    at 8065.dc1a39c075c0467c.js:1
    at Array.forEach (<anonymous>)
    at a.parseFilters (8065.dc1a39c075c0467c.js:1)
    at 8065.dc1a39c075c0467c.js:1
    at Generator.next (<anonymous>)
    at e (main.692390e49836c05b.js:1)
    at h (main.692390e49836c05b.js:1)
    at v.invoke (polyfills.968451426b9442c4.js:1)
    at Object.onInvoke (main.692390e49836c05b.js:1)
    at v.invoke (polyfills.968451426b9442c4.js:1)
    at z (polyfills.968451426b9442c4.js:1)
    at polyfills.968451426b9442c4.js:1
    at e (main.692390e49836c05b.js:1)
    at p (main.692390e49836c05b.js:1)
    at v.invoke (polyfills.968451426b9442c4.js:1)
    at Object.onInvoke (main.692390e49836c05b.js:1)
    at v.invoke (polyfills.968451426b9442c4.js:1)
    at M.run (polyfills.968451426b9442c4.js:1)
    at polyfills.968451426b9442c4.js:1
    at v.invokeTask (polyfills.968451426b9442c4.js:1)
```
You can recreate it locally (drev-setup) by loading the page in private window.
My guess is you'll need to add a ` == null` check, which you didn't observe because some filter was selected due to caching.

client/src/app/site/base/base-filter.service/base-filter-list.service.ts
```diff
@@ -382,6 +382,32 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements
...
     protected abstract getFilterDefinitions(): OsFilter<V>[];
 
+    private async loadFilters(): Promise<OsFilter<V>[] | null> {
+        return this.parseFilters(await this.activeFiltersStore.load<V>(this.storageKey));
+    }
+
+    private parseFilters(oldFilters: OsFilter<V>[]): OsFilter<V>[] {
+        const newFilterDefs = this.getFilterDefinitions();
+        newFilterDefs.forEach(definition => {
+            const oldDefinition = oldFilters ? oldFilters.find(old => old.property === definition.property) : null;
```
jsangmeister, line 392:
```suggestion
            const oldDefinition = oldFilters?.find(old => old.property === definition.property);
```

jsangmeister (changes requested):
minor change

jsangmeister (approved):
code looks good, not tested

MSoeb:
looks good, no more errors.

MSoeb (approved):

PR #1440, opened 2022-07-22
Fix the pdf export from the motion Diff version

### LINKED ISSUE #1401: PDF export: Incorrect in the summary of the changes
* If a change goes over several lines, the last line is missing in the summary.
* If several paragraphs are affected in one change, only the line numbers of the first change are displayed. 
* These errors occur also with recommended changes.
* Test if there are any other errors and if the layout could be adjusted.

m-schieder:
* Fix #1401
* If a change spanned several lines, the last line was missing in the summary.
* If an amendment changed multiple paragraphs there were duplications per changed paragraph.


luisa-beerboom (approved):
I have no Idea how these services work, but the changes seem sensible to me in the context of the problems this seeks to fix.
Although I may not be sure what you are doing in certain places, I cant find any fault with the programming. 
If this works I can't complain, so I'm just approving this.

MSoeb (approved):

PR #1442, opened 2022-07-22
Change default of motionlist from tiles to list

### LINKED ISSUE #1400: Make default of the motion view "list" instead of "tiles"

luisa-beerboom:
Closes #1400 

Please check if this works.
Apparently old settings are saved in the cache, so maybe that should be emptied before testing

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.ts
@@ -164,7 +164,7 @@ export class MotionListComponent extends BaseMeetingListViewComponent<ViewMotion
                 .pipe(map(categories => categories?.length > 0))
                 .subscribe(isAvailable => {
                     this._hasCategories = isAvailable;
-                    this.setupTileView(isAvailable);
+                    this.setupListView();
                 }),
             this.motionBlockController.getViewModelListObservable().subscribe(motionBlocks => {
                 this._hasMotionBlocks = motionBlocks.filter(motionBlock => !motionBlock.internal).length > 0;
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.ts
```diff
@@ -366,11 +366,7 @@ export class MotionListComponent extends BaseMeetingListViewComponent<ViewMotion
...
     }
 
-    private setupTileView(isAvailable: boolean): void {
-        if (isAvailable) {
-            this.storage.get<MotionListviewType>(`motionListView`).then(type => (this.selectedView = type || `tiles`));
-        } else {
-            this.selectedView = `list`;
-        }
+    private setupListView(): void {
+        this.storage.get<MotionListviewType>(`motionListView`).then(type => (this.selectedView = type || `list`));
```
jsangmeister, line 370: 2 possibilities:
- you keep the `isAvailable` parameter, because if the saved view type is `tiles`, but none are available (e.g. categories were deleted in between), you will still set the view to `tiles` this way, which will result in an invalid view
- You remove the whole feature of saving the last used view in the storage. The only point where this is used now (meaning one opens the list of motions and sees the tile view) is if the user previously closed the list of motions while in tile view - as soon as he clicked on one of the categories, the last saved type would be `list`. This is a very rare edge case (one will almost always open the list in one category and not stay on the overview page) and also one where it would be okay for me to do the additional click on the tile button to get back into the tile view.

I think I prefer the second approach since it makes the code a lot smaller and easier to read, and the feature is of almost no use anymore.

jsangmeister (changes requested):

jsangmeister (approved):

MSoeb (approved):

PR #1445, opened 2022-07-25
Fix workflow state restriction selection and make run-cleanup

### LINKED ISSUE #1396: Fix `make run-cleanup`
Currently, `make run-cleanup` does not seem to work. It fails with:
```
/app/node_modules/ts-node/src/index.ts:500
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
../tsconfig.json:16:9 - error TS5023: Unknown compiler option 'noImplicitOverride'.

16         "noImplicitOverride": true,
           ~~~~~~~~~~~~~~~~~~~~
../tsconfig.json:18:9 - error TS5023: Unknown compiler option 'noPropertyAccessFromIndexSignature'.

18         "noPropertyAccessFromIndexSignature": true,
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

    at createTSError (/app/node_modules/ts-node/src/index.ts:500:12)
    at reportTSError (/app/node_modules/ts-node/src/index.ts:504:19)
    at create (/app/node_modules/ts-node/src/index.ts:515:36)
    at Object.register (/app/node_modules/ts-node/src/index.ts:398:19)
    at main (/app/node_modules/ts-node/src/bin.ts:162:19)
    at Object.<anonymous> (/app/node_modules/ts-node/src/bin.ts:512:3)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
```
This should be fixed.

Also, (I think at least the last time I tried it) this does only work if the container is not currently running (at least it build the container again even if it is running, which is unnecessary). We should rename it to `make run-cleanup-standalone` (or similar) and add a new `make run-cleanup` which `exec`s into the running client container and executes the cleanup directly.

### LINKED ISSUE #1437: workflow: switiching restrictions several times chrashes page
On the workflow page, a crash occurs when the Restrictions permission is changed multiple times in a row. The page hangs and errors are displayed in the console.

**Error message:** ERROR TypeError: Cannot read properties of undefined (reading 'includes')

**Reproduction:** 
1. open a meeting
2. open motions
3. click on the three-dot-menu and choose workflow
4. now open simple workflow
5. look out for permission restrictions
6. change in rejected the restriction to "can manage motions", reject it, grant it again.
7. Error appears. Page crashes.

**What should happen:** permissions can be selected and deselected without a error. 

Screenshot_20220721_163805

jsangmeister:
fixes #1437 
fixes #1396

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+docker-run=docker run -ti -v `pwd`/client/src:/app/src -p 127.0.0.1:9001:9001/tcp openslides-client-dev
+
 build-dev:
 	docker build -t openslides-client-dev -f Dockerfile.dev .
 
@@ -5,13 +7,16 @@ build-prod:
 	docker build -t openslides-client -f Dockerfile .
 
 run-dev: | build-dev
-	docker run -ti -v `pwd`/client:/app -p 127.0.0.1:9001:9001/tcp openslides-client-dev
+	$(docker-run)
 
 run-dev-interactive: | build-dev
-	docker run -ti -v `pwd`/client:/app -p 127.0.0.1:9001:9001/tcp openslides-client-dev sh
+	$(docker-run) sh
 
-run-cleanup: | build-dev
-	docker run -t -v `pwd`/client:/app openslides-client-dev npm run cleanup
+run-cleanup-standalone: | build-dev
+	$(docker-run) npm run cleanup
+
+run-cleanup:
+	docker exec -it $$(docker ps -a -q  --filter ancestor=openslides-client-dev) npm run cleanup
 
 run-tests: | build-dev
 	docker run -t openslides-client-dev npm run lint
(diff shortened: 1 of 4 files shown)


luisa-beerboom (approved):
I can't give an educated opinion on the docker stuff, but the rest looks fine

jsangmeister:
> I can't give an educated opinion on the docker stuff, but the rest looks fine

For your (and also @bastianjoel) information: You can now execute `make run-cleanup` in the client root folder while the dev setup is running (meaning you executed `make run-dev` from the OS4 root folder) to clean up your code. Alternatively, if the dev setup is not running, you can call `make run-cleanup-standalone` to the same effect, except that maybe the container has to be rebuilt first. Generally, it is not (and should never be) needed to have a local installation of the dependencies, since we are developing exclusively inside docker containers.

PR #1453, opened 2022-07-26
fix scroll position restore if scrollViewport not available yet

### LINKED ISSUE #1436: Scroll position problems: on dashboard level lists scroll position save didn't work / error message in console
Two problems occur in relation to scroll position issue #1380 

- [ ] same problem like in the mentioned issue #1380. difference is only the level. here it affects the dashboard level lists. i assume, that all lists on this level are affected.
- [ ] The second problem is an error message in the console, which currently generates no apparent problems. The error probably occurs when a list view contains few objects. 
- Example on page: https://testintern.openslides.com/70/motions. Open console and navigate to participants and back to motions. 
- The error message is: ERROR TypeError: Cannot read properties of undefined (reading 'measureScrollOffset')

What should happen: 

- Dashboard lists should have the possibility to save the scroll position like meeting lists
- no error message in console

bastianjoel:
resolves  #1436

### DIFF
--- a/client/src/app/site/base/base-list-view.component.ts
+++ b/client/src/app/site/base/base-list-view.component.ts
@@ -102,7 +102,7 @@ export abstract class BaseListViewComponent<V extends BaseViewModel> extends Bas
 
     private saveScrollOffset(): void {
         if (this.listStorageIndex) {
-            this.storage.set(createStorageOffsetIndex(this.listStorageIndex), this.listComponent!.currentOffset);
+            this.storage.set(createStorageOffsetIndex(this.listStorageIndex), this.listComponent?.currentOffset);
         }
     }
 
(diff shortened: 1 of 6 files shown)


jsangmeister (approved):
not tested

MSoeb:
fix don't work: issue still exists for lists committees and accounts on dashboard level.

bastianjoel:
Committee and accounts use different list components. I will have a look if it makes more sense to change the used component or to implement the position restore functionality for that lists too. 

Nevertheless I would still suggest merging this PR even without that change as it fixes some newly introduced issues when accessing other lists.

bastianjoel:
Works now for committee and accounts list

MSoeb (approved):

m-schieder (approved):

PR #1454, opened 2022-07-27
[WIP] reduce autoupdate connections

### LINKED ISSUE #1339: New update strategy for the client
Currently, the client requests updates for collections by making several TCP-connections to the AU-service per tab per browser.
There could be up to 20 connections per tab. This is inefficient, if a user uses OpenSlides in multiple tabs.

There are the following options to improve the user experience:
- Combine different connections to one (this will end up in ~4 connections per tab)
- Use the local storage of a browser

The last option is the preferred one. Implement it as described:
- Every time a connection should be established, it will be checked, if there is already made a connection for the same name in another tab
- If a connection was established, append an established connection to a list and write the list in the local storage of a browser
- Write incoming data as a map into the local storage, as well, in the following form:
```ts
interface Map {
  [collection: string]: {
    [id: number]: BaseModel;
  }
}
```
- Listen in the repositories to the data in the local storage instead of their in-memory storage

This should be equivalent to the autoupdate format receiving from the AU-service.

bastianjoel:
This is a first attempt to resolve #1339.

The PR tries to reduce the open connections by collecting autoupdate subscription requests for a short duration and then sending them in one request.

### DIFF
--- a/client/src/app/site/pages/meetings/services/sequential-number-mapping.service.ts
+++ b/client/src/app/site/pages/meetings/services/sequential-number-mapping.service.ts
@@ -87,7 +87,8 @@ export class SequentialNumberMappingService {
         }
         this._modelRequestSubscription = this.autoupdateService.subscribe(
             await this.modelRequestBuilder.build(this.getSequentialNumberRequest()),
-            MODEL_REQUEST_DESCRIPTION
+            MODEL_REQUEST_DESCRIPTION,
+            true
         );
     }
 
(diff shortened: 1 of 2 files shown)


jsangmeister:
What's the status here?

bastianjoel:
Before merging it needs to be tested if this has any side effects. Also I would like to check if it might be better to place this logic inside the `CommunicationManagerService.` 
Furthermore it should be discussed if this is needed anyway, because with the upcoming change to the autoupdate service where we can add subscriptions to a existing connection this would be obsolete.

bastianjoel:
this logic will be implemented inside the worker in #1556

PR #1461, opened 2022-08-02
replace deprecated apis from SlideManagerService

### LINKED ISSUE #1239: Remove deprecated APIs and use latest APIs of Angular for the `SlideManagerService`
Since Angular@v13, there are some APIs (Compiler, ComponentFactory, NgModuleFactory) that are in use by the `SlideManagerService` and are deprecated. However, there are also other APIs that achieve the same behavior, but aren't deprecated. Those APIs should be used and replace the deprecated ones.

bastianjoel:
resolves #1239


rrenkert (approved):
Looks good

Issue #1312, opened 2022-06-23
In the call list: Put a motion as a child of another motion leads to incorrectly sorted motion list


luisa-beerboom:

luisa-beerboom:
If the correct result is to keep the sorting in the motionlist the same as before, I cannot replicate this.

jsangmeister:
me neither

luisa-beerboom:
If anyone can replicate this, please fix it, or at least write a description of how to do it

PR #1475, opened 2022-08-08
Fix sorting by calllist

jsangmeister:
fixes https://github.com/OpenSlides/openslides-client/issues/1312

fixes an error in the tree service where it was always assumed that the children property was named `children` and also changed it. Also fixes another error where sorting by an array of fields was not correctly saved and loaded.

### DIFF
--- a/client/src/app/site/base/base-sort.service/base-sort-list.service.ts
+++ b/client/src/app/site/base/base-sort.service/base-sort-list.service.ts
@@ -181,12 +181,12 @@ export abstract class BaseSortListService<V extends BaseViewModel>
         }
     }
 
+    /**
+     * Determines if the given properties are either the same property or both arrays of the same
+     * properties.
+     */
     private getPropertiesEqual(a: OsSortProperty<V>, b: OsSortProperty<V>): boolean {
-        if (Array.isArray(a) && Array.isArray(b)) {
-            return a.every((val, idx) => val === b[idx]);
-        } else {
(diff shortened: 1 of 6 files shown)


client/src/app/site/base/base-sort.service/base-sort-list.service.ts
```diff
@@ -181,6 +181,14 @@ export abstract class BaseSortListService<V extends BaseViewModel>
         }
     }
 
+    private getPropertiesEqual(a: OsSortProperty<V>, b: OsSortProperty<V>): boolean {
```
luisa-beerboom, line 184: Just an idea, but have you thought about maybe making this a more generally accessible function (as in `getPropertiesEqual(a: any, b: any)` and exporting it maybe from somewhere in the infrastructure folder? This looks like the type of functionality which we might need in all kinds of different places in the future...

luisa-beerboom (approved):
Looks good, I have a minor comment but it's nothing mandatory.
If you feel like it's not too much of a bother you can do that before merging, otherwise just merge this as is.

client/src/app/site/base/base-sort.service/base-sort-list.service.ts
jsangmeister: Good idea, will do that

jsangmeister: Only extracted the array equality part since comparing sort properties is a rather special case.

MSoeb (approved):

PR #1477, opened 2022-08-08
General bug fixes and improvements for user imports

### LINKED ISSUE #1406: csv user import: Various attributes generate error messages when read in / comment field is missing
When importing users via csv file, various attributes generate error messages. 

Furthermore, the comment field that was already present in OS3 is completely missing in OS4. However, it is present in the csv file. 

Affected attributes:

- [x] german gender terms could not be used
- [x] a physical person can only be read in with the value 1 
- [x] the attribute 'active' can only be read in with value 1
- [x] the user comment field is missing 
- [x] pronoun field is missing in csv example file. import not possible.
- [x] New data is not added for existing user when referencing is done

Reproduction:
1. open a meeting
2. open participants
3. open three-point-menu and select import
4. Download csv sample file
5. fill file with affected attributes (see list) and upload.

What should happen: 
- german gender terms could be used in csv file
- zero is a valid value for 'active' and 'physical person' in csv file
- comment field is integrated
- pronouns could be imported
- new data should/could be added for existing users
  - all new meeting specific information should be kept
  - all orga information (title etc.), if the user is in the meeting's scope (see https://github.com/OpenSlides/OpenSlides/wiki/Users#user-scopes)

Example images:

Screenshot_20220718_075140

Screenshot_20220718_080841

luisa-beerboom:
Closes #1406 

Done:

- Can use gender terms of selected language
- Pronouns included in imports/exports
- Fixed problem with boolean fields (where only true values were accepted)
- Comments can now be imported/exported and can be set/viewed by users with user.can_manage permission
- I believe the application of new data after referencing should work now

TODO:

- [x] Look if the account import is functioning correctly

### DIFF
--- a/client/src/app/site/base/base-user-import.service.ts
+++ b/client/src/app/site/base/base-user-import.service.ts
@@ -1,12 +1,14 @@
 import { Directive } from '@angular/core';
 
-import { User } from '../../domain/models/users/user';
+import { GENDERS, User } from '../../domain/models/users/user';
 import { ImportServiceCollectorService } from '../services/import-service-collector.service';
 import { BaseImportService } from './base-import.service';
 
 @Directive()
 export abstract class BaseUserImportService extends BaseImportService<User> {
     public override requiredHeaderLength = 3;
+
+    private readonly currentLangGenders = GENDERS.map(gender => this.translate.instant(gender));
 
     public constructor(importServiceCollector: ImportServiceCollectorService) {
         super(importServiceCollector);
@@ -15,6 +17,10 @@ export abstract class BaseUserImportService extends BaseImportService<User> {
     protected override pipeParseValue(value: string, header: keyof User): any {
         if (header === `is_active` || header === `is_physical_person`) {
             return this.toBoolean(value);
+        }
+
+        if (header === `gender`) {
+            return this.getGenderInEnglish(value);
         }
 
         if (header === `first_name` || header === `last_name` || header === `username`) {
@@ -36,5 +42,19 @@ export abstract class BaseUserImportService extends BaseImportService<User> {
         } else {
             throw new TypeError(`Value cannot be translated into boolean: ` + data);
         }
+    }
+
+    /**
+     * translates gender string to the english version if necessary
+     *
+     * @param data
+     * @returns a boolean from the string
+     */
+    private getGenderInEnglish(data: string): string {
+        const location = this.currentLangGenders.findIndex(gender => gender === data);
+        if (location !== -1) {
+            return GENDERS[location];
+        }
+        return data;
     }
 }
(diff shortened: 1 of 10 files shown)


client/src/app/site/base/base-user-import.service.ts
```diff
@@ -16,6 +18,9 @@ export abstract class BaseUserImportService extends BaseImportService<User> {
         if (header === `is_active` || header === `is_physical_person`) {
             return this.toBoolean(value);
         }
+        if (header === `gender`) {
+            return this.getGenderInEnglish(value);
```
jsangmeister, line 22: Is this analogous to the group name translation matching? I can't seem to find it.

luisa-beerboom: It probably is, I wrote this function because I didn't find the place where we do that, but I guess I'll look again.

luisa-beerboom: Regrettably I still can't find it

m-schieder:
Can this be tested or is it still being worked on?

client/src/app/site/base/base-user-import.service.ts
jsangmeister: ok

luisa-beerboom:
> Can this be tested or is it still being worked on?

Since I probably won't be able to find the place where we handle the translation of the group names, and since this is functionally complete, I don't think anything more will happen here for now, so feel free to test this.

MSoeb:
The import works as it should. 

But, the import UI misses a coherent structure for attributes with the values 1, 0, true and false. OS3 had check boxes for this values. But this is another issues. ;) I will create one for it.

MSoeb (approved):

client/src/app/site/base/base-user-import.service.ts
```diff
@@ -1,13 +1,15 @@
...
-import { User } from '../../domain/models/users/user';
+import { GENDERS, User } from '../../domain/models/users/user';
 import { ImportServiceCollectorService } from '../services/import-service-collector.service';
 import { BaseImportService } from './base-import.service';
 
 @Directive()
 export abstract class BaseUserImportService extends BaseImportService<User> {
     public override requiredHeaderLength = 3;
 
+    private readonly currentLangGenders = GENDERS.map(gender => this.translate.instant(gender));
```
jsangmeister, line 11: This looks risky: depending on the loading order, the translations might not be loaded yet when this class is initialized. `translate.instant` does not take this into account and just returns the english word if no translations are loaded yet (see https://github.com/ngx-translate/core#methods). Maybe you should use `get`/`stream` instead and update it on updates to the translation.

Another thing: What happens if you change the language while on this page? That would probably break the import as well.

jsangmeister (changes requested):

client/src/app/site/base/base-user-import.service.ts
luisa-beerboom: Good point, I'll change it

luisa-beerboom: Although: About the "Change the language"-thing. As long as you have uploaded a file, that would probably happen anyway unless the uploaded file is in English, so I probably won't be going any further than changing the way currentLangGenders works.

jsangmeister (approved):

PR #1479, opened 2022-08-09
Imports: Filtered out empty strings after splitting data

### LINKED ISSUE #1371: CSV-Imports: Filter out empty strings when splitting up data
CSV-imports:
There are some fields where multiple entries can be given by separating them through a semicolon.
If there is a semicolon at the end of the line, this may cause problems, as the client may attempt to send a create request with an empty string as a payload to the backend (causing errors).

This can be problematic for massive imports that create (f.e.) a lot of users.

Proposed solution:
Before processing field entries, trim empty spaces from the end and remove the last letter if it is a semicolon.

luisa-beerboom:
Closes #1371


MSoeb (approved):

MSoeb:
works fine

rrenkert (approved):

PR #1484, opened 2022-08-10
Non group-related meeting user updates no longer set group to standard

### LINKED ISSUE #1480: When users are set as non-natural person by multiple selection, the group assignment changes
When users who already exist in a meeting are changed to a non-natural person via multiple selection, the group assignment also changes. The group changes to default. 

**Reproduction:**
1. open meeting
2. open participants menu
3. click on _three-dot menu_ and _select multiple selection_
4. select 1-2 users with a group assignment (for example delegates / NOT standard)
5. click again on _three-point menu_, select _Set natural person_ and click on _no natural person_
--> Group assignment changes to standard

**What should happen:**
When users are set as non-natural persons, the group assignment should be preserved.

luisa-beerboom:
Closes #1480 

The change in this PR holds some danger of affecting some other functions, that are actually supposed to set groups. 

In essence: 
There _may_ now be problems if you attempt to remove the participant from all groups. Instead of being set to standard, the user _may_ now be kicked out of the meeting instead.

I have tried changing a users group in that way via the dialog in the participant list and also in the edit view and it still worked as intended, so I'm hopeful, but this should definately be tested more thouroughly.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/participants/services/common/participant-controller.service/participant-controller.service.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/services/common/participant-controller.service/participant-controller.service.ts
@@ -232,7 +232,7 @@ export class ParticipantControllerService extends BaseMeetingControllerService<V
                 [this.activeMeetingId!]: participant.structure_level
             },
             group_$_ids: participant.group_$_ids || {
-                [this.activeMeetingId!]: participant.group_ids || [this.activeMeeting.meeting!.default_group_id]
+                [this.activeMeetingId!]: participant.group_ids
             },
             number_$: participant.number_$ || { [this.activeMeetingId!]: participant.number },
             vote_weight_$: participant.vote_weight_$ || {


jsangmeister:
It seems like you last changed this regarding the motion import: https://github.com/OpenSlides/openslides-client/commit/11212076c3127b56510ff7ca5a2e285fc19d93e4. This should therefore also be checked. I assume that this is not going to work now since I don't see where the default group will be set on the user, but I think it was the wrong place before anyway: the import should set the group in the participant object, then the changes of this PR can stay.

luisa-beerboom:
> It seems like you last changed this regarding the motion import: [1121207](https://github.com/OpenSlides/openslides-client/commit/11212076c3127b56510ff7ca5a2e285fc19d93e4). This should therefore also be checked. I assume that this is not going to work now since I don't see where the default group will be set on the user, but I think it was the wrong place before anyway: the import should set the group in the participant object, then the changes of this PR can stay.

I tried it out, it still seems to work just fine

jsangmeister:
> I tried it out, it still seems to work just fine

ok, interesting, but even better

MSoeb:
Works fine. Can't find any problem.

MSoeb (approved):

rrenkert (approved):

PR #1487, opened 2022-08-10
Can now access motion block speaker list before projection

### LINKED ISSUE #1482: New Motion Block has no List of Speakers until it is projected
A new motion block is created without a visible list of speakers. The list of speakers will not be visible until the motion block is projected onto a projector. 

Note: It was also reported by a user that when opening the list of speakers of a motion block in the agenda, an error message occurs. This is "_undefined is not an object (evaluating 'S.id')_". Unfortunately, I have not been able to reproduce this so far. I suspect that this is related to the error described. 

**Reproduction:**
1. open motions menu
2. click on three-point menu and select _motion blocks_
3. create a new motion block in the motion block menu. To do this, click on the _plus symbol_, assign a title and click _Save_.
4. The new application block is now visible, but without the list of speakers icon. 
5. Click on the camera icon, select the standard projector and click on Project.
6. The list of speakers icon appears. 

**Image with missing list of speakers icon:**
Screenshot_20220810_090258

**Image with list of speakers icon after projection:**
Screenshot_20220810_090318

**What should happen?**
The list of Speakers icon should be visible immediately after creating a motion block. (see comparison OS3)

luisa-beerboom:
Closes #1482

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/config/model-subscription.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/config/model-subscription.ts
@@ -21,7 +21,7 @@ export const getMotionBlockSubscriptionConfig = (id: Id, hasMeetingIdChangedObse
     modelRequest: {
         viewModelCtor: ViewMeeting,
         ids: [id],
-        follow: [`motion_block_ids`]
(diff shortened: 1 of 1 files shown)


rrenkert (approved):
Works for me.

PR #1488, opened 2022-08-10
Changed styling for meeting preview participant badges for big numbers

### LINKED ISSUE #1410: committee menu: blue circle can't display large number of participants
When a meeting has more than 99 participants, a layout problem occurs. The blue circle is too small to display large numbers correctly. A three-digit value fits just inside the circle. But it already touches the outer border, which is not optimal. If the value has four digits, the number is no longer displayed.

Reproduction
Create three meetings in committee. One with 99, 999 and more than 1000 users.

What should happen: The blue circle should be large enough to correctly contain and display at least four-digit or five-digit numbers. This might require rearranging the layout.

Example image: meeting three has 1031 participants
Screenshot_20220720_102559

luisa-beerboom:
… up to 9999

closes #1410 

Part of the problem was, that the badges (i.e. the blue circles) weren't big enough to fit 4-digit numbers. They'd be displayed as something like '1...', '2...', and so on, which is why such numbers were hidden.

This PR addresses this by making the font size smaller for numbers that are bigger than 9999 and raising the maximum limit to 99999 (after which the numbers will still be hidden). Numbers bigger than 999 will be shown as >1k, >2k, >3k, ...

### DIFF
--- a/client/src/app/site/pages/organization/pages/committees/modules/committee-meeting-preview/committee-meeting-preview.component.html
+++ b/client/src/app/site/pages/organization/pages/committees/modules/committee-meeting-preview/committee-meeting-preview.component.html
@@ -64,13 +64,13 @@
 </mat-card>
 
 <ng-template #canNotManageUsersTemplate>
-    <div class="fake-button-no-touch" matTooltip="{{ 'Participants' | translate }}">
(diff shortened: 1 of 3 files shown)


luisa-beerboom:
@m-schieder I just pushed a new commit, the test instance should probably be updated

jsangmeister (approved):

emanuelschuetze (approved):

PR #1493, opened 2022-08-12
enable karma tests

### LINKED ISSUE #1001: Re-activate the karma tests
They are currently de-activated because they made to many problems in the past. However, some of them are really good worked out (for example the tests for the motion-diff algorithm) and should be re-activated.

bastianjoel:
resolves #1001 

This PR includes the following changes:
- disalbes emitDecoratorMetadata **(this might need some discussion)**
- Use jsdom as default browser for karma tests
- Use chromium for live testing
- Disable all auto generated tests
- Uncomment karma tests in Makefile
- Add run script for tests to Makefile
- Run tests in CI (build-and-check-dev-image)
- Add tests from OS3
  - DiffService -> MotionDiffService
  - LinenumberingService -> LineNumberingService


jsangmeister (changes requested):
Looks very good! Some points:
- Why did you change all `describe`s to `xdescribe`s?
- Some tests are commented out, is this on purpose?
- Can you maybe provide a default configuration for all `TestBed`s automatically where the `E2EImportsModule` is already imported, so that you don't have to do that for all tests where it is needed?

bastianjoel:
The `x` Infront of `describe` disables a test. 
Most of these disabled tests fail due to not added imports and declarations to the TestBed and because they only created components or services and checked if they were successfully instantiated I decided to disable them. My main goal with this PR is to get the karma test running and add the motion diff tests from OS3. 

Commenting tests out was not on purpose. I will change them to `xdescribe`.

I think adding the `E2EImportsModule` could be done later when we know how many tests actually need it.

jsangmeister:
> The `x` Infront of `describe` disables a test.

Ah, I confused it with `fdescribe`. Ok, sounds good

jsangmeister (approved):

PR #1496, opened 2022-08-16
bullet point ammendment fix

### LINKED ISSUE #1382: Amendment with new line above bullet points generates line 0 in motion and marks original content incorrectly red
If a user creates an amendment which adds a line above bullets points, a line with number zero is created and the motion is marked incorrectly in red. 

Reproduction:
1. create motion with two paragraphes and a bullet point list with at least one entry
2. create an amendment. 
3. Select the part with the bullet point. 
4. Create a new line at the beginning of the bullet point with Enter, delete the new generated bullet point and add a new line of text. -> line zero appears in amendment.
5. Switch back to motion. All of original content is marked red. 

What should happen: Generate a new line above a bullet point in an amendment, should only add a new line instead of creating line zero and crashing the diff view.

Example: https://testintern.openslides.com/2/motions/64

Screenshot_20220718_122003

bastianjoel:
resolves #1382


luisa-beerboom (approved):

MSoeb:
works fine.

MSoeb (approved):

PR #1497, opened 2022-08-16
Validation of meeting time

### LINKED ISSUE #1368: validate meeting date
When creating a meeting the start_date and stop_date are not validated.
Specifically end_date can be lower than start_date which shouldn't be the case.

This should be possible relatively easy using angulars validators I think.

luisa-beerboom:
Closes #1368 

It should now be impossible to create a meeting with an end-time that is earlier than the start time.
This should be the same for editing meetings

### DIFF
--- a/client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/committee-detail-meeting.module.ts
+++ b/client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/committee-detail-meeting.module.ts
@@ -3,6 +3,7 @@ import { NgModule } from '@angular/core';
 import { FormsModule, ReactiveFormsModule } from '@angular/forms';
 import { MatCardModule } from '@angular/material/card';
 import { MatFormFieldModule } from '@angular/material/form-field';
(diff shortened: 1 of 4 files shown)


jsangmeister (approved):

emanuelschuetze (changes requested):
An error message is not really nice... I expected: If I changed the start date, the end date should be set to start date. Than I can save directly (without changing end date) if my meeting is a one-day-event.

luisa-beerboom:
> An error message is not really nice... I expected: If I changed the start date, the end date should be set to start date. Than I can save directly (without changing end date) if my meeting is a one-day-event.

So in other words: 
If the user changes one date and the result is valid, do nothing further. 
If the user changes one date and the result is invalid, change the other so that they are the same date?

jsangmeister (changes requested):
I would remove the message completely if we go the way of automatically changing the time and not change the other date upon submition of the form, but as soon as the other date is changed. So, I enter an end date which is earlier than my start date, my start date will automatically be set to the new end date. This will also simplify the code a lot.

jsangmeister (approved):

emanuelschuetze (approved):

PR #1499, opened 2022-08-16
fix ammendment text import

### LINKED ISSUE #1407: motion csv import bug: amendments not possible
After a motion has been imported via csv import, it is not possible to create an amendment. When you try to create the amendment, the text sections that can be selected are missing the motion text. 

Reproduction:
1. open a meeting
2. open menu Motions
3. open three-item menu and select Import
4. download example csv file and import the same file via "Select file"
5. open one of the uploaded motions 
6. create an amendment via plus symbol and select a paragraph -> no text available

What should happen: After motion import via csv file it should be possible to create amendments with correct available text sections.

Example:
Screenshot_20220720_085144

bastianjoel:
resolves #1407

For the amendment line selection the text needs to be surrounded by HTML tags. This PR adds surrounding `p` tags to motion text if not existing and tries to divide the text into sections.


luisa-beerboom (approved):

emanuelschuetze:
@MSoeb Can you test it?

MSoeb:
looks fine.

PR #1503, opened 2022-08-17
display error if min votes is larger than amount candidates

### LINKED ISSUE #720: Warn if the minimal votes exceed the amount of candidates
If the amount of votes (in assignment poll voting) is larger than the amount of candidates the poll cannot be send. There should be a warning and saving should not be possible.

bastianjoel:
resolves #720


luisa-beerboom (approved):
Looks good, not tested

emanuelschuetze (approved):

PR #1517, opened 2022-08-23
update dates after language change

### LINKED ISSUE #1350: After a language change, all dates should be updated
1. goto https://demo.os4.openslides.com/
2. change your language (either en->de or vice-versa)
3. the dates in the meeting list are only updated when you click something (e.g. on a meeting)

They are however updated before the redirecting takes place, so it seems to me like a missed change detection call.

bastianjoel:
resolves #1350


luisa-beerboom (approved):

luisa-beerboom:
Not tested

PR #1518, opened 2022-08-23
beautify vote entitled users

### LINKED ISSUE #852: Beautify detail view of a vote
The list "Single votes" looks very nice. The first line contains the name of the person, the second line contains the structure level (text is grayed out, which looks nice).

The list "Entitled users" on the other hand is the ugly brother/sister. The design of the list here should be just as pretty as the "Single votes" list is. So:

- The first line contains just the name of the person
- The second line contains the structure level (text grayed out).
- The list uses its width wisely (less wrapping)

The icon for "Vote submitted" (I assume) is currently a check mark. This is a bit confusing, because users might assume that the icon indicates presence. But this is not the case.

That's why I suggest: The icon should be changed to `<mat-icon>how_to_vote</mat-icon>`  and behind it should be the text "vote submitted" or "vote has been cast" or something like that.

bastianjoel:
resolves #852

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.html
+++ b/client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.html
@@ -5,19 +5,35 @@
             <!-- Content -->
             <div *osScrollingTableCell="'user_id'; row as entry">
                 <div *osScrollingTableCellLabel>{{ 'Participant' | translate }}</div>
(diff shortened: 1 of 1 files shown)


luisa-beerboom (approved):

emanuelschuetze (changes requested):
Changes looks nice. But still missing is the column with:
Please move the icon column to the right and increase the name column to maximum to prevent line breaks for long names.

emanuelschuetze (approved):

PR #1524, opened 2022-08-24
Made meeting-internal delete use handle separately route

### LINKED ISSUE #1523: Use handle_separately route for meeting-internal delete action
This should fix the problem described in OpenSlides/openslides-backend#1426

luisa-beerboom:
Closes #1523 

This solves the problem described in OpenSlides/openslides-backend#1426 where trying to remove multiple participants from a meeting using multiselection would cause datastore broken lock errors when among the selected users the following was true:
There is at least one group that has:
1. At least two selected users
2. Among the selected users of that group there is at least one who is in another meeting, and at least one who is in no other meeting.

### DIFF
--- a/client/src/app/gateways/actions/action.service.ts
+++ b/client/src/app/gateways/actions/action.service.ts
@@ -6,6 +6,7 @@ import { ActionRequest, isActionError, isActionResponse } from './action-utils';
 
 type ActionFn = () => boolean;
 const ACTION_URL = `/system/action/handle_request`;
+const ACTION_SEPARATELY_URL = `/system/action/handle_separately`;
 
 let uniqueFnId = 0;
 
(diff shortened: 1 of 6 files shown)


client/src/app/gateways/actions/action.service.ts
```diff
@@ -49,10 +50,17 @@ export class ActionService {
         return null;
     }
 
+    /**
+     * @deprecated this does not offer the handle_separately route option, which is vital for certain types of bulk requests, it's better to use `createFromArray` instead.
+     */
     public create<T>(...requests: ActionRequest[]): Action<T> {
```
jsangmeister, line 56: Can't you just add the optional parameter to this method? I don't see the use of the new `createFromArray` method.

jsangmeister (changes requested):

client/src/app/gateways/actions/action.service.ts
jsangmeister: Ah, I see. I can't estimate the time to fix this, but if it's too much, I'm fine with the way it is.

jsangmeister (approved):

emanuelschuetze (approved):

PR #1526, opened 2022-08-24
Beautification of the password reset page

### LINKED ISSUE #757: Password reset page
There are some low priority issues on the password reset page

1. It would be nice, if the focus would be on the e-mail field. Currently it is necessary to click on it.
2. After clicking on the "reset" button, nothing happens (if the server is slow). There should be some feedback, that the button was clicked. For example disable it.

luisa-beerboom:
Closes #757 

Currently the page will throw an error: `Error when trying to forward to only meeting: Operator has not fully loaded yet.` if one clicks the reset password button.

This is triggered by a navigation command. 
Once I have the information, where that button is supposed to navigate to, I shall fix that and request reviews

### DIFF
--- a/client/src/app/site/pages/login/pages/reset-password-confirm/components/reset-password-confirm/reset-password-confirm.component.scss
+++ b/client/src/app/site/pages/login/pages/reset-password-confirm/components/reset-password-confirm/reset-password-confirm.component.scss
@@ -1,6 +1,9 @@
 h3 {
     margin-top: 0;
     font-weight: normal;
+    width: 100%;
+    margin-bottom: 30px;
+    text-align: center;
 }
 
 mat-form-field {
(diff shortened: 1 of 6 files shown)


luisa-beerboom:
It should navigate back to login page and open a snackbar like "Es wurde eine E-Mail mit Link zum Passwort-Zurücksetzen gesendet." there

emanuelschuetze (changes requested):
The forget-password view works now as expected (in issue #757).
But after clicking on email confirm link you see the forget-password-confirm which has broken css...
password-confirm

luisa-beerboom:
> The forget-password view works now as expected (in issue #757). But after clicking on email confirm link you see the forget-password-confirm which has broken css... password-confirm

That looks bad indeed, how do I get to that page in my dev environment?

emanuelschuetze (approved):

bastianjoel (approved):

PR #1538, opened 2022-08-29
Fixed broken assignment poll result display for deleted candidates

### LINKED ISSUE #1531: Election result is deleted after deleting a user
Problem: If a user, who was part of an election, is deleted out of the Accounts list, the election result is broken. The result are no more visible. Additionally the console shows an error.

error message
Screenshot_20220825_135316

broken election result after deletion
Screenshot_20220825_135337

Reproduction:
1. open a meeting
2. open elections
3. create an election via plus-icon, fill out the form (need titel. other content doesn't matter) and save
4. Add a few candidates and one account, that should be deleted
5. click on button "new ballot"
6. fill out form. only point that should be changed is "enttiled to vote". here you should enter your group name. save it. 
7. Now start voting by clicking on "start voting"-button
8. Vote and click on "Stop voting"-button and than "stop & Publish"-Button
9. Go to the Accounts menu and delete one of candidates. 
10. open the election result again. 

What should happen:
After deleting a user, the election result should be available and the field with the deleted user account should be replaced with a field like "gelöschte*r Nutzer*in".

luisa-beerboom:
Closes #1531

### DIFF
--- a/client/src/app/gateways/repositories/assignments/assignment-candidate-repository.service/assignment-candidate-repository.service.ts
+++ b/client/src/app/gateways/repositories/assignments/assignment-candidate-repository.service/assignment-candidate-repository.service.ts
@@ -4,6 +4,7 @@ import { Identifiable } from 'src/app/domain/interfaces';
 import { Action } from 'src/app/gateways/actions';
 import { BaseMeetingRelatedRepository } from 'src/app/gateways/repositories/base-meeting-related-repository';
 import { ViewAssignmentCandidate } from 'src/app/site/pages/meetings/pages/assignments';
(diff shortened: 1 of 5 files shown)


emanuelschuetze (changes requested):
Works good for assignment detail view. But if you project the assignment poll (with deleted candidate) the projector/autopilot shows type error:

```TypeError: Cannot read properties of undefined (reading 'collection')
TypeError: Cannot read properties of undefined (reading 'collection')
    at Object.getOptionTitle (poll-slide.component.ts:122:94)```

emanuelschuetze (changes requested):
`NullInjectorError: No provider for AssignmentPollService!` if you project an assignment poll (with deleted candidate) on a _fullscreen_ projector (not autopilot)

PR #1541, opened 2022-08-30
1190 fix font deselection

### LINKED ISSUE #1190: Cannot deselect custom font
When a custom font is selected, it cannot be deselected.

rrenkert:
Closes #1190

### DIFF
--- a/client/src/app/domain/models/base/base-model.ts
+++ b/client/src/app/domain/models/base/base-model.ts
@@ -45,7 +45,7 @@ export abstract class BaseModel<T = any> implements Identifiable, Deserializable
         const updateCopy = { ...update }; // To not modifying the original update
         for (const key of Object.keys(update)) {
             const value = update[key as keyof T] as unknown;
-            if (isTemplateField(key) && Array.isArray(value) && value.length >= 0) {
+            if (isTemplateField(key) && Array.isArray(value)) {
                 this.handleRemovedTemplateFields({ origin, update: updateCopy, key: key as keyof T });
             }
         }
--- a/client/src/app/domain/models/base/base-model.ts
+++ b/client/src/app/domain/models/base/base-model.ts
@@ -45,7 +45,7 @@ export abstract class BaseModel<T = any> implements Identifiable, Deserializable
         const updateCopy = { ...update }; // To not modifying the original update
         for (const key of Object.keys(update)) {
             const value = update[key as keyof T] as unknown;
-            if (isTemplateField(key) && Array.isArray(value) && value.length === 0) {
+            if (isTemplateField(key) && Array.isArray(value) && value.length >= 0) {
                 this.handleRemovedTemplateFields({ origin, update: updateCopy, key: key as keyof T });
             }
         }
--- a/client/src/app/domain/models/mediafiles/mediafile.ts
+++ b/client/src/app/domain/models/mediafiles/mediafile.ts
@@ -47,13 +47,7 @@ export class Mediafile extends BaseModel<Mediafile> {
      */
     public used_as_logo_in_meeting_id(place?: string): Id | null {
         if (!place) {
-            for (let i = 0; i < this.used_as_logo_$_in_meeting_id.length; i++) {
-                const path = `used_as_logo_$${this.used_as_logo_$_in_meeting_id[i]}_in_meeting_id` as keyof Mediafile;
-                if (path in this) {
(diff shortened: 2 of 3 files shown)


client/src/app/domain/models/base/base-model.ts
```diff
@@ -45,7 +45,7 @@ export abstract class BaseModel<T = any> implements Identifiable, Deserializable
         const updateCopy = { ...update }; // To not modifying the original update
         for (const key of Object.keys(update)) {
             const value = update[key as keyof T] as unknown;
-            if (isTemplateField(key) && Array.isArray(value) && value.length === 0) {
+            if (isTemplateField(key) && Array.isArray(value) && value.length >= 0) {
```
luisa-beerboom, line 48: `value.length >= 0` should always be true, either remove it or replace it with a condition that does, what this is intended to do there

client/src/app/domain/models/mediafiles/mediafile.ts
```diff
@@ -38,12 +38,44 @@ export class Mediafile extends BaseModel<Mediafile> {
...
-    public used_as_logo_in_meeting_id(place: string): Id | null {
+    /**
+     * Check if the image is used at the given place. If place is empty, checks
+     * if the image is used in any place and returns the first meeting id found.
+     *
+     * @param place The places the image can be used
+     * @returns the meeting id or `null` if the image is not used.
+     */
+    public used_as_logo_in_meeting_id(place?: string): Id | null {
+        if (!place) {
```
luisa-beerboom, line 49: This function is extremely similar to `used_as_font_in_meeting_id`. Maybe look into creating a helper function that handles the shared functionality.

luisa-beerboom (changes requested):

client/src/app/domain/models/base/base-model.ts
rrenkert: Ups. You are right ;)
Removed it.

client/src/app/domain/models/mediafiles/mediafile.ts
luisa-beerboom: This is better, but wouldn't it be possible to also handle the rest of these functions in `used_in_meeting`, as the old code was already basically the same?

client/src/app/domain/models/mediafiles/mediafile.ts
```diff
@@ -38,16 +38,46 @@ export class Mediafile extends BaseModel<Mediafile> {
...
+     */
+    public used_as_font_in_meeting_id(place?: string): Id | null {
+        if (!place) {
+            return this.used_in_meeting(this.used_as_font_$_in_meeting_id, 'font');
+        }
         const path = `used_as_font_$${place}_in_meeting_id` as keyof Mediafile;
         return (this[path] as Id) || null;
     }
 
+    private used_in_meeting(list: string[], type: string): Id | null {
```
luisa-beerboom, line 71: I think this could still easily handle the rest of the original functionalities.

Also if you're already creating the path explicitly, you might as well get the list within this function by using the bracket notation (i.e. by calling `this['used_as_${type}_$_in_meeting_id']` inside the method). 
That way you don't have to specifically ask for the list as a parameter.

luisa-beerboom (changes requested):

client/src/app/domain/models/mediafiles/mediafile.ts
rrenkert: For sure this is possible! I thought of these function a bit more type specific like 'isFont' or something. Therefore I left the original function names and the code duplication.
To resolve this like you stated means the type has to be a parameter. Would that be better?

luisa-beerboom: I think whether or not you should keep the original function names depends on how you feel about changing all the function calls, but the code duplication should probably be built out anyway.

luisa-beerboom (approved):

emanuelschuetze (changes requested):
Tested: I get a lot of type errors after opening media list view which contains 2 font files. Cannot see 3-dot-menu for 2nd file.
```
TypeError: Cannot read properties of undefined (reading 'length')
    at Mediafile.used_in_meeting (mediafile.ts:66:38)
```

rrenkert:
Should be fixed now. For new files the list of possible places of usage is undefined not empty array.

PR #1544, opened 2022-08-30
Fix ICC by bypassing the ngsw

jsangmeister:
The previous method of telling the service worker to ignore certain paths and not cache them did not work reliably anymore (for unknown reasons). This lead to ALL http calls being intercepted by the service worker, which was annoying for "normal" calls, but fatal for blocking calls, since the service worker did not seem to be able to handle these, which lead in turn to the ICC service not working correctly because even the first message, the channel id, did not get through. 

This PR introduces an alternative solution using the `ngsw-bypass` header to explicitly bypass the service worker for all http calls. These are currently calls to the AU, backend, auth, vote and notify service, which should all never be cached, so I inserted the parameter at the lowest level in the http service since I assume that all requests from inside the client will be of such nature, even in the future.

### DIFF
--- a/client/ngsw-config.json
+++ b/client/ngsw-config.json
@@ -25,16 +25,5 @@
                 "files": ["/assets/**", "/*.(svg|cur|jpg|jpeg|png|apng|webp|avif|gif|otf|ttf|woff|woff2)"]
             }
         }
-    ],
-    "dataGroups": [
-        {
-            "name": "api",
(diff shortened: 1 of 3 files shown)


luisa-beerboom (approved):

peb-adr (approved):
With the second commit this yields the expected behavior.

luisa-beerboom (approved):

Issue #1439, opened 2022-07-22
Permission revocation of navigation menu items requires forced reload or opening of another menu point to be active for users on their device


MSoeb:
If a user's permission to see a menu item is revoked, but the user has opened the menu, the content of the menu will still be displayed in the user's browser. Only the menu item disappears from the navigation bar. Only when the user changes the menu or reloads the page he has no further access.

Reproduction: 
1. Two Accounts are needed; Superadmin und normal user account. Normal should be group staff. 
1. open meeting with both accounts in different browsers
2. navigate with the normal user account to the participant menu
3. open as superadmin the participant menu. open the three-dot-menu and choose groups. revoke permission for group staff to see other participants
4. check the normal user view. menu is gone, but the user still sees the content.

What should happen: When the permission to see a menu item is revoked, there should be an automatic redirection to the first available menu item. Example: If available, then redirect to home page, otherwise redirect to next available menu item.

jsangmeister:
Goes in the same direction as #1421: I don't think this is important, since it's just visual and for a short moment. I don't think this situation happens often (if ever) in reality, since the permissions are not changed on the fly, and the amount of work required for this exceeds the benefit.

luisa-beerboom:
> Goes in the same direction as #1421: I don't think this is important, since it's just visual and for a short moment. I don't think this situation happens often (if ever) in reality, since the permissions are not changed on the fly, and the amount of work required for this exceeds the benefit.

I agree that this would be a bit of work, but I don't actually think it would be so much that it would exceed the benefits.
This would likely require a new service that watches the current permissions and checks if the user is still allowed on his current page whenever there is an update. Some of the checking and re-navigating logic could then be handeled by the auth-check.service and the reroute.service (I believe that these two do, in fact already contain large parts of the required logic, so the new service wouldn't have to be that big (we may also just skip the creation of a new service and just handle it all in the auth-check now that I think about it)).

luisa-beerboom:
#1553 adds a permissionObservable that always delivers the current permissions. This could be used to implement this functionality. 

It would also be a good idea to take a look at the various permission directives, I don't know if they currently update automatically

jsangmeister:
If it's not much work, we could move this into the 4.0 milestone and tackle this soon.

Issue #1421, opened 2022-07-20
when autopilot right is revoked, user continues to see updates in autopilot


MSoeb:
If a user's permission to use the autopilot is revoked during a meeting, but he remains in this menu, he will continue to see all updates projected into the autopilot. Only by manually leaving the autopilot he has no further access rights.

Reproduction:
Reproduction requires two users; an admin and another user in the Delegates group.

1. Both users must be logged into the meeting in differnent browsers.
2. delegate opens autopilot
3. admin projects a motion
4. admin removes delegate autopilot rights in group rights menu. 
5. Admin projects another motion. Delegate sees update in autopilot.

What should happen from my point of view: When a user is revoked the permission to see the autopilot, the user should be automatically navigated to the home page to not see any more updates.

jsangmeister:
Automatic redirection on permission loss is something we've never done, AFAIK. The same thing would be needed for all other views, e.g. motions, agenda, etc., when the respective rights are revoked.

The continuing autoupdates happend because the autopilot just wraps some other OS functionality which has its own access rights, e.g. the projector is controlled by `projector.can_see`. So it is logical that the user continues to see updates, since he is permitted to do so.

luisa-beerboom:
This is basically just a sub-problem of #1439 so I'll close this as a duplicate

PR #1553, opened 2022-09-01
Made livestream hide when perms are changed

### LINKED ISSUE #1552: Missing autoupdate for show/hide livestream window
* login as admin and delegate user, delegates cannot see live stream
* add livestream url (e.g. a youtube video url)
* activate permission "can see livestream" for delegates
* delegeate user has to reload to see the livestream window; same problem after deactivate livestream window

luisa-beerboom:
Closes #1552

### DIFF
--- a/client/src/app/site/services/operator.service.ts
+++ b/client/src/app/site/services/operator.service.ts
@@ -270,7 +270,6 @@ export class OperatorService {
             }
         });
         this.groupRepo.getGeneralViewModelObservable().subscribe(group => {
-            console.log(`Groups UPDATED!!!`);
             if (!this.activeMeetingId || !group) {
                 return;
             }
--- a/client/src/app/site/services/operator.service.ts
+++ b/client/src/app/site/services/operator.service.ts
@@ -137,6 +137,10 @@ export class OperatorService {
         return this._operatorReadySubject.asObservable();
     }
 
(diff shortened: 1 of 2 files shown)


client/src/app/site/services/operator.service.ts
```diff
@@ -258,6 +270,7 @@ export class OperatorService {
             }
         });
         this.groupRepo.getGeneralViewModelObservable().subscribe(group => {
+            console.log(`Groups UPDATED!!!`);
```
jsangmeister, line 273: remove

jsangmeister (changes requested):

jsangmeister:
https://github.com/OpenSlides/openslides-client/issues/1439 and https://github.com/OpenSlides/openslides-client/issues/1421 seem similar. Would it be possible to find a generic solution where the operator gets automatically updated when a permission changes and the page updates automatically accordingly?

emanuelschuetze (approved):

luisa-beerboom:
> #1439 and #1421 seem similar. Would it be possible to find a generic solution where the operator gets automatically updated when a permission changes and the page updates automatically accordingly?

You're right, and the permission-observable that is introduced in this PR should probably be used as a solution for those, but it should probably still be done in a different PR

jsangmeister (approved):

PR #1556, opened 2022-09-02
handle autoupdate via shared worker

bastianjoel:
resolves part of #1339

### TODO
- [x] connection closing
  - [x] do not close connections if used by another client
  - [x] close connections when client closed
- [x] connection management
  - [ ] _improve duplicate detection_
  - [x] improve health check
    - [x] display offline banner
  - [x] reconnect if connection lost
  - [x] health check
  - [x] error handling
  - [x] send data only to clients which requested it
- [x] stream numbering
- [x] history mode
- [x] handle connection closed by server
- [x] fix karma
- [x] do not resend all data on new chunk
- [x] add concept from #1454
  - [ ] _improve merging_
  - [ ] _add only requested fields to subscriptions_
  - [x] reconnect handling (after ending history mode)
- [x] ~~check if working with Service Worker~~ (using service worker slows down inital page load heavily)
- [x] add fallback
- [x] refactoring
- [x] JSDoc

The italic Todos are not needed for a first merge but would likely improve the performance and memory usage on client side. Therefore I will create new issues for these points after merging. 

### What needs be tested
- Usage with browsers that don't support Shared Workers (https://caniuse.com/sharedworkers)
- General functionality
- Production environment 
- Meetings with a lot of data
- History mode
- Error handling
  - Going offline (>10 sec and <10 sec)
  - Autoupdate unavailable/unhealty
  - Error from autoupdate
  - Authentication error


client/src/app/worker/autoupdate-stream-pool.ts
```diff
@@ -0,0 +1,243 @@
...
+                for (const port of subscription.ports) {
+                    if (ports.indexOf(port) === -1) {
+                        ports.push(port);
+                    }
+                }
+            }
+        }
+
+        for (const port of ports) {
+            port.postMessage({
```
luisa-beerboom, line 172: Just put this in the above loop as well

something like:
```
            ...
            ports.push(port)
            port.postMessage({...})
            ...
```

No need to iterate over everything _again_.

Also: The deeply nested for loops make this a bit unreadable, maybe create a helper function.

Also: If you'd prefer not to move postMessage into the big for loop, maybe you could turn the ports-Array into a set?
That way you could skip the explicit duplication check and just use `.add(port)`

client/src/app/worker/autoupdate-stream.ts
```diff
@@ -0,0 +1,259 @@
...
+            signal: this.abortCtrl.signal,
+            method: this.endpoint.method,
+            headers,
+            body: JSON.stringify(this.subscriptions.map(s => s.request))
+        });
+
+        const reader = response.body.getReader();
+        let next = null;
+        let result: ReadableStreamDefaultReadResult<Uint8Array>;
+        while (!(result = await reader.read()).done) {
```
luisa-beerboom, line 124: This is too long and too nested. Maybe create helper functions if it makes sense

luisa-beerboom (changes requested):
Some coding style things. I've said most of them only once, please take a look where else they apply. Especially the thing about the return values happens basically everywhere.

client/src/app/worker/sw-autoupdate.ts
```diff
@@ -0,0 +1,110 @@
...
+    context.addEventListener(`message`, e => {
+        const receiver = e.data?.receiver;
+        if (!receiver) {
+            return;
+        }
+
+        const msg = e.data?.msg;
+        const action = msg?.action;
+        const params = msg?.params;
+        if (receiver === `autoupdate`) {
```
luisa-beerboom, line 96: Use a switch statement here

luisa-beerboom (changes requested):

luisa-beerboom (approved):

bastianjoel:
Switching to draft because there is a authentication problem when opening new tabs

client/src/app/worker/autoupdate-stream.ts
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * Opens a new connection to autoupdate.
+     * Also this function registers this stream inside all subscriptions
+     * handled by this stream.
+     *
+     * resolves when fetch connection is closed
+     */
+    public async start(
+        force?: boolean
+    ): Promise<{ stopReason: 'error' | 'aborted' | 'unused' | 'resolved' | 'in-use'; error?: any }> {
```
luisa-beerboom, line 65: `'error' | 'aborted' | 'unused' | 'resolved' | 'in-use'` looks like you may turn it into a type or an enum or smth

luisa-beerboom (approved):
Looks good, just one minor remark

emanuelschuetze (approved):

PR #1565, opened 2022-09-05
Fix participant creation wizard

jsangmeister:
The `default_vote_weight` was mistakenly sent together with the user, which resulted in an error if the requets user did not have sufficient permissions.

@emanuelschuetze I could not reproduce the error of being able to change the account data.


emanuelschuetze (changes requested):
Works for normal accounts. But if I try to import a (super)admin account I get this error:
`"Your organization management level is not high enough to change a user with a Level of can_manage_users!"`

jsangmeister:
> Works for normal accounts. But if I try to import a (super)admin account I get this error: `"Your organization management level is not high enough to change a user with a Level of can_manage_users!"`

Currently working as intended, see https://github.com/OpenSlides/OpenSlides/wiki/user.update#permissions. I will change it accordingly in the backend

emanuelschuetze (approved):

Issue #1525, opened 2022-08-24
Function "Invite to conference room" didn't work


MSoeb:
Problem description:
When a user who is on the list of speakers is invited to the conference room by using the "Invite to conference room" function, nothing happens. Neither the admin nor the user sees anything in the browser. Only the console gives an error message to the admin when the button is clicked. 

Reproduction:
1. Setup Jitsi for a meeting:
1.1 Open committees on organisation-level and choose a committee
1.2 choose a meeting, click on the three-dot-menu, select edit
1.3. go to superadmin setting and fill out the lines jitsi domain and jitsi room name (if you need data, please ask @m-schieder ). save everything.
2. now open the meeting
3. open settings: check if "Show live conference window" is activated. If not, do so.
4. open participants: create a new user, add it to a group staff.
5. go to motions: create a motion
6. open the motion menu. click on the camera in front of the created motion and select project. Now the motion with a list of speakers appears in the autopilot.
7. open the Autopilot and add the user to the list of speakers.
8. open an incognito tab, open the instance page and log in as the created user. open console there.
9. head back to the admin instance page: open console. 
10. open the autopilot. click on the three-dot-menu in the user line and choose "Invite the conference room". Nothing happens.
11. Error message appears in console. 400 error
12. check the icognito tab double check if an action happens

Error message: 
![Screenshot_20220824_160143](https://user-images.githubusercontent.com/105491030/186438565-e0f65f1f-d636-424b-814f-36707e3f8768.png)

What should happen:
From my point of view, it is not clear, what should happen. I can assume three options:
1. User gets forced to join the room and it opens automatically in the client browser. May be, this is to harsh.
2. A request box appears which ask the user to join the room. So the user could decide what to do. I would prefer this option. 
3. The function is obsolte can be deleted. 

@emanuelschuetze What should be from your point of view the correct way to handle this?

luisa-beerboom:
Can you show the exact error message, or add a picture?

MSoeb:
> Can you show the exact error message, or add a picture?

Ah yes, forgot it. Added it to post.

MSoeb:
some more infomation about the error:

![Screenshot_20220825_090853](https://user-images.githubusercontent.com/105491030/186598597-732e4c94-0e77-4cd4-805f-1f978b9ece47.png)

![Screenshot_20220825_090622](https://user-images.githubusercontent.com/105491030/186598355-b112340b-646e-4ebb-ae81-759439f9d28a.png)

[...]

PR #1573, opened 2022-09-06
Fix the problem when joining the Jitsi room

m-schieder:
* Thanks to @YveIce for the fix :)
 * Update Jitsi API


rrenkert (approved):

emanuelschuetze (changes requested):
Invite as admin another user. User (opens in private tab) gets dialog and two TypeErrors:
```
TypeError: Cannot read properties of undefined (reading 'nativeElement')
[..]
TypeError: Cannot read properties of undefined (reading 'nativeElement')
    at RtcService.setOptions (rtc.service.ts:384:40)
```

m-schieder:
Looks like two dialogs are created to enter the conference.
The first dialog is broken. When the first dialog is closed, the second one appears. This one works.
I will investigate the behavior later.

m-schieder:
I can no longer reproduce this error. Also not in a productive environment. @emanuelschuetze can you test this again? Maybe you can still reproduce it.

emanuelschuetze (changes requested):
I can still reproduce the error:
* use a private tab and log in as delegate user
* add delgate user to current list of speakers
* as admin: invite user to jitsi
* delgate user gets dialog -> user has to click 2(!) time on OK button

m-schieder:
This PR fixes a general issue with entering a Jitsi room. I note the other issues with inviting people into the conference room in the associated [issue](https://github.com/OpenSlides/openslides-client/issues/1525) and merge this PR.

PR #1579, opened 2022-09-07
Readd datagroup in ngsw-config and cleanup code

### LINKED ISSUE #1568: Mediafiles: Files are notdisplayed
`…/…/mediafiles`

### Problem:
If you click on a file (under `Files` or on the link of an attachment under `Agenda` `Motions` or `Assignments`) the file/URL to the file is opened in a new tab. So far, so good.

But the file is not displayed. The URL `…/system/media/get/…` is not followed, but there is a redirect to the organization level ("Calendar").

### Desired behavior:
When the link of a file is clicked, then the file is opened/displayed properly (as it was until a few days ago) in a new tab.

jsangmeister:
fixes #1568

### DIFF
--- a/client/src/environments/environment.prod.ts
+++ b/client/src/environments/environment.prod.ts
@@ -1,6 +1,5 @@
 export const environment = {
     production: true,
-    urlPrefix: `/apps`,
     authUrlPrefix: `system/auth`,
     authSecurePrefix: `secure`
 };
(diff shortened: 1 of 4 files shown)


jsangmeister:
Works now again in test instance. @peb-adr seems like the datagroups had another use after all, although I'm not quite sure why :D

client/src/environments/environment.prod.ts
```diff
@@ -1,6 +1,5 @@
 export const environment = {
     production: true,
-    urlPrefix: `/apps`,
```
jsangmeister, line 3: It does not seem like this variable is used anywhere, so I removed it.

emanuelschuetze (approved):

PR #1581, opened 2022-09-08
keep selection in scroll tables if data filtered

bastianjoel:


emanuelschuetze (approved):

rrenkert:
Ok for now, but seems like a workarround to keep the selection manually. We should have a look where the real problem comes from.

rrenkert (approved):

PR #1584, opened 2022-09-12
Motion list forwarding filters

luisa-beerboom:
Closes #1577

PR #1601, opened 2022-09-15
Cleaned up forwarding filter

luisa-beerboom:
See comment in #1584


client/src/app/site/pages/meetings/pages/motions/view-models/view-motion.ts
```diff
@@ -222,7 +222,7 @@ export class ViewMotion extends BaseProjectableViewModel<Motion> {
 
     public get forwardingStatus(): ForwardingStatus {
         let status = ForwardingStatus.none;
-        if (!!this.all_origin_ids?.length) {
```
jsangmeister, line 225: I suspect you have to change the motion model subscription as well, or is the origin id already included there?

jsangmeister (changes requested):

emanuelschuetze (approved):
works

PR #1602, opened 2022-09-15
fix BaseModelRequestHandlerComponent onParamsChanged not called

### LINKED ISSUE #1592: Motion forwarding needs reload to show timestamp
* create forward setup (committee A should forward to B)
* create motion A1 in meeting of committee A
* forward A1 to meeting of B
=> message "successfully forwared" is shown BUT NO timestamp information in meta column in motion detail view. You have to reload to see this information. Should work without reload.

bastianjoel:
resolves #1592

### DIFF
--- a/client/src/app/site/base/base-model-request-handler.component/base-model-request-handler.component.ts
+++ b/client/src/app/site/base/base-model-request-handler.component/base-model-request-handler.component.ts
@@ -35,7 +35,7 @@ export class BaseModelRequestHandlerComponent extends BaseUiComponent implements
     private _destroyed = new EventEmitter<boolean>();
     private _openedSubscriptions: string[] = [];
     private _currentMeetingId: Id | null | undefined = undefined; //This is to ensure that the if-check in ngOnInit also fires if the application isn't in a meeting
-    private _currentParams: string | undefined = undefined;
+    private _currentParams: any = {};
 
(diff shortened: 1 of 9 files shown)


luisa-beerboom (changes requested):
`onParamsChanged` is a function that is basically only used to open subscriptions when meetings are changed. Expanding it to fire basically whenever there is a new route is therefore probably redundant, especially since you're just seeking to fix one problem, in one single place. Is there no other solution at a lower level that doesn't require the client to try and open new subscriptions on every page change?
Perhaps you're not subscribed to something, or the component watches changes within the viewed motion and therefore, while it will notice changes in the array of forwarded motions, it will not notice anything that goes on within these motions (i.e. new timestamps)

bastianjoel:
@luisa-beerboom I have looked through the overrides of `onParamsChanged` and saw that most of those methods seem to assume that this is fired whenever a parameter is changed. 
Also there is a `onNextMeetingId` method that should be used to listen for meeting changes. 

I will update the places where `onParamsChanged` is used instead of `onNextMeetingId` and add a parameter to the `onParamsChanged` method that contains the old parameter values.

emanuelschuetze (approved):

luisa-beerboom (approved):
Look good, not tested

PR #1603, opened 2022-09-15
PDF export, Let texts have multiple decoration

### LINKED ISSUE #1385: PDF Export,Texts can have only one decoration
If an amendment contains a link, the link will not be shown crossed out in the pdf export of the motion.

Repro:
1. Create motion
2. create an amendment with a link inside the text, NOT at the line start.
3. switch back to the motion.
4. open the 3-dot-menu. choose pdf. 
5. open pdf file. -> link is not crossed out. 

What should happen: The link should be crossed out even in the pdf export file. 

Example image: OS <--> pdf file
Screenshot_20220705_160316

m-schieder:
Fix #1385


luisa-beerboom (approved):
Looks good to me, not tested though

emanuelschuetze (approved):

PR #1621, opened 2022-09-26
improve cypress tests

bastianjoel:
closes #1476

applies the following changes to the cypress tests
- update cypress version
- disables videos by default
- allows retries in run mode
- improves startup time of `make run-cypress`
- cleanup of test data after login test


jsangmeister (approved):
Looks good! Could you check the tests deleted inhttps://github.com/OpenSlides/OpenSlides/pull/6486 from the main repo and transfer the ones which do not exist in the client repo, so we can delete the tests from the main repo?

jsangmeister (approved):

PR #1622, opened 2022-09-27
Possible fix for safari sidebar arrow bug

### LINKED ISSUE #1281: Sidebar-arrow gets stuck to wrong position on apple-devices
Open Safari
-> open sidebar (by using the arrow (>))
-> close sidebar (same button)
=> the (>)-Button gets stuck to the position it would have, if the sidebar was open, instead of moving back to the left side of the page.
Navigating to other pages does not return the button to the correct position.

Some views (for example the projector view) do work as intended, however most other pages seem to have this error.

luisa-beerboom:
Closes #1281 

Had an idea what might be causing this bug and changed the corresponding part of the code.
I don't actually have a way to test this because I don't have access to an apple device, so that will have to be done by someone else.

### DIFF
--- a/client/src/app/ui/modules/sidenav/components/sidenav/sidenav.component.html
+++ b/client/src/app/ui/modules/sidenav/components/sidenav/sidenav.component.html
@@ -63,20 +63,12 @@
     <!-- Toggle-button -->
     <div class="nav-toggle-button-container" *ngIf="!isMobile">
         <button
-            *ngIf="isOpen"
             mat-raised-button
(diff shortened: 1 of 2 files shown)


emanuelschuetze (changes requested):
does not work

emanuelschuetze (changes requested):
same problem

MSoeb:
I would take myself out for the review. Currently no pad and time.

emanuelschuetze:
Closed. Already fixed with #1641

PR #1641, opened 2022-10-03
Fix navbar arrow position on WebKit

### LINKED ISSUE #1281: Sidebar-arrow gets stuck to wrong position on apple-devices

tsiegleauq:
Fixes a bug where the "close" arrow was not repositioned on WebKit browsers Closes #1281


emanuelschuetze (approved):
Tested. Thank you, Sean!

PR #1642, opened 2022-10-04
fix none selection of agenda parent item

### LINKED ISSUE #1636: Selecting parent agenda item "-" results in error
When creating a new topic selecting the "-" option as parent agenda item results in the following error:
"Error: data.agenda_parent_id must be bigger than or equal to 1"

bastianjoel:
resolves #1636


luisa-beerboom (approved):

emanuelschuetze (approved):
.

PR #1660, opened 2022-10-05
[For testing] Added cryptographic option to poll creation

luisa-beerboom:
@ostcar with this you can create cryptographic polls from the client

For now this will only work for motion & assignment polls


ostcar:
@luisa-beerboom can this be closed?

luisa-beerboom:
@ostcar yes

PR #1662, opened 2022-10-05
cache has voted requests in vote repository

### LINKED ISSUE #1633: Investigate why the client makes so many `voted`-requests to the vote service
Whenever the page is opened, the client sends a bunch of repetitive `voted`-requests, even requesting the same exact data twice.

- [ ] Investigate when this happens and why this happens.
- [ ] If it can be fixed, fix it.

There is a danger of clients spamming a massive amount of unnecessary requests to the vote-service if there are a lot of polls in an instance.

bastianjoel:
This resolves #1633 by adding a cache and request bundling mechanism to the VoteRepositoryService.

Requested ids from the voted endpoint get cached for 1 second and not cached ids will be collected for 50ms before they will be requested together.


luisa-beerboom (approved):
Looks fine, but could you maybe look into just reducing unnecessary calls to `VoteControllerService.hasVotedFor`?

That one is the only function that calls this method, and it doesn't exactly return anything. All it does is extract the relevant data from the response and write it into the model.
If it wasn't constantly called with the exact same values, it would probably be better

client/src/app/gateways/repositories/polls/vote-repository.service/vote-repository.service.ts
```diff
@@ -42,9 +54,57 @@ export class VoteRepositoryService extends BaseMeetingRelatedRepository<ViewVote
...
 
     public async hasVotedFor(...ids: Id[]): Promise<HasVotedResponse | undefined> {
-        if (ids.length) {
-            return await this.http.get(`${HAS_VOTED_URL}?ids=${ids.join()}`);
+        if (!ids.length) {
+            return undefined;
+        }
+
+        const result = this.getIdsFromCache(ids);
+        if (ids.some(id => !!result[id])) {
```
jsangmeister, line 62:
```suggestion
        if (ids.every(id => !!result[id])) {
```
or else you would return potentially incomplete results

jsangmeister (changes requested):

emanuelschuetze (approved):

PR #1668, opened 2022-10-06
Fix icc service communication problem and invite to conference function

### LINKED ISSUE #1525: Function "Invite to conference room" didn't work

luisa-beerboom:
Closes #1525 

Already got the ICC service gateways to function correctly again (They previously wouldn't open streams, inhibiting any Notify- or ApplauseService activities)

TODO: 

- [x] Fix invite to conference function

### DIFF
--- a/client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
+++ b/client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
@@ -166,6 +166,7 @@ export class InteractionService {
 
     private async onCallInvite(): Promise<void> {
         if (!this.isInCall) {
+            console.log(`ON CALL INVITE`)
             const accept = await this.promptService.open(CallInviteTitle);
 
             if (accept) {
--- a/client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
+++ b/client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
@@ -1,23 +1,14 @@
 import { Injectable } from '@angular/core';
-import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
-import {
-    BehaviorSubject,
-    combineLatest,
-    distinctUntilChanged,
-    filter,
-    firstValueFrom,
-    map,
-    merge,
-    Observable
-} from 'rxjs';
+import { map, Observable } from 'rxjs';
 import { Id } from 'src/app/domain/definitions/key-types';
-import { NotifyResponse, NotifyService } from 'src/app/gateways/notify.service';
+import { NotifyService } from 'src/app/gateways/notify.service';
 import { OperatorService } from 'src/app/site/services/operator.service';
 import { PromptService } from 'src/app/ui/modules/prompt-dialog';
 
 import { ViewUser } from '../../../view-models/view-user';
 import { BroadcastService } from './broadcast.service';
 import { CallRestrictionService } from './call-restriction.service';
+import { InteractionReceiveService } from './interaction-receive.service';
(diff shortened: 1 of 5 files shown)


client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
```diff
@@ -78,6 +80,8 @@ export class InteractionService {
         private promptService: PromptService,
         private broadcast: BroadcastService
     ) {
+        InteractionServiceRegistry = InteractionServiceRegistry.concat(this);
```
jsangmeister, line 83: Mmh, I don't particularily like this solution (it seems like the underlying problem is somewhere in the DI and this is just symptom fixing), but it's fine by me if it works.

client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
```diff
@@ -179,4 +183,15 @@ export class InteractionService {
             });
         }
     }
+
+    private isFirstInteractionService(): boolean {
+        for (let i = 0; i < InteractionServiceRegistry.length; i++) {
+            if (InteractionServiceRegistry[i] === this) {
+                break;
+            } else if (!!InteractionServiceRegistry[i]) {
```
jsangmeister, line 191: When would this ever happen? `this` is the only thing added and that can never be `null`, right?

jsangmeister (changes requested):

client/src/app/site/pages/meetings/pages/interaction/services/interaction.service.ts
luisa-beerboom: This problem indeed lies with the DI. Mainly it is because this service is not provided in `root` and a new instance of it is therefore created whenever it is injected into a new module. I _tried_ fixing this by changing where it is provided but failed due to its dependencies. I _could_ try to create a new `root` service for the purpose of handling received messages, but there is no guarantee that this would work either.

jsangmeister (approved):
looks good!

emanuelschuetze (approved):

PR #1678, opened 2022-10-11
Fixed bug where mediafiles couldn't be attached to motions

### LINKED ISSUE #1677: Can't attach mediafiles to motions
1. Upload a mediafile to the meeting
2. Create a new motion
3. Assign the mediafile to the motion
4. reload -> The mediafile will not be part of the motion anymore

luisa-beerboom:
Closes #1677

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-view/motion-detail-view.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-view/motion-detail-view.component.ts
@@ -74,6 +74,8 @@ export class MotionDetailViewComponent extends BaseMeetingComponent implements O
         return this._motion;
     }
 
+    private originalMotion: ViewMotion;
+
     public temporaryMotion: any = {};
 
     public canSave = false;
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-view/motion-detail-view.component.ts
```diff
@@ -221,6 +223,7 @@ export class MotionDetailViewComponent extends BaseMeetingComponent implements O
 
     public enterEditMotion(): void {
         this.editMotion = true;
+        this.originalMotion = JSON.parse(JSON.stringify(this.motion));
```
jsangmeister, line 226: This always seems hacky... What is the root problem? Is there another way to solve this?

jsangmeister (changes requested):

emanuelschuetze (approved):
Tested. Works for me.

emanuelschuetze:
I'll merge it now.
@jsangmeister if you have a better solution please open a new issue.

jsangmeister:
I don't have a better solution, I'd have to dig deeper into this as well. The problem I see with merging such "hacks" is that the software degrades over time as more and more of these are added, which multiplies the problem because with more hacks it becomes more and more difficult to find the true root of such bugs and leads to evens more hacks.

PR #1684, opened 2022-10-12
prevent opening http streams while cypress is running

bastianjoel:
Cypress tests are failing again after an issue with opening icc connections was resolved. 

For that reason I have added a check that prevents opening http streams when the client is used by cypress.


luisa-beerboom (approved):
Not very pretty, but evidently it works (for now).

Since this will torpedo any attempts to test http-stream-reliant functionalities (f.E. anything to do with the ICC), we'll have to make sure to change it back as soon as Cypress can properly work with streams, though.

rrenkert:
I completely agree with @luisa-beerboom. We should not forget the TODOs!

rrenkert (approved):

PR #1695, opened 2022-10-13
use radio buttons for motion analog vote majority selection

### LINKED ISSUE #1469: analog voting: bug - multiple Majorities can be selected simultaneously
Problem description:
It is possible to select different majorities at the same time as the result in analog elections. This is not possible in a vote.

Reproduction:
1. log in a meeting as admin/superadmin.
2. open menu motions
3. create a new motion with the plus symbol.
4. Create a vote in the motion. (if necessary, adjust the status until the vote button appears).
5. in selection mask select as voting type analog. Then check the box for majorities for yes, no and abstain. 

What should happen:
If one of the check marks is set, no further check marks can be set. The check marks should be grayed out. Different majorities cannot exist at the same time.

Screenshot_20220803_144418

bastianjoel:
resolves #1469 for motion votes.


luisa-beerboom (approved):

emanuelschuetze (changes requested):
cannot deselect a selected radio button. I think a (disabled) checkbox is a better component instead radio button which cannot deselect.

luisa-beerboom (approved):

emanuelschuetze (approved):
works now as expected

PR #1699, opened 2022-10-14
Grey out motion detail nav buttons

### LINKED ISSUE #1698: Improve motion navigation
When navigating between motion detail pages and reaching the first/last Element, the previous/next motion button should be greyed out, instead of vanishing

luisa-beerboom:
Closes #1698 

TODO: Find out what shall be done with the amendment navigation (when on a motion that has amendments, the next-motion-button will lead to the first of those amendments. There is no way to reach the next non-amendment motion, aside from navigating to the motion-list and manually selecting it)

What do you think @emanuelschuetze ?


luisa-beerboom:
Extracted the amendment navigation bug to issue #1704

bastianjoel (approved):

emanuelschuetze (changes requested):
min-width would be helpful

MSoeb (approved):

emanuelschuetze (approved):

PR #1700, opened 2022-10-14
Solved a few problems with the account search selector

### LINKED ISSUE #1502: Problems with user search mask in meeting creation/edit
When trying to use the search selector for meeting admins in the meeting edit view, I found two different problems:

1. Trying to search for another user through the user search selector for meeting admins, while being logged in as a committee manager without OML.user_can_manage throws massive amounts of missing permission errors.
   => The search selector should probably check whether the operator has the correct permissions and not even try to search otherwise, only allowing the operator to select himself.
2. If (in dev mode) you
   1. As an orga admin use the user search selector for meeting managers and
   2. Search for the superadmin (i.e. 'admin')
   3. You'll see that his name is displayed as `null Administrator`
   => The search selector seems to be somewhat unable to correctly display users without a first name under these circumstances. (Which is strange because the superadmin himself sees the correct name)

luisa-beerboom:
Closes #1502

### DIFF
--- a/client/src/app/site/pages/organization/modules/account-search-selector/components/account-search-selector/account-search-selector.component.ts
+++ b/client/src/app/site/pages/organization/modules/account-search-selector/components/account-search-selector/account-search-selector.component.ts
@@ -58,14 +58,15 @@ export class AccountSearchSelectorComponent extends BaseSearchSelectorComponent 
 
     private async searchAccount(name: string): Promise<void> {
         const user = this.userController.parseStringIntoUser(name);
-        if (this.operator.hasOrganizationPermissions(OML.can_manage_users)){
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

emanuelschuetze (changes requested):
Now, as committee admin, I cannot find/add any account in meeting admin search field...

luisa-beerboom:
> Now, as committee admin, I cannot find/add any account in meeting admin search field...

@emanuelschuetze This is unsurprising, since pure committee admins lack the rights required by the presenter, which is why I disabled the search function for them. This would therefore require backend changes

emanuelschuetze (approved):

PR #1703, opened 2022-10-17
Extend history to users and assignments

jsangmeister:


luisa-beerboom (approved):

emanuelschuetze (changes requested):
this needs some discussion and rework

emanuelschuetze:
backend PR: https://github.com/OpenSlides/openslides-backend/pull/1478

jsangmeister:
Changes:
- add history to menus in assignment and participant detail views
- fqids in history entries are replaces by the model's title
- wording of the history selection changed

TODO:
- to replace the fqids, the local model store is used. If the model was not loaded before, this will not work. We need to implement a way to dynamically load the models we need
- for some collections (e.g. meeting), the replaced name should be a link to the respective model for easier access

emanuelschuetze:
I get a strange issue in my test: I opened the detail view of an existing user -> history. The history table contains wrong motion stuff (motion created, poll stopped, ...) - This is NOT a user history. See screenshot.
user-history

emanuelschuetze:
user history seems to be not working - see also openslides-backend/pull/1478

luisa-beerboom (approved):

emanuelschuetze (approved):
Tested.

emanuelschuetze:
One remark for motion history: Please use recommendation label instead of state label if recommendation is set.
see https://github.com/OpenSlides/openslides-client/issues/1855

PR #1710, opened 2022-10-19
fix single candidate YN 100% base

### LINKED ISSUE #1696: Elections/votes: Incorrect result graphic and calculation
`…/…/assignments/…`

OpenSlides Version: 4.0.0-beta-20221012-1e75339

### Bug description
- If "yes/no per candidate" is set as 100% base, the graphic still shows the abstentions.
- For the yes and no votes, it says 100% each, which is wrong. There seems to be no calculation.

### Expected behavior
- If 100% base is set as "yes/no per candidate, then the graphic shows only the yes and no votes.
- The calculation for the yes and no votes is correct.

### Screenshots
Bildschirmfoto 2022-10-13 um 16 44 35

bastianjoel:
resolves #1696

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/chart/chart.component.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/components/chart/chart.component.ts
@@ -157,7 +157,7 @@ export class ChartComponent {
                 // removes undefined and null values
                 return chartDate.data.filter(data => !!data);
             });
-            if (!this.circleColors) {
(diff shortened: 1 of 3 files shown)


emanuelschuetze (changes requested):
* Create YNA-assignment poll for one candidate with YN-100%base
* Go to home page -> Reload -> go back to assignment detail
* Now switch 100% base to YNA
=> abstain section in chart is green NOT grey!

emanuelschuetze (approved):

PR #1715, opened 2022-10-19
info if http stream got first data

### LINKED ISSUE #1238: Implement the info if an HTTP stream has got its first data
Sometimes, it is necessary to know, if there are already data came from one HTTP stream (from the AU-service) to handle empty lists or anything else.

This could improve the following views:
- Motion list: The client could determine if the tile view is shown after it knows if there are categories
- The sequential number mapping should be obsolete: We could await until the list of models has loaded
- Detail views of some models (motion, assignment, ...): We could await until the data for a specific model is loaded to show its detail view or to show the information, that a model is not found

bastianjoel:
resolves #1238 

I'm currently using this only for some detail views. While loading sequential numbers this prevents displaying a not found message.
To test this change https://github.com/OpenSlides/openslides-client/blob/f11409a8421959e68018043126bbdfec7e99674a/client/src/app/site/services/autoupdate/autoupdate.service.ts#L110 to
```ts
setTimeout(() => {
    this.handleAutoupdate({ autoupdateData: data.data, id: data.streamId, description: data.description });
}, data.description.indexOf(`prepare`) !== -1 ? 5000 : 0);
```


bastianjoel:
TODO: 

- [ ] add info when loading
- [ ] display headbar

emanuelschuetze (approved):

bastianjoel:
Do not merge yet. There seems to be a problem with not found pages.

emanuelschuetze (approved):

luisa-beerboom (approved):

PR #1720, opened 2022-10-20
recheck current route guards on permission change

### LINKED ISSUE #1439: Permission revocation of navigation menu items requires forced reload or opening of another menu point to be active for users on their device

bastianjoel:
resolves #1439


luisa-beerboom (approved):

luisa-beerboom (approved):

client/src/app/site/guards/permission.guard.ts
```diff
@@ -13,26 +14,25 @@ export class PermissionGuard implements CanLoad {
...
     ) {}
 
-    public async canLoad(route: Route, segments: UrlSegment[]): Promise<boolean> {
+    public async canLoad(route: Route, segments: UrlSegment[]): Promise<boolean | UrlTree> {
         const url = this.getCurrentNavigationUrl();
         if (this.osRouter.isOrganizationUrl(url)) {
             if (!(await this.authCheck.isAuthorizedToSeeOrganization())) {
-                this.reroute.forwardToOnlyMeeting(url === `/info` ? [`info`] : []);
-                return false;
+                const meetingId = this.operator.onlyMeeting.toString();
```
luisa-beerboom, line 25: Why not continue using forwardToOnlyMeeting here?
Haven't you changed it to do exactly what you are also doing here?

client/src/app/site/services/reroute.service.ts
```diff
@@ -63,16 +65,16 @@ export class RerouteService {
     /**
      * If the user requested a route without direct meaning, forward to their meaningful meeting
      */
-    public forwardToOnlyMeeting(segments: string[] = []): void {
+    public forwardToOnlyMeeting(segments: string[] = []): UrlTree | void {
```
luisa-beerboom, line 68: With your changes to take out the navigation most of the `RerouteService`'s method names don't fit anymore, please change names like `forwardToOnlyMeeting` that insinuate that they also handle the navigation to something more fitting like maybe `getOnlyMeetingUrlTree` (please don't forget to change the documentation either)

luisa-beerboom (changes requested):

luisa-beerboom (approved):

emanuelschuetze (changes requested):
Go as normal user to e.g. agenda list.
Now remove agenda.can_see permission.
-> User gets Authorization Error for agenda_item.can_see (OK)
Now click on "Back to OpenSlides" button which shows again a 2nd Authorization Error (for meeting.can_see_frontpage).
User has to click 2-3 times on "Back to OpenSlides" to go back to frontpage.

bastianjoel:
I can't replicate this. Anyway I've changed the way the error page redirects and hope this fixes the issue.

emanuelschuetze (changes requested):
same problem, reproduceable for me

bastianjoel:
I can reproduce this now. Happens only in chromium.

luisa-beerboom (approved):

luisa-beerboom (approved):

emanuelschuetze (changes requested):
Retested: Same error.
Get "Error: Uncaught (in promise): [object Undefined]" after clicking 2-3 times on "Back to OpenSlides"
Tested with Chromium and Firefox.

emanuelschuetze (approved):
Looks good now!

PR #1740, opened 2022-10-25
Do not expand motion text in amendment diff view

### LINKED ISSUE #1721: Do not expand motion text in amendment when adding change request
- create motion with multiple lines of text
- create amendment for that motion
- make a change recommendation inside the amendment

**Current behavior**
- Diff version: The whole motion text is displayed inside the amendment.

**What should happen**
Only the text that was displayed before adding the change recommendation should be displayed.

bastianjoel:
resolves #1721


luisa-beerboom (approved):

rrenkert (approved):

emanuelschuetze (commented):
Tested

PR #1757, opened 2022-10-27
Export also under points in the nested-lists

### LINKED ISSUE #1161: Also export the subitems in the nested lists in the PDF export.
If a motion contains a nested list, the subitems are not exported in the PDF. 
However, these should be exported as well.

m-schieder:
Fix  #1161


luisa-beerboom (approved):
Looks good, not tested

emanuelschuetze (approved):

PR #1761, opened 2022-10-28
Use playwright instead of cypress

bastianjoel:


bastianjoel:
Do not merge yet. 

I want to check if I can enable the parallel option when #1423 got merged.

rrenkert (approved):

PR #1771, opened 2022-11-03
hide finished speakers from projector

### LINKED ISSUE #1614: List of speakers projection problems
* Create new Agenda item
* Create new projector for list of speakers
* Project list of speakers

Result: List of speakers is empty; Subtilte contains "0 speakers"
Expected: Show speaker count only if >0

* Add Speaker
* Start and stop 

Result: The speaker remains in the list even though the setting "Number of last speakers to be shown on the projector" ist set to 0.
Expected: List should be empty.

bastianjoel:
resolves #1614


luisa-beerboom (changes requested):
What happens if "Number of last speakers to be shown on the projector" is set to a number bigger than 0?
Seems to me like in that case the finished speakers should be shown.
Perhaps, instead of deleting the finished speakers, that value should be requested and checked.

PR #1781, opened 2022-11-08
fix mediafile scrolling table after hide

### LINKED ISSUE #1768: media files: upload-UI problems
**Description:** The file upload has UI problems. If a user selects a file, delete it and choose another one, the UI doesn't display the upload menu correct anymore. Two problems occur:
- [ ] select the same file again for upload -> upload details not visible anymore
- [ ] select another file -> broken UI appears. 

This problem occurs on the organisation and meeting level. Both file menus are affected. 

Normal UI:
Screenshot_20221102_130057

UI first problem
Screenshot_20221102_130344

UI second problem:
Screenshot_20221102_130326

**Reproduction:**
1. open an instance
2. open the file menu
3. try to upload a new file via plus-icon
4. deleted the selected file and add it again as selection (first case)
5. try to add another file (second case)

**Expected behavior:** The user should be able to delete selected file without getting a broken UI for all follow-up selections.

bastianjoel:
resolves #1768 

Currently when deleting all files from from file upload list the scrolling table gets destroyed. This also deletes the previously defined columns. With this pr the scrolling table only gets hidden.

### DIFF
--- a/client/src/app/ui/modules/file-upload/components/file-upload/file-upload.component.html
+++ b/client/src/app/ui/modules/file-upload/components/file-upload/file-upload.component.html
@@ -15,7 +15,7 @@
         <!-- Additional content -->
         <ng-container *ngTemplateOutlet="additionalContent"></ng-container>
     </div>
-    <div class="upload-area--bottom-content">
+    <div [hidden]="isEmpty" class="upload-area--bottom-content">
         <ng-container *ngTemplateOutlet="tableTemplate"></ng-container>
     </div>
 </div>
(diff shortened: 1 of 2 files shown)


luisa-beerboom (approved):

emanuelschuetze (changes requested):
* works: select another file
* does not work: select the same file again for upload -> upload details not visible anymore

luisa-beerboom (approved):

emanuelschuetze (approved):

PR #1785, opened 2022-11-09
Projector can now show pdfs again

### LINKED ISSUE #1773: Projection of pdf file not possible
**Description:** In OS3 it was possible to project a pdf file in for example the autopilot. This doesn't work currently in OS4. The projector is empty. 

**Reproduction:**
1. open a meeting
2. open files
3. Upload a pdf file
4. fiel menu should be opened after upload: click now on the three-dot-menu behind the file and choose project. Choose now "default projector".
5. open now the Autopilot and check, if the pdf file is project. Addtionally check it in the projector menu. -> file isn't displayed anywhere.

**Expected behavior:** A pdf file should be projectable on any available projector like in OS3.

luisa-beerboom:
Closes #1773


emanuelschuetze (changes requested):
* please show pdf in fullscreen to prevent scrollbars
* how can you control the pdf pages (next/previous page, zoom)? seems to be missing in projector detail view (mediafile section in drop down) -> see OS3 demo
* if you project a image show it always in fullscreen (use class 'fullscreen' instead 'nofullscreen')

luisa-beerboom:
>     * how can you control the pdf pages (next/previous page, zoom)? seems to be missing in projector detail view (mediafile section in drop down) -> see OS3 demo

Seems to me like that currently isn't implemented. I'll create a new issue for that

emanuelschuetze (changes requested):
With chromium there are still scroll bars after projecting a 16:9 pdf on 16:9 projector

client/package.json
```diff
@@ -119,6 +119,7 @@
     "po2json": "^1.0.0-beta-2",
     "prettier": "^2.7.1",
     "ts-node": "^10.7.0",
+    "tsconfig-paths": "^4.1.0",
```
bastianjoel, line 122: I'm not sure if this should be included in this PR

bastianjoel (approved):
`client/src/assets/pdfworker/pdf.worker.min.js` should not be included in the project because updating the pdf viewer library without updating that file would break the projector pdfs again. 
I think this is ok for now but we should create an issue to change that later.

client/package.json
luisa-beerboom: Probably not, I'll remove it

bastianjoel: `client/package-lock.json` needs to be reset too then.

emanuelschuetze (approved):

bastianjoel (approved):

PR #1012, opened 2022-04-23
Bugfix/diff line numbers should refer to whole lines

CatoTH:

PR #1794, opened 2022-11-12
Fix line number collision detection, missing annotations

CatoTH:
This fixes the collision detection, which probably broke as part of https://github.com/OpenSlides/openslides-client/pull/1012 . An example where no collision is was discovered anymore:
- Amendment 1 covers line 5 (lineFrom = 5, lineTo = 5)
- Amendment 2 covers line 5 (lineFrom = 5, lineTo = 5) too

(The failing test appears to be a general issue in the pipeline, as it also happens in https://github.com/OpenSlides/openslides-client/actions/runs/3446108245/jobs/5750608572)


emanuelschuetze:
The collision detection works now as expected/OS3. But I found annother linenumber bug: If you create a new amendment the wizard shows in step 1 the main motion paragraphs with wrong line numbers (e.g. "line 1-3" for first paragraph, but line 1-4 would be correct) -> the "to" line number needs +1.

CatoTH:
@emanuelschuetze the second commit should fix this, too.

emanuelschuetze (approved):

PR #1797, opened 2022-11-14
Pdf projection: Can now switch pages and scroll again

### LINKED ISSUE #1786: Pdf projection: Make it possible to switch pages
PR #1785 will fix a bug where pdfs couldn't be projected, but this leaves a problem where apparently, once a pdf is projected, it is currently not possible to switch pages or zoom.

The `mediafile-slide.component` in it's current form does allow zooming and page-setting, but the current zoom level and the current page have to be saved in the backend and then sent in the slide options.

Please implement a method for setting the current page and the zoom analog to what happens in OS3.

luisa-beerboom:
Closes #1786


emanuelschuetze (approved):
Works good. - One minor remark: Please switch pdf zoom-in and zoom-out buttons in mediafile control panel (icon out should be left, icon in should be right - as projektor zoom buttons)

bastianjoel (approved):

emanuelschuetze (approved):

Issue #1172, opened 2022-05-20
Support for DIN SPEC 91379 in PDF prints


alveolus:
Some letters and characters are not printed in PDF files. These are characters that should be processed according to DIN SPEC 91379 ("Characters in Unicode for the electronic processing of names and data exchange in Europe"). These letters and characters are used especially in names whose spellings are variations of basic Latin. Like the letter `Ş` in my surname. ;-)

[DIN SPEC 91379.pdf](https://github.com/OpenSlides/openslides-client/files/8745787/DIN.SPEC.91379.pdf)

I have attached a PDF showing which letters and characters are printed and which are not. The printed text is this example file from DIN 91379: https://github.com/String-Latin/DIN-91379-Characters-and-Sequences/blob/main/latin_list_all_en.txt

For more information see: https://github.com/String-Latin/DIN-91379-Characters-and-Sequences

jsangmeister:
I think this is a bug rather than a feature - maybe @m-schieder an take a look at this?

m-schieder:
I think this is because the fonts we use for PDF export do not contain these characters. I'll take a closer look.

m-schieder:
The problem also exists with other fonts that contain these characters. These characters are also transferred correctly to pdfmake.  In [pdfmake playground](http://pdfmake.org/playground.html) such characters can be displayed. Either pdfmake cannot interpret the font I am using correctly or there is something wrong with our implementation of pdfmake.

emanuelschuetze:
@alveolus can you create a pull request - or send me the new woff files

alveolus:
@emanuelschuetze Use [this file](https://github.com/String-Latin/DIN-91379-Characters-and-Sequences/blob/de3e846e674111c5f6d6ca40a145f24c52f818cf/latin_letters_1.3.txt) for comparison: If you see more characters with the new fonts than before (or than in [this file](https://github.com/OpenSlides/openslides-client/files/10014640/DIN-91379-Characters-and-Sequences_latin_letters_1.3.txt.pdf), created with the font files without Latin extended), then everything works fine.

In the user interface you can see the change best in the account list: Create an account that contains the letter `Ş` in its name. If the letter is displayed in the same font and size as the rest of the name/text, then everything works here too.

PR #1808, opened 2022-11-15
Updated the fira sans font files with charset latin-extended

alveolus:
Fixed https://github.com/OpenSlides/openslides-client/issues/1172


emanuelschuetze:
I tested your new font files by creating a new motion with [theses chars](https://raw.githubusercontent.com/String-Latin/DIN-91379-Characters-and-Sequences/de3e846e674111c5f6d6ca40a145f24c52f818cf/latin_letters_1.3.txt).

With current (upstream) latin font files:
latin-font-1

With your extendet latin files:
latin-font-2

There are a lot of new supported chars now. But also some unsupported chars. Can we ignore it?
User list PDF with your full name works as expected.

alveolus:
@emanuelschuetze We could also support all 7 character sets: latin, latin-ext, greek, greek-ext, cyrillic, cyrillic-ext, vietnamese. That would be my favorite.

For the font files in woff format this would mean:
Fira Sans, version 16 (2022-09-22)
Fira Sans Condensed, version 10 (2022-09-22)

alveolus:
As discussed with @emanuelschuetze: We include the latin and latin-ext charsets for the following fonts in the release:

Fira Sans, version 16 (2022-09-22)
Fira Sans Condensed, version 10 (2022-09-22)

For the charsets greek, greek-ext, cyrillic, cyrillic-ext and vietnamese I will create a FAQ.

emanuelschuetze (approved):

PR #1809, opened 2022-11-15
dont display "no meetings available" before first autoupdate

### LINKED ISSUE #1238: Implement the info if an HTTP stream has got its first data

### LINKED ISSUE #1803: Show if meetings are still loading on calendar

bastianjoel:
resolves #1803 

This also adds a method `waitSubscriptionReady` to `ModelRequestService` to check if a subscription got data. Therefore improves #1715 and further resolves #1238.


luisa-beerboom (approved):

bastianjoel:
Currently `no meetings available` pops up for a short moment before the meetings are shown.

emanuelschuetze (approved):

PR #1819, opened 2022-11-18
Client side vote decrypt handling

### LINKED ISSUE #1744: Implement client-side crypto-voting logic
https://github.com/OpenSlides/vote-decrypt
https://github.com/OpenSlides/openslides-vote-client

luisa-beerboom:
Closes #1744


client/src/app/gateways/vote-decrypt-gateway.service.ts
```diff
@@ -0,0 +1,82 @@
...
+        const hkdfKey = await this.generateHkdfKey(sharedSecret)
+        console.log(`hkdfKey: `, hkdfKey);
+
+        const encrypted = new Uint8Array(await (crypto.subtle.encrypt({name: `AES-GCM`, iv: nonce}, hkdfKey, plaintext) as Promise<ArrayBuffer>));
+        console.log(`encrypted: `, encrypted);
+
+        return this.createPayload(ephPublKey, nonce, encrypted);
+    }
+
+    private async generateKeysAndSecret(cryptKey: Uint8Array): Promise<{ephPrivKey: Uint8Array, ephPublKey: Uint8Array, sharedSecret: Uint8Array}> {
```
ostcar, line 56: The private key does not have to be returned. Only the public key and the sharedSecret are needed

ostcar:
> @rrenkert @ostcar I'm currently at my wits end with my `encryptVote` method.
> 
> I've tried a lot of things, but instead of the votes I still always get an `encryption not valid` error when I stop the poll. Could you have a look at my code? Maybe you can see what I am doing wrong....

From looking at the code, it looks fine.

I am not so good at the javascript environment. When I test you branch, I only get the console log `----------[encryptVote]----------`but not the messages from line 35, 45, etc. I tested this in firefox and chrome. Can you tell me what I have to do to get this messages?

ostcar:
I tried to debug the client but failed. If something is wrong, then it silently stops without printing any error (for example when poll.crypt_key is undefined, this happens, when a wrong version of the backend is used).

I had expected, that typescript makes sure, that all variables have the correct type, but this does not be the case.

Could you add a test that calls the function with specific inputs? So it can be run and debug the one test without running everything else?

luisa-beerboom:
> I am not so good at the javascript environment. When I test you branch, I only get the console log `----------[encryptVote]----------`but not the messages from line 35, 45, etc. I tested this in firefox and chrome. Can you tell me what I have to do to get this messages?

The only way I could imagine this might happen is if you aren't receiving all of the keys, in which case I would assume that your environment is not set up correctly.
For now, I've built in an Error message for that.

Please check if you've updated all relevant repositories to the correct commit, if all relevant services are initialized correctly and the vote decrypt is initializing correctly.
If not please fix that first.

If you're then stopped when the vote should be verified, that means that the ed25519 library is not loaded in your instance (for some reason).

After that the function **should** terminate at least.

luisa-beerboom:
> I tried to debug the client but failed. If something is wrong, then it silently stops without printing any error (for example when poll.crypt_key is undefined, this happens, when a wrong version of the backend is used).

I've also noticed this behavior of not doing anything, sadly there isn't much I can do about most of these situations (in the case of empty poll keys it is to be expected that it wouldn't work though)

> I had expected, that typescript makes sure, that all variables have the correct type, but this does not be the case.

What do you mean by that?

> Could you add a test that calls the function with specific inputs? So it can be run and debug the one test without running everything else?

How will that help? I won't exactly be capable of checking whether the encryption is correct unless I have a working vote-decrypt-service. Generating example data and checking whether the result is correct is hard since I can't just compare it to a previously generated result (that would have been generated with a different private key) and mirroring the vote-decrypt services decryption would also be rather pointless, given that I couldn't be sure about its accuracy.
Not to mention, that it would be a massive amount of work.

luisa-beerboom:
Update: `encryptVote` works now

client/src/app/gateways/vote-decrypt-gateway.service.ts
```diff
@@ -0,0 +1,185 @@
...
+        console.log(`public key generated`);
+
+        const secret = curve25519.scalarMult(privKey, cryptKey);
+        return {ephPublKey: publKey, sharedSecret: secret};
+    }
+
+    private async generateHkdfKey(sharedSecret: Uint8Array): Promise<CryptoKey> {
+        return crypto.subtle.deriveKey(
+            {name: `HKDF`, hash: `SHA-256`, salt: new ArrayBuffer(0), info: new ArrayBuffer(0)},
+            await crypto.subtle.importKey(`raw`, sharedSecret.buffer, `HKDF`, true, [`deriveKey`]),
```
rrenkert, line 171: Please set the 4th param to false. A suppressed exception says th base for derived keys cannot be extractable.

luisa-beerboom:
The weird bug where the encryption/verification is sometimes not working is somehow connected to the autoupdate (which returns an error when the request for the organization subscription is sent). 
Looking at the logs, the autoupdate seems to be logging an error "Error: getting next message: creating first time data: get restricted data: restricting data: grouping keys by collection: getting restriction Mode for organization/1/vote_decrypt_public_main_key: fqfield "organization/vote_decrypt_public_main_key" is unknown, maybe run go generate ./... to fetch all fields from the models.yml"

This is strange since I am running both a backend and an autoupdate service that are updated to function with vote-decrypt. It is also strange that it starts to work when I reload the client. 

Steps to reproduce:

1. Log out
2. Reload page (press F5)
3. Log back in
4. The verification won't work for finished crypto polls and any new crypto votes aren't sent.

I am currently suspecting the autoupdate service and/or the client-side autoupdate service worker. Since there are errors in the autoupdate logs, I'm thinking it might be the autoupdate (I'm using the commit in OpenSlides/openslides-autoupdate-service#549 ).
@ostcar @bastianjoel since you're the specialists on the aforementioned parts of the code, do you have any ideas what might cause this? Is this a new bug, or is this maybe already fixed in some other newer PR? A lot of the code that I am running is a few commits behind after all.

client/src/app/gateways/vote-decrypt-gateway.service.ts
```diff
@@ -0,0 +1,185 @@
...
+            {name: `AES-GCM`, length: 256},
+            true,
```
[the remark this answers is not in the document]
luisa-beerboom: Done

client/src/app/gateways/vote-decrypt-gateway.service.ts
```diff
@@ -0,0 +1,359 @@
...
+                })
+            )
+            .subscribe(verificationData =>
+                this.updateVerification(
+                    verificationData.map(date => (({ options, ...others }) => others as VoteVerificationData)(date))
+                )
+            );
+    }
+
+    public async encryptVote(poll: ViewPoll, vote: string): Promise<string> {
```
bastianjoel, line 113: It might be helpful to add docs to public member methods.

client/src/app/site/pages/meetings/modules/poll/services/poll-controller.service/poll-controller.service.ts
```diff
@@ -1,22 +1,48 @@
...
 import { PollRepositoryService } from 'src/app/gateways/repositories/polls/poll-repository.service';
+import { VoteDecryptGatewayService } from 'src/app/gateways/vote-decrypt-gateway.service';
 import { BaseMeetingControllerService } from 'src/app/site/pages/meetings/base/base-meeting-controller.service';
 import { MeetingControllerServiceCollectorService } from 'src/app/site/pages/meetings/services/meeting-controller-service-collector.service';
+import { LifecycleService } from 'src/app/site/services/lifecycle.service';
 
 import { ViewPoll } from '../../../../pages/polls';
 import { PollServiceModule } from '../poll-service.module';
 
+function getRandomString(length: number) {
```
bastianjoel, line 16: Could also be moved to `client/src/app/infrastructure/utils/functions.ts`

client/src/app/gateways/vote-decrypt-gateway.service.ts
```diff
@@ -0,0 +1,359 @@
...
+            const raws = uInt8Enc(date.raws, false);
+            let resultVerificationErrors: string[] = [];
+            try {
+                const verified = await verify(signature, raws, this._publicMainKey);
+                resultVerificationErrors = verified ? resultVerificationErrors : [`Signature verification failed`];
+            } catch (e) {
+                resultVerificationErrors.push(`Signature verification failed with error: "${e.message}"`);
+            }
+            try {
+                resultVerificationErrors = resultVerificationErrors.concat(this.verifyBackendPollResults(model));
```
bastianjoel, line 199: I think it might be better to handle the throwing method call within `verifyBackendPollResults` as it returns a error array already.

bastianjoel (changes requested):

client/src/app/gateways/vote-decrypt-gateway.service.ts
```diff
@@ -0,0 +1,359 @@
...
+        ];
+        await this.hasPublicMainKey;
+        await this.verify(verificationData);
+    }
+
+    private async updateVerification(verificationData: VoteVerificationData[]): Promise<void> {
+        const prev = this._lastVoteVerificationData.sort((a, b) => a.id - b.id);
+        const curr = verificationData.sort((a, b) => a.id - b.id);
+
+        const toUnverify: VoteVerificationData[] = [];
```
bastianjoel, line 157: I think something like `toRevoke` would be clearer here.

bastianjoel (changes requested):

client/src/app/gateways/vote-decrypt-gateway.service.ts
luisa-beerboom: It may be more practical to write into the documentation of `verifyBackendPollResult` that the function may throw an error from the `getSortedVoteArrays` function

luisa-beerboom:
I've moved the change requests to a new issue `Clean up vote decrypt code` (#1867), I'll merge this into the feature branch now and handle the cleanup later.

PR #1825, opened 2022-11-22
sequential number service performance

bastianjoel:
With this PR the AU connection `SequentialNumberMappingService:prepare` gets replaced by single AU requests which will only be sent on demand. 
The sequential numbers will also be subscribed within list subscriptions.


luisa-beerboom (approved):

luisa-beerboom:
Does this work if I just call a url with a sequential number that hasn't been loaded yet? If yes, then I have no complaints.

rrenkert (approved):

PR #1828, opened 2022-11-24
check cdk viewport size on scrolling table refresh

### LINKED ISSUE #1816: UI bug in motion blocks: more than two motions are not always correctly displayed
**Description:** The motion block detail view has an UI bug. Only two motions are displayed. The other motions appears, if the user reloads the page. Note: The counter of the motion block show the correct number of included motions, but not the listing. 

Additionally: If a motion is deleted from the listing, the next one on the list became visible.

Screenshot_20221116_132001

**Reproduction:**
1. open a meeting
2. open motions
3. create four motions
4. open the motions menu, click on the 3-dot-menu and choose motion blocks
5. create a new motion block
6. go back to the motions menu and add all motions to the created block. you can do this by opening the 3-dot-menu, choose multiselect, click again on the 3-dot-menu, choose select all, again 3-dot-menu and choose set motion block. Choose the created block and click on OK. leave multiselect
7. go back to motion blocks menu
8. open the motion block and check how many motions are visible. 
9. delete the second one from the listing. 

**Expected behavior:**  All motions inside a motion block should be visible.

bastianjoel:
resolves #1816 

With #1816 you can also resize your window to get all elements displayed. Therefore it seems like the cdk viewport size needs to be rechecked manually sometimes if new items were added to a scrolling table.

### DIFF
--- a/client/src/app/ui/modules/scrolling-table/components/scrolling-table/scrolling-table.component.ts
+++ b/client/src/app/ui/modules/scrolling-table/components/scrolling-table/scrolling-table.component.ts
@@ -277,5 +277,6 @@ export class ScrollingTableComponent<T extends Partial<Mutable<Identifiable>>>
     private refresh(): void {
         this._dataSource.next(this._source);
         this.cd.markForCheck();
+        this.scrollViewport.checkViewportSize();
     }
 }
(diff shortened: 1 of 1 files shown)


rrenkert (approved):

emanuelschuetze (approved):
Tested. I added a `?`for better error handling. see extra commit.

PR #1836, opened 2022-11-29
fix stored filters overwritten in BaseFilterListService

### LINKED ISSUE #1831: Reload motion page with active filters deletes set filter options
Description: If a user has set filters in category or tags, reloading the page will delete the filter options. The unfiltered motion list will be loaded.

Reproduction:
1. open meeting
2. open motions
3. create a motion 
4. create a tag and a category over three-dot-menu in motions menu
5. open the motion list and set category or tag as a filter and reload the page.

Expected behavior: After page reload the filters should stay active.

bastianjoel:
resolves #1831


emanuelschuetze (changes requested):
Does not work: Add tag filter in motions list -> reload -> filter disappears! full motion list is visible.

bastianjoel:
TODO: Check if it has any consequences to change the initial value of viewModelListSubject to null

emanuelschuetze (changes requested):
Retested on testintern with big meeting:
* Test 1: filter motion list by tag (or motionblock/forwards/speakers/amendments/favorit/personal-notes/motion-comment) -> reload => works: filter is still active
* Test 2: filter motion list by category or state or recommendation -> reload => does NOT work: filter is gone!

bastianjoel:
This PR together with #1825 would resolve the issue from the second test case.

rrenkert (approved):

emanuelschuetze (approved):
works now as expected

PR #1845, opened 2022-12-02
fix autoupdate fallback when logged in

### LINKED ISSUE #1844: autoupdate shared worker fallback gets stuck at loading screen
In case the browser does not support shared workers there is a fallback for the autoupdate worker to run within the window. 

With this fallback the user sometimes gets stuck when logged in at the loading screen. This seems to be caused by the `ready` state exchange. 

**Reproduction**
* Disable shared worker
* Login
* Reload

bastianjoel:
resolves #1844 

This PR resolves two issues within the SharedWorkerService and also adds a configuration to the environment to disable the shared worker. The shared worker is now disabled in non production setups by default to make debugging easier. 

**SharedWorkerService issues**   
1. It was possible that the ready state would not be set because the subscription for messages was added after the worker was already started. 
2. The AU service would block the ready state if the whoami request was resolved before the application was ready. This happened because the AU service uses a `setTimeout` to renew auth tokens. To resolve this the autoupdate worker was moved outside the angular zone to not effect the ready state. Note that this only happened for the Shared Worker fallback.


emanuelschuetze (approved):
Not tested/reproduced

rrenkert (approved):

PR #1853, opened 2022-12-08
Added ErrorMap for vote service

emanuelschuetze:


luisa-beerboom (approved):
Code-wise this is fine. There are some duplications among the error replacements, wouldn't it better if they were a bit more specific.

emanuelschuetze:
> There are some duplications among the error replacements, wouldn't it better if they were a bit more specific.

right, I don't like to make these voting errors more specific - in fact the users have an error while voting. we/admin will check the reason (e.g. wrong config which a user cannot change)

PR #1857, opened 2022-12-09
Add parameter to exclude ids from search in base search selector.

### LINKED ISSUE #1802: List of speakers search selector searches by id
List of speakers search selector searches by internal id, when it should search by participant number

rrenkert:
resolves #1802

Introduces a parameter to exclude IDs from search.
Maybe we should have a look at other search fields that can make usage of this parameter.

Changes base class! Please test other search components that use this base component.


emanuelschuetze (approved):
Tested for speakers, submitters and candidates selector

PR #1860, opened 2022-12-13
Rework of action worker handling

### LINKED ISSUE #1847: Repair and rework action worker dialog
In PR #1843 the action worker dialog was crippled to the point of undermining it's base functionality.
This was necessary since the dialog was too long and too technical.

We should:

1. Re-introduce the action worker dialog's base functionality of allowing the user to choose to wait or to stop waiting
2. Use short and concise info texts that are actually fitting for the given situation
3. Relocate the functionality to somewhere where it isn't as invasive
4. Re-evaluate when certain options should be given to counteract mindless wait-clicking

After discussing with @rrenkert we had the following ideas:

- [x] 3 -> Instead of using a dialog, relocate the functionality to the banner or something similar. There you can show a short info text along with the relevant option buttons.
- [x] 2-> The info texts should be something along the lines of:
    - [x] on status not written (notWritten): "The process will be started in the future, please wait."
    - [x]  after 2sek since last update (slow): "The process is still running, please wait."
    - [x] after 20sek since last update (inactive): "The process may have stopped running."
    - [x] after 1min since last update (dead): Automatically stop waiting and inform the user that the client has stopped waiting (there should be an option to open an info text with all of the relevant information for debugging (worker_id, worker_name, last activity timestamp, timestamp at which the waiting was stopped, etc))
- [x] 1 & 4 -> Upon 'notWritten' and 'slow' there should only be the `wait` option, not the `close` option. For 'inactive' the `close` option should additionally be shown.

luisa-beerboom:
Closes #1847


bastianjoel (approved):

rrenkert (approved):
Showing the info in the banner is very good!
Only the JSON output of the  dialog in case the the backend action seems to be dead can have a better layout, but its good enough for me.

PR #1879, opened 2022-12-16
Made flex-vertical-center style's alignment safe

### LINKED ISSUE #568: The UI of the CSV-Import of topics is ugly
UI: The content of the column `text` exceeds its dimensions (when too long)

luisa-beerboom:
To fix participant import text field (which cut off the top of the content if the content got too long).

Closes #568 

To do this I changed a global style (`flex-vertical-center`) which is used in all kinds of components:
Among others:

- basically all major lists (participant, motions, import lists, file lists, etc.)
- account dialog, user-delete-dialog
- participant detail view
- dashboard
- search selectors

I don't think this will actually break anything (as far as I have seen it doesn't), but it would probably be a good idea to check, if the related views look weird, when testing.

### DIFF
--- a/client/src/styles.scss
+++ b/client/src/styles.scss
@@ -802,7 +802,6 @@ button.mat-menu-item.selected {
 
 .flex-vertical-center {
     display: flex;
-    // align-items: safe center;
     flex-direction: row;
 
     &.align-left {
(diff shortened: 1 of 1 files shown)


bastianjoel:
`align-items: safe center` seems to be only supported by Firefox. 

https://caniuse.com/mdn-css_properties_align-items_flex_context_safe_unsafe

bastianjoel (approved):

emanuelschuetze (approved):

PR #1883, opened 2023-01-03
Motion list displays last_modified instead of created timestamp if th…

### LINKED ISSUE #1747: motions: sort option "last modified" misses a status to recognize what is old/new
**Description:** If a user uses the sort option "last modified" it's very hard to to recognize which sorting order displays the last modified motion. None of the motions has an indicator, when it was last modified. It's only visible in the history.

**Reproduction:** 
1. open a meeting
2. go to motions
3. create some motions with the plus icon (at least four)
4. open the motions menu
5. use the sort option and select "last modified"

**Hint:** The issues needs the #1727.

**Expected behavior:** Optimal would be an information in the motion overview, exaclty in the motion line behind created. So it's easy to regconize which motion is the last modified one. 

@emanuelschuetze What did you think? Should we implement this information?

luisa-beerboom:
…ere have been modifications

Closes #1747

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.html
@@ -101,8 +101,8 @@
 
                     <span>
                         <span *ngIf="motion.submitters.length || showSequential">&middot;&nbsp;</span>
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

emanuelschuetze (approved):

emanuelschuetze:
Wait for merging because new translations are required...

PR #1887, opened 2023-01-04
Add consecutive selections to committee recommendation selection

### LINKED ISSUE #1754: Motion detail view: Add multi select to the recommendation field "committees"
`…/…/motions/…`

### Feature description

1. In the selection field "Committees" it is possible to select several committees as targets in one step via multiple selection. These are then transferred to the field "Extension". The names of the committees are separated by a · 
2. Furthermore, the selection field folds out in such a way that the names of the committees are displayed as completely as possible without having to scroll horizontally (at least twice as wide as currently). → The screenshot is a bad example, most committees have much longer names.

 → 

### Problem · Background
It is currently not possible to select multiple target committees in one step. This currently has to be done one by one and the selection field has to be expanded again and again. The names of the committees are lined up directly one after the other without any space between them.

jsangmeister:
resolves #1754

also increases width of committee search selector

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-extension-field/motion-extension-field.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-extension-field/motion-extension-field.component.html
@@ -41,10 +41,11 @@
     <mat-form-field *ngFor="let list of searchLists; let i = index" [formGroup]="extensionFieldForm">
         <os-list-search-selector
             formControlName="{{ 'list' + i + '' }}"
-            [inputListValues]="list"
-            placeholder="{{ searchListLabels[i] | translate }}"
-            [keepOpen]="searchListsKeepOpen[i]"
+            [inputListValues]="list.observable"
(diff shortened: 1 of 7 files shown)


luisa-beerboom (approved):

rrenkert (dismissed):
There are two points that could be better:
* As long as the selection box is open one can select a target multiple times and it appears multiple times in the recomendation field. I know there is no reference if a target is selected, but we should find a way to improve this.
* If the selection box is closed and reopened, selected targets have no separator.

jsangmeister:
> * As long as the selection box is open one can select a target multiple times and it appears multiple times in the recomendation field. I know there is no reference if a target is selected, but we should find a way to improve this.

This left as-is on purpose as this was the previous functionality and maybe someone wants to select the same committee twice. But if you don't think so, we can change that - I would have left it to the user to choose if he wants to select anything multiple times or not.

> * If the selection box is closed and reopened, selected targets have no separator.

Okay, I will change this.

emanuelschuetze (approved):
Tested - also the remarks of @rrenkert works now

PR #1889, opened 2023-01-04
Fix forwarding filtering in search selector

### LINKED ISSUE #1888: Filtering results in a search value selector from the motion forwarding service does not work
When the result of the `MotionForwardDialogService` (the `forwardingCommitteesObservable`) is filtered in a search selector (e.g. in the motion detail view when editing the recommendation extension), the filtering does not work.

jsangmeister:
fixes #1888

The `toString()` function was missing on the "artificial models" returned by the forwarding service, so they could not be correctly filtered by the search selector.


luisa-beerboom (approved):

luisa-beerboom:
Looks fine, but what's going on with the tests?

jsangmeister:
I don't know, I'm looking into it

jsangmeister:
I just missed a rename of the function, should work now. Interesting to note, however, that the prod test failed with the correct output of the error because it could not compile the code, while the dev setup only failed internally and did not output any error except the timeout in the `wait-on` - something to keep in mind for situations like these.

rrenkert (approved):

PR #1890, opened 2023-01-04
Implement global webinterface logo

### LINKED ISSUE #1752: Feature: add global settings for fonts and logos / logo on orga-level could be changed
**Currnet situation:** 
Currently it is not possible to change the logo on the Organization-level. Also it is not possible to set a logo and fonts for all meetings as a default value.

**Expected behavior:** 
- [ ] The logo on the organization level could be changed 
- [ ] it is possible to set a logo and fonts as default values for all meetings. 

**Reason:** 
The reason for this feature request is, that most organizations have a corporate design which is organization wide and used on every level. With this feature an organization admin could set logos and fonts as defaults values for all meetings according to the CD.

**How it could be implemented:**
1. File menu is needed on organization level. (This is done by one of the next updates)
2. create different checkboxes in three-dot-menu behind uploaded files. same like the ones in meeting, only with different content.
Checkboxes for images:
- [ ] organization level logo
- [ ] set logo as global default 

Checkboxes for fonts:
- [ ] use font as global for .... (here we need several options like the ones in a meeting; like bold etc.)

3. After the files are set global, they are copied in a new meeting. This is necessary to change them on meeting level and switch back. 

**Important:** 
Most importan is the possibility to change the logo on organization level. The global setting would be a nice quality of live improvement for large organizations.

luisa-beerboom:
Closes #1752

### DIFF
--- a/client/src/app/domain/models/mediafiles/mediafile.constants.ts
+++ b/client/src/app/domain/models/mediafiles/mediafile.constants.ts
@@ -1,5 +1,8 @@
 import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 
+/**
+ * Acceptable values are different places where logos can be placed.
+ */
 export type LogoPlace =
     | 'projector_main'
     | 'projector_header'
@@ -20,6 +23,9 @@ export const LogoDisplayNames: { [place in LogoPlace]: string } = {
     pdf_ballot_paper: _(`PDF ballot paper logo`)
 };
 
+/**
+ * Acceptable values are different places where fonts can be placed.
+ */
 export type FontPlace =
     | 'regular'
     | 'italic'
--- a/client/src/app/domain/models/mediafiles/mediafile.constants.ts
+++ b/client/src/app/domain/models/mediafiles/mediafile.constants.ts
@@ -1,21 +1,6 @@
 import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 
-const PossibleLogoPlaces = [
-    `projector_main`,
-    `projector_header`,
-    `web_header`,
-    `pdf_header_l`,
-    `pdf_header_r`,
-    `pdf_footer_l`,
-    `pdf_footer_r`,
-    `pdf_ballot_paper`
-] as const;
-
-/**
- * Acceptable values are different places where logos can be placed.
- */
-export type LogoPlace = typeof PossibleLogoPlaces[number];
-export const LogoDisplayNames: { [place in LogoPlace]: string } = {
+export const LogoDisplayNames = {
     projector_main: _(`Projector logo`),
     projector_header: _(`Projector header image`),
     web_header: _(`Web interface header logo`),
@@ -25,27 +10,10 @@ export const LogoDisplayNames: { [place in LogoPlace]: string } = {
     pdf_footer_r: _(`PDF footer logo (right)`),
     pdf_ballot_paper: _(`PDF ballot paper logo`)
 };
-/**
- * All different places where logos can be placed.
- */
-export const LogoPlaces = Array.from(PossibleLogoPlaces.values());
+export type LogoPlace = keyof typeof LogoDisplayNames;
+export const LOGO_PLACES = Object.keys(LogoDisplayNames) as LogoPlace[];
 
-const PossibleFontPlaces = [
-    `regular`,
-    `italic`,
-    `bold`,
-    `bold_italic`,
-    `monospace`,
-    `chyron_speaker_name`,
-    `projector_h1`,
-    `projector_h2`
-] as const;
-
-/**
- * Acceptable values are different places where logos can be placed.
- */
-export type FontPlace = typeof PossibleFontPlaces[number];
-export const FontDisplayNames: { [place in FontPlace]: string } = {
+export const FontDisplayNames = {
     regular: _(`Font regular`),
     italic: _(`Font italic`),
     bold: _(`Font bold`),
@@ -55,6 +23,7 @@ export const FontDisplayNames: { [place in FontPlace]: string } = {
     projector_h1: _(`Projector h1`),
     projector_h2: _(`Projector h2`)
 };
+export type FontPlace = keyof typeof FontDisplayNames;
 export const FontDefaults: { [place in FontPlace]: string } = {
     regular: `assets/fonts/fira-sans-latin-400.woff`,
     italic: `assets/fonts/fira-sans-latin-400italic.woff`,
@@ -65,7 +34,4 @@ export const FontDefaults: { [place in FontPlace]: string } = {
     projector_h1: `assets/fonts/fira-sans-latin-500.woff`,
     projector_h2: `assets/fonts/fira-sans-latin-400.woff`
 };
-/**
- * All different places where fonts can be placed.
- */
-export const FontPlaces = Array.from(PossibleFontPlaces.values());
+export const FONT_PLACES = Object.keys(FontDisplayNames) as FontPlace[];
--- a/client/src/app/domain/models/mediafiles/mediafile.constants.ts
+++ b/client/src/app/domain/models/mediafiles/mediafile.constants.ts
@@ -1,17 +1,20 @@
 import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 
+const PossibleLogoPlaces = [
+    `projector_main`,
+    `projector_header`,
+    `web_header`,
+    `pdf_header_l`,
+    `pdf_header_r`,
+    `pdf_footer_l`,
+    `pdf_footer_r`,
+    `pdf_ballot_paper`
+] as const;
+
 /**
  * Acceptable values are different places where logos can be placed.
  */
-export type LogoPlace =
-    | 'projector_main'
-    | 'projector_header'
-    | 'web_header'
-    | 'pdf_header_l'
-    | 'pdf_header_r'
-    | 'pdf_footer_l'
-    | 'pdf_footer_r'
-    | 'pdf_ballot_paper';
+export type LogoPlace = typeof PossibleLogoPlaces[number];
 export const LogoDisplayNames: { [place in LogoPlace]: string } = {
     projector_main: _(`Projector logo`),
     projector_header: _(`Projector header image`),
@@ -22,19 +25,26 @@ export const LogoDisplayNames: { [place in LogoPlace]: string } = {
     pdf_footer_r: _(`PDF footer logo (right)`),
     pdf_ballot_paper: _(`PDF ballot paper logo`)
 };
+/**
+ * All different places where logos can be placed.
+ */
+export const LogoPlaces = Array.from(PossibleLogoPlaces.values());
+
+const PossibleFontPlaces = [
+    `regular`,
+    `italic`,
+    `bold`,
+    `bold_italic`,
+    `monospace`,
+    `chyron_speaker_name`,
+    `projector_h1`,
+    `projector_h2`
+] as const;
 
 /**
- * Acceptable values are different places where fonts can be placed.
+ * Acceptable values are different places where logos can be placed.
  */
-export type FontPlace =
-    | 'regular'
-    | 'italic'
-    | 'bold'
-    | 'bold_italic'
-    | 'monospace'
-    | 'chyron_speaker_name'
-    | 'projector_h1'
-    | 'projector_h2';
+export type FontPlace = typeof PossibleFontPlaces[number];
 export const FontDisplayNames: { [place in FontPlace]: string } = {
     regular: _(`Font regular`),
     italic: _(`Font italic`),
@@ -55,3 +65,7 @@ export const FontDefaults: { [place in FontPlace]: string } = {
     projector_h1: `assets/fonts/fira-sans-latin-500.woff`,
     projector_h2: `assets/fonts/fira-sans-latin-400.woff`
 };
+/**
+ * All different places where fonts can be placed.
+ */
+export const FontPlaces = Array.from(PossibleFontPlaces.values());
--- a/client/src/app/site/pages/meetings/services/media-manage.service.ts
+++ b/client/src/app/site/pages/meetings/services/media-manage.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { BehaviorSubject, Observable } from 'rxjs';
+import { BehaviorSubject, merge, Observable } from 'rxjs';
 import { Id } from 'src/app/domain/definitions/key-types';
 import {
     FontDefaults,
@@ -9,6 +9,7 @@ import {
     LogoPlace
 } from 'src/app/domain/models/mediafiles/mediafile.constants';
 import { MeetingMediaAdapterService } from 'src/app/gateways/meeting-media-adapter.service';
+import { MediafileRepositoryService } from 'src/app/gateways/repositories/mediafiles/mediafile-repository.service';
 
 import { ViewMediafile } from '../pages/mediafiles';
 import { ActiveMeetingService } from './active-meeting.service';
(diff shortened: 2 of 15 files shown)


jsangmeister (changes requested):
This misuses the `token` functionality in a really bad way. Especially the list-emulation in a string field is really hacky. We need to find a better approach for this feature.

The `token` field was copied from the old `ressource` collection and was never meant to be used in such a scope. It had a single purpose, displaying the organization wide `web_header` on the login page (and, potentially, being one day extended to some fonts or something). (Altough in retrospect, I don't understand why we did not design it as an array in the first place.) Implementing this (the first bullet point from https://github.com/OpenSlides/openslides-client/issues/1752) is no problem, but the other points with the default mediafiles for meetings require backend changes to be implemented nicely.

We have two options as I see it:
- reduce the content of this PR to the first bullet point, since the other were optional QoL improvements anyway and will require some time to be working properly.
- Enhance the backend with relations for the new feature. While we're at it, I would like to redesign the `token` field into a relation to `organization` and handle them almost analogously to the meeting relations. Preferably, this would be done using template fields, although that's maybe not optimal considering that we're planning to remove them soon:
```
organization:
  [...]
  logo_$_id: [...]
  default_logo_$_id: [...]
  default_font_$_id: [...]

mediafile:
  used_as_logo_$_in_organization_id: [...]
  used_as_default_logo_$_in_organization_id: [...]
  used_as_default_font_$_in_organization_id: [...]
```
This could also be modeled without template fields, it would just become much more verbose. If the organization also needs fonts, these can be implemented analogously.

We should discuss this more in-depth.

luisa-beerboom:
UPDATE: For now we're going with the first option

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -48,16 +49,19 @@ export class MediaManageService {
```
jsangmeister, line 49: Please unify the lists of places at the start of this class with the ones in the `mediafile.constants.ts`

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -48,16 +49,19 @@ export class MediaManageService {
...
     public constructor(
         private activeMeetingService: ActiveMeetingService,
-        private mediaAdapter: MeetingMediaAdapterService
+        private mediaAdapter: MeetingMediaAdapterService,
+        private mediaRepo: MediafileRepositoryService
     ) {
-        this.activeMeetingService.meetingObservable.subscribe(_ => {
-            for (const place of Object.keys(this.logoUrlSubjects)) {
-                (<any>this.logoUrlSubjects)[place].next(this.getLogoUrl(place as LogoPlace));
+        merge(this.activeMeetingService.meetingObservable, this.mediaRepo.getViewModelListUnsafeObservable()).subscribe(
```
jsangmeister, line 55: The `unsafe` prefix worries me, is it necessary here?

jsangmeister (changes requested):
- since tokens are not lists anymore, you can replace any `includes` checks which you do on the token field with a simle `===` and all array emulation on it
- since we won't implement the orga defaults via tokens, you can remove everything regarding the `orga`/`global` prefixes and simplify the code where needed

client/src/app/site/pages/meetings/services/media-manage.service.ts
luisa-beerboom: Yes

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -68,8 +72,12 @@ export class MediaManageService {
...
 
     public getLogoUrl(place: LogoPlace): string | null {
-        // Note: we are not fetching the mediafile view model at any place.
-        const mediafileId = this.activeMeetingService.meeting?.logo_id(place);
+        // Note: we are not fetching the mediafile view model at any place except when filtering for the defaults.
+        const mediafileId =
+            (this.activeMeetingService.meetingId
+                ? this.activeMeetingService.meeting?.logo_id(place)
+                : this.getGeneralMediafileIdByToken(`orga`, place)) ??
+            this.getGeneralMediafileIdByToken(`global`, place);
```
luisa-beerboom, line 80: Take orga out of this calculation (it can't be filled).
Also remove the replacement with global default if there is no id, instead put separate check for global default into `else if ` block  if `mediafileId` is empty

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -18,46 +21,31 @@ import { ActiveMeetingService } from './active-meeting.service';
 })
 export class MediaManageService {
     public get allLogoPlaces(): LogoPlace[] {
```
jsangmeister, line 23: Can be reduced to a simple readonly variable instead of a getter

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -18,46 +21,31 @@ import { ActiveMeetingService } from './active-meeting.service';
...
+        private mediaAdapter: MeetingMediaAdapterService,
+        private mediaRepo: MediafileRepositoryService
     ) {
-        this.activeMeetingService.meetingObservable.subscribe(_ => {
-            for (const place of Object.keys(this.logoUrlSubjects)) {
-                (<any>this.logoUrlSubjects)[place].next(this.getLogoUrl(place as LogoPlace));
+        merge(this.activeMeetingService.meetingObservable, this.mediaRepo.getViewModelListUnsafeObservable()).subscribe(
+            _ => {
+                for (const place of Object.keys(this.logoUrlSubjects)) {
+                    (<any>this.logoUrlSubjects)[place].next(this.getLogoUrl(place as LogoPlace));
```
jsangmeister, line 42: Can you improve this typing? Casting to `any` is always a bad idea, maybe you could e.g. cast `place` to `LogoPlace` first, or what is the problem why the typing isn't recognized correctly without casting?

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -68,10 +56,15 @@ export class MediaManageService {
     }
 
     public getLogoUrl(place: LogoPlace): string | null {
-        // Note: we are not fetching the mediafile view model at any place.
-        const mediafileId = this.activeMeetingService.meeting?.logo_id(place);
+        // Note: we are not fetching the mediafile view model at any place except when filtering for the defaults.
```
jsangmeister, line 59: Update the comment, the defaults no longer exist

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -117,6 +110,16 @@ export class MediaManageService {
         await this.mediaAdapter.unsetFont(place, this.activeMeetingService.meetingId!);
     }
 
+    private getGlobalMediafileIdByToken(place: LogoPlace | FontPlace): number {
+        return (
+            this.mediaRepo.getViewModelListUnsafe().find(file => {
+                return (
+                    file.isImage() && Object.keys(LogoDisplayNames).includes(place) && file.token === `global-${place}`
```
jsangmeister, line 117:
```suggestion
                    file.isImage() && Object.keys(LogoDisplayNames).includes(place) && file.token === `${place}`
```

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -68,10 +56,15 @@ export class MediaManageService {
     }
 
     public getLogoUrl(place: LogoPlace): string | null {
-        // Note: we are not fetching the mediafile view model at any place.
-        const mediafileId = this.activeMeetingService.meeting?.logo_id(place);
+        // Note: we are not fetching the mediafile view model at any place except when filtering for the defaults.
+        const mediafileId = this.activeMeetingService.meetingId
+            ? this.activeMeetingService.meeting?.logo_id(place)
+            : null;
+        const generalMediafileId = this.getGlobalMediafileIdByToken(place);
```
jsangmeister, line 63: Move this into the if-else to only search for the token file if no meeting is active

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -68,10 +56,15 @@ export class MediaManageService {
     }
 
     public getLogoUrl(place: LogoPlace): string | null {
-        // Note: we are not fetching the mediafile view model at any place.
-        const mediafileId = this.activeMeetingService.meeting?.logo_id(place);
+        // Note: we are not fetching the mediafile view model at any place except when filtering for the defaults.
+        const mediafileId = this.activeMeetingService.meetingId
+            ? this.activeMeetingService.meeting?.logo_id(place)
+            : null;
```
jsangmeister, line 62:
```suggestion
        const mediafileId = this.activeMeetingService.meeting?.logo_id(place)
```

client/src/app/site/pages/meetings/services/media-manage.service.ts
```diff
@@ -117,6 +110,16 @@ export class MediaManageService {
         await this.mediaAdapter.unsetFont(place, this.activeMeetingService.meetingId!);
     }
 
+    private getGlobalMediafileIdByToken(place: LogoPlace | FontPlace): number {
+        return (
+            this.mediaRepo.getViewModelListUnsafe().find(file => {
+                return (
+                    file.isImage() && Object.keys(LogoDisplayNames).includes(place) && file.token === `global-${place}`
+                );
+            })?.id ?? null
```
jsangmeister, line 119: I think the `?? null` is redundant

client/src/app/site/pages/organization/pages/mediafiles/modules/organization-mediafile-list/components/organization-mediafile-list/organization-mediafile-list.component.ts
```diff
@@ -114,6 +123,42 @@ export class OrganizationMediafileListComponent
         this.cd.detach();
     }
 
+    public isMediafileUsed(file: ViewMediafile, place: LogoPlace): boolean {
+        const mediafile = this.repo.getViewModel(file.id)!;
+        if (file.isImage() && !Object.keys(LogoDisplayNames).includes(place)) {
+            return false;
+        }
+        return mediafile.token === `global-${place}`;
```
jsangmeister, line 131:
```suggestion
        return mediafile.token === `${place}`;
```

client/src/app/site/pages/organization/pages/mediafiles/modules/organization-mediafile-list/components/organization-mediafile-list/organization-mediafile-list.component.ts
```diff
@@ -114,6 +123,42 @@ export class OrganizationMediafileListComponent
...
+
+    public async toggleMediafileUsage(event: Event, mediafile: ViewMediafile, place: LogoPlace): Promise<void> {
+        // prohibits automatic closing
+        event.stopPropagation();
+        this.updatingLogoAndFontSettings = true;
+        const file = this.repo.getViewModel(mediafile.id);
+        if (!file || (file.isImage() && !Object.keys(LogoDisplayNames).includes(place))) {
+            throw new Error(!file ? `File has been deleted` : `Invalid mediafile type for place.`);
+        }
+        const fullPlace = `global-${place}`;
```
jsangmeister, line 142: remove

client/src/app/site/pages/organization/pages/mediafiles/modules/organization-mediafile-list/components/organization-mediafile-list/organization-mediafile-list.component.ts
```diff
@@ -114,6 +123,42 @@ export class OrganizationMediafileListComponent
...
+        if (!(file.token === fullPlace)) {
```
jsangmeister, line 143:
```suggestion
        if (file.token !== fullPlace) {
```

client/src/app/site/pages/organization/pages/mediafiles/modules/organization-mediafile-list/components/organization-mediafile-list/organization-mediafile-list.component.ts
```diff
@@ -114,6 +123,42 @@ export class OrganizationMediafileListComponent
...
+            for (let filteredFile of this.repo
+                .getViewModelList()
+                .filter(filterFile => filterFile.token === fullPlace && filterFile.id !== file.id)) {
+                await this.repo.update({ token: null }, filteredFile);
+            }
+            await this.repo.update({ token: fullPlace }, file);
```
jsangmeister, line 149: This should be done in one transaction in the backend. I will open an issue for this. For now, this can stay as-is.

client/src/app/site/pages/organization/pages/mediafiles/modules/organization-mediafile-list/components/organization-mediafile-list/organization-mediafile-list.component.ts
```diff
@@ -228,4 +273,9 @@ export class OrganizationMediafileListComponent
             this.directorySubscription = null;
         }
     }
+
+    private isUsedAs(file: ViewMediafile): boolean {
+        const places = this.logoPlaces;
+        return places[`global`].some(place => this.isMediafileUsed(file, place));
```
jsangmeister, line 279:
```suggestion
        return this.logoPlaces.some(place => this.isMediafileUsed(file, place));
```

jsangmeister (changes requested):

client/src/app/site/pages/meetings/services/media-manage.service.ts
luisa-beerboom: No, logos can still have a default

jsangmeister: Ah, the web_header is the default, right

client/src/app/domain/models/mediafiles/mediafile.constants.ts
```diff
@@ -19,16 +25,26 @@ export const LogoDisplayNames: { [place in LogoPlace]: string } = {
     pdf_footer_r: _(`PDF footer logo (right)`),
     pdf_ballot_paper: _(`PDF ballot paper logo`)
 };
+/**
+ * All different places where logos can be placed.
+ */
+export const LogoPlaces = Array.from(PossibleLogoPlaces.values());
```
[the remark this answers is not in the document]
luisa-beerboom: Generating an Array from the keys of `LogoDisplayNames` will just generate a string array (which I guess you could cast to be an `LogoPlace` array, but if you're going to have to process something anyway why not  generate it from the original data to begin with?)

`PossibleLogoPlaces` is cast as a `const ` so that I can use it to generate the Type. It will not be recognized as an array and therefore can't be used in places where the requirement is to have `(LogoPlace)[]`.
So I have to somehow generate an array from that => `LogoPlaces`

I could try to somehow generate `PossibleLogoPlaces` from the keys of `LogoDisplayNames`, but I believe that would mean that `LogoDisplayNames` isn't explicitly tied to the type any more, which also isn't very nice.

That is what led to this version,but  I can change it if it really bothers you.
What would you prefer instead?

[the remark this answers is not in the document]
jsangmeister: I would prefer only one place to gather the values and display names:
```
export const LogoPlaces = {
    projector_main: _(`Projector logo`),
    projector_header: _(`Projector header image`),
    web_header: _(`Web interface header logo`),
    pdf_header_l: _(`PDF header logo (left)`),
    pdf_header_r: _(`PDF header logo (right)`),
    pdf_footer_l: _(`PDF footer logo (left)`),
    pdf_footer_r: _(`PDF footer logo (right)`),
    pdf_ballot_paper: _(`PDF ballot paper logo`)
};
export type LogoPlace = keyof typeof LogoPlaces;
```
This way, there is no redundancy and one cannot e.g. add a new place without adding a display name for it. You can use `LogoPlace` just the way before and it should simplify the typing in using classes. What do you think?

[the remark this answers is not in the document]
luisa-beerboom: I can try it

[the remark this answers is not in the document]
jsangmeister: You could also leave the name `LogoDisplayNames` and add a `LogoPlaces = Object.keys(LogoDisplayNames)` if you don't want to write that everytime you access the keys. Also, what's the naming scheme in the client? Should `LogoPlaces` be called 'LOGO_PLACES` since it's a constant and not a type?

jsangmeister (approved):

emanuelschuetze (changes requested):
* set custom global logo
* open meeting and upload/set custom meeting logo
* go back to calendar -> OS logo

emanuelschuetze (changes requested):
Works good now. But currently, only a superadmin can upload files on orga level. You have to allow orga admins to see new files main menu entry.

emanuelschuetze (approved):

jsangmeister (approved):

PR #1897, opened 2023-01-10
participant import - use only username for duplicate check if given

### LINKED ISSUE #1896: Participant import duplicate check false positive if email already used
When trying to import a participant with an not yet used username but an email address already used within that meeting they are currently marked as duplicate.

bastianjoel:
resolves #1896 

If a username is given on participant import the email will now be excluded from the search for duplicates.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
@@ -127,7 +127,8 @@ export class ParticipantImportService extends BaseUserImportService {
     }): Promise<ImportModel<User>> {
         const username = input.username ? input.username : `${input.first_name} ${input.last_name}`;
         const userEmailUsername = `${username}/${input.email}`;
-        const duplicates = this._existingUserMap[userEmailUsername] ?? [];
+        const userUsername = `${username}/`;
+        const duplicates = this._existingUserMap[userUsername] ?? this._existingUserMap[userEmailUsername] ?? [];
         const newEntry = duplicates.length === 1 ? { ...duplicates[0], ...input } : input;
         const hasDuplicates =
             duplicates.length > 1 ||
--- a/client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
@@ -144,8 +144,7 @@ export class ParticipantImportService extends BaseUserImportService {
                     return { username: entry.username };
                 }
 
-                const username = !!entry.username ? entry.username : `${entry.first_name} ${entry.last_name}`;
-                return { username, email: entry.email };
+                return { username: `${entry.first_name} ${entry.last_name}`, email: entry.email };
             })
         });
         return result;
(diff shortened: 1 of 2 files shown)


client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
```diff
@@ -140,6 +140,10 @@ export class ParticipantImportService extends BaseUserImportService {
         const result = await this.presenter.call({
             permissionRelatedId: this.activeMeetingId,
             searchCriteria: entries.map(entry => {
+                if (entry.username) {
+                    return { username: entry.username };
+                }
+
                 const username = !!entry.username ? entry.username : `${entry.first_name} ${entry.last_name}`;
                 return { username, email: entry.email };
```
luisa-beerboom, line 148: Could probably remove the username from this second return statement then. It will be undefined anyway..

jsangmeister (changes requested):
When username and email are given, as I understand it, both should match to correctly identify the user as a duplicate. This is ignored by this change.

I think we should rather adjust the presenter to search for users matching username AND email, if both are given. This will require changing the usage of the presenter slightly in other places. If the current payload is `{search: [{username, email}]}`, then after the change one can achieve a similar result with the payload `{search: [{username}, {email}]}`, only the result parsing has to be adjusted since both entries are now returned as different result entries.

Alternatively, we could introduce a flag like `match_both` or `exact` per search entry to indicate if they should both match or only one of them.

luisa-beerboom (approved):
One minor proposal, otherwise it looks fine

bastianjoel:
> When username and email are given, as I understand it, both should match to correctly identify the user as a duplicate. This is ignored by this change.

I don't see a benefit in checking both for an exact match because usernames need to be unique anyway. Am I missing something?

jsangmeister:
> I don't see a benefit in checking both for an exact match because usernames need to be unique anyway. Am I missing something?

No, I had it twisted, I was thinking about the meeting import where username and email could be used to identify a user. For participant import, this should be fine. then only @luisa-beerboom's comment stands to be changed.

client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
```diff
@@ -140,8 +140,11 @@ export class ParticipantImportService extends BaseUserImportService {
         const result = await this.presenter.call({
             permissionRelatedId: this.activeMeetingId,
             searchCriteria: entries.map(entry => {
-                const username = !!entry.username ? entry.username : `${entry.first_name} ${entry.last_name}`;
-                return { username, email: entry.email };
+                if (entry.username) {
+                    return { username: entry.username };
+                }
+
+                return { username: `${entry.first_name} ${entry.last_name}`, email: entry.email };
```
jsangmeister, line 147:
```suggestion
                return { username: `${entry.first_name}${entry.last_name}`, email: entry.email };
```

Since https://github.com/OpenSlides/openslides-backend/pull/1458, usernames are generated without a space in them, so the search should not include it either.

jsangmeister (changes requested):

luisa-beerboom (approved):

jsangmeister (approved):

luisa-beerboom (approved):

emanuelschuetze (changes requested):
Matching does not work correctly:
- Create new account: "Max Mustermann", max@example.com, username: max
- Create new meeting
- Go to participants import view
- Download csv example -> delete all entries and add one line with:
firstname: Max, lastname: Mustermann, mail: max@example.com and username: max
- Now select this csv file for import
-> preview shows, that user would be created (instead of referenced)
- click "Import"
=> Error by user.create request: "Fehler: A user with the username max already exists."
Matching was successfully by search_users_by_name_or_email request (exactly one result), but client does not show "merge" icon to update user.

luisa-beerboom (approved):

PR #1902, opened 2023-01-11
Projector button rework

### LINKED ISSUE #607: Make the ProjectorButtonComponent more re-usable and unify the "projector-buttons" in the ProjectorDetailComponent
The ProjectorButtonComponent can be slightly enhanced:

- Use `<ng-content />` instead of `@Input() text: string`
- Use an `<ng-template />` tag instead of using the same button three times

Also, the ProjectorDetailComponent uses three buttons, that are working similar to the ProjectorButtonComponent, but are just buttons instead of ProjectorButtonComponents.

### LINKED ISSUE #1923: Projector detail component: Hide useless projector buttons
The projector detail component has a current LoS section where one can decide to project the current LoS as an overlay, a chyron or a slide.

The option slide necessarily doesn't work with the standard projector, since the current LoS is defined as "The corresponding LoS to whichever model is currently being projected on the standard projector". A LoS on the standard projector will therefore do nothing but try to find a LoS concerning _itself_ (which doesn't exist).

That's why the operator shouldn't even get the option to project the current LoS on the standard projector.

I.e.: IF the projector detail view is showing the current standard projector HIDE the LoS slide projection button

### LINKED ISSUE #1924: Projector detail component: Easy access to current list of speakers
There is no "easy" way to project a LoS as a slide in the Projector detail component. As a user you want to have following behavior:

I can press a button that toggles between the projection of the current  model and the list of speakers of this model. A possibly activated overlay or[ lower third](https://en.wikipedia.org/wiki/Lower_third)  should be also toggled. 

See also  #1923

luisa-beerboom:
Closes #607 
Closes #1923
Closes #1924

### DIFF
--- a/client/src/app/site/pages/meetings/pages/projectors/modules/projector-detail/components/projector-detail/projector-detail.component.ts
+++ b/client/src/app/site/pages/meetings/pages/projectors/modules/projector-detail/components/projector-detail/projector-detail.component.ts
@@ -6,6 +6,7 @@ import { BehaviorSubject, Observable, switchMap } from 'rxjs';
 import { Id } from 'src/app/domain/definitions/key-types';
 import { Permission } from 'src/app/domain/definitions/permission';
 import { ScrollScaleDirection } from 'src/app/gateways/repositories/projectors/projector.action';
+import { BaseViewModel } from 'src/app/site/base/base-view-model';
 import { BaseMeetingComponent } from 'src/app/site/pages/meetings/base/base-meeting.component';
 import { ViewProjection } from 'src/app/site/pages/meetings/pages/projectors';
 import { ProjectorControllerService } from 'src/app/site/pages/meetings/pages/projectors/services/projector-controller.service';
@@ -70,7 +71,7 @@ export class ProjectorDetailComponent extends BaseMeetingComponent implements On
     public editQueue = false;
 
     public get hasSlide(): boolean {
-        return !!this.projector.current_projections.filter(projection => projection.stable === false).length;
+        return !!this.projector.nonStableCurrentProjections;
     }
 
     public get currentProjectionIsLoS(): boolean {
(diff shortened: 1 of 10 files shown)


luisa-beerboom:
Found and fixed a bug concerning current list of speakers projection (which would throw a bunch of errors) otherwise the projector buttons (changes especially happened for the ones in the projector detail view and the menu of the LoS detail view) should all function as they did before

bastianjoel (approved):

bastianjoel (changes requested):
You could use `nonStableCurrentProjections` here and in the following occurrences of filtering non stable projections.

client/src/app/site/pages/meetings/pages/projectors/modules/projector-detail/components/projector-detail/projector-detail.component.ts
```diff
@@ -66,6 +69,21 @@ export class ProjectorDetailComponent extends BaseMeetingComponent implements On
      */
     public editQueue = false;
 
+    public get hasSlide(): boolean {
+        return !!this.projector.current_projections.filter(projection => projection.stable === false).length;
```
bastianjoel, line 73: You could use `this.projector.nonStableCurrentProjections` here and in the following occurrences of filtering non stable projections.

bastianjoel (changes requested):

client/src/app/site/pages/meetings/pages/projectors/modules/projector-detail/components/projector-detail/projector-detail.component.ts
```diff
@@ -66,6 +70,21 @@ export class ProjectorDetailComponent extends BaseMeetingComponent implements On
      */
     public editQueue = false;
 
+    public get hasSlide(): boolean {
+        return !!this.projector.nonStableCurrentProjections;
+    }
+
+    public get currentProjectionIsLoS(): boolean {
+        for (let projection of this.projector.current_projections.filter(projection => projection.stable === false)) {
```
bastianjoel, line 78:
```suggestion
        for (let projection of this.projector.nonStableCurrentProjections) {
```

client/src/app/site/pages/meetings/pages/projectors/modules/projector-detail/components/projector-detail/projector-detail.component.ts
```diff
@@ -201,10 +225,35 @@ export class ProjectorDetailComponent extends BaseMeetingComponent implements On
...
 
+    public getCurrentLoSBuildDesc(overlay: boolean): ProjectionBuildDescriptor {
+        return this.currentListOfSpeakersSlideService.getProjectionBuildDescriptor(overlay);
+    }
+
+    public getCurrentProjectionLoSToggleBuildDesc(): ProjectionBuildDescriptor | Projectable | null {
+        try {
+            for (let projection of this.projector.current_projections.filter(
+                projection => projection.stable === false
+            )) {
```
bastianjoel, line 236:
```suggestion
            for (let projection of this.projector.nonStableCurrentProjections) {
```

bastianjoel (changes requested):

bastianjoel (approved):

emanuelschuetze (approved):

PR #1906, opened 2023-01-12
Motion exports should now be using sequential numbers

### LINKED ISSUE #1905: Motion import uses id instead of sequential number

luisa-beerboom:
Closes #1905 

@emanuelschuetze please test thoroughly whether the different exports now function correctly in relation to ids and sequential numbers


bastianjoel (approved):

bastianjoel (approved):

emanuelschuetze (changes requested):
CSV and XLSX works (I added a xslx header fix for correct translation string.).
But PDF export _with_ squential numbers does not export numbers.

emanuelschuetze (approved):
Fixed pdf. Works now.

PR #1908, opened 2023-01-12
fix meeting list getting cleared

### LINKED ISSUE #1892: Committee-detail-view: Setting a meeting as a public template empties the views meeting list
Reproduction:
1. Set up a committee with two meetings
2. Use the menus of the views meeting list to set the first meeting to `public template` => The other meeting will vanish from the list
3. Reload => The meeting will be in the list again.
4. Reverse what you did in step 2 => Both meetings will vanish from the list

bastianjoel:
resolves #1892

Some relations currently have a isFullList property which indicates if a view model contains a complete list. When the property is set it seems to influence the behavior if a element is deleted on autoupdate.   

The problem is that it depends if a model is a full list on whether the user is inside a meeting or not. For the meeting list this is causing the elements to be deleted because their groups (which are marked as full list) are requested. 

~~This PR tries to fix that by giving the model request builder context on which model is the current root model.~~
This PR tries to fix that by adding a isFullList property to model requests which can overwrite the value specified for a relation.


client/src/app/site/services/model-request-builder/model-request-builder.service.ts
```diff
@@ -258,7 +265,7 @@ export class ModelRequestBuilderService {
     ): DescriptorResponse<RelationFieldDescriptor> {
         const foreignCollection = relation.foreignViewModel!.COLLECTION;
         const modelRequestObject = new ModelRequestObject(foreignCollection, follow, {});
-        if (relation.isFullList) {
+        if (relation.isFullList && relation.ownViewModels[0].name === this.rootViewModel) {
```
jsangmeister, line 268: Can you use `COLLECTION` instead of `name`? That seems a little less hacky.

jsangmeister (changes requested):
I don't think that this is the best way to solve this, but if it works it's fine since I don't understand the `isFullList` either :D

jsangmeister (approved):

client/src/app/site/services/model-request-builder/model-request-builder.service.ts
```diff
@@ -126,6 +129,10 @@ export class ModelRequestBuilderService {
         console.log(`apps loaded!`);
     }
 
+    public setRootViewModel(rootCollection: string) {
```
luisa-beerboom, line 132:
```suggestion
    public setRootViewModel(rootCollection: string): void {
```

luisa-beerboom (approved):
I don't really get the isFullList attribute either, but based on your explanation I think that, aside from one minor style thing, this looks good

luisa-beerboom:
@emanuelschuetze can you take a look if this also fixes issue #1892 ?
That one looks a bit like it may have the same cause.

bastianjoel:
> I don't think that this is the best way to solve this

It might be better to be able to specify and overwrite the isFullList property within the model request. What do you think about that @jsangmeister?

emanuelschuetze:
> @emanuelschuetze can you take a look if this also fixes issue #1929 ? That one looks a bit like it may have the same cause.

#1929 is not fixed with this PR: After deleting a meeting, the calender still shows the deleted meeting -> reload required.

emanuelschuetze (changes requested):
#1892 is fixed now, but similar issue #1929 is not fixed. Can you fix it in this PR?

luisa-beerboom:
I would propose to do #1929 separately if that is the case, let's merge this first

jsangmeister:
> > I don't think that this is the best way to solve this
> 
> It might be better to be able to specify and overwrite the isFullList property within the model request. What do you think about that @jsangmeister?

Sounds better! If it's not too much work, I would favor this approach.

bastianjoel:
@emanuelschuetze I think #1929 should be done within another PR too.

luisa-beerboom (approved):

emanuelschuetze (approved):

PR #1921, opened 2023-01-14
Cleanup

jsangmeister:
- remove unnecessary & inefficient shuffle function
- unify truthy env values with other services
- unify `BaseComponent` & `BaseUiComponent` (the latter provided a subset of the functionality of the former, which leads to the unfortunate class hierarchy where `BaseComponent` extends `BaseUiComponent`, but renaming them would not be worse the hassle, I think)
  - also moved some classes to inherit from `BaseComponent` instead of `BaseUiComponent` - this fixes some errors where subclasses used the e.g. `raiseError` function from `BaseUiComponent` which was previously the empty function (wtf?)
- unified & fixed timestamp handling in history list & motion list sort
- removed unused variables in `ViewOrganization`, `OrganizationService` and `SpinnerService`

### DIFF
--- a/client/src/app/infrastructure/utils/functions.ts
+++ b/client/src/app/infrastructure/utils/functions.ts
@@ -1,5 +1,4 @@
 import { Decimal } from '../../domain/definitions/key-types';
-import { deepCopy } from './transform-functions';
 
 export function toBase64(data: File | Blob): Promise<string> {
     return new Promise<string>(async (resolve, reject) => {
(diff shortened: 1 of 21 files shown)


luisa-beerboom (approved):

jsangmeister:
@emanuelschuetze @MSoeb can one of you test this? Since some base components were changed, it would be nice to check if any views were broken by this.

PR #1463, opened 2022-08-02
Various UI fixes

bastianjoel:
resolves #1462 
resolves #1331

PR #1931, opened 2023-01-18
fix os-banner causing extra scroll bar

### LINKED ISSUE #548: "Archived" bar creates a scroll bar
It seems like the "archived" bar is in the wrong container, so the whole page becomes scrollable with the bar.

bastianjoel:
resolves #548 

Fixes the mentioned scrollbar using the approach from #1463 but addresses the resulting issue with scroll lists (https://github.com/OpenSlides/openslides-client/pull/1463#issuecomment-1202593637). 

This PR changes the scrolling behavior of the scrolling table. The global headbar and page specific headbar gets hidden when scrolling.


emanuelschuetze (changes requested):
This breaks the csv import preview table: the vertical scrolling area of the preview table does not fit in the mat card...

emanuelschuetze (approved):

PR #1935, opened 2023-01-18
fix removal of single au connections

### LINKED ISSUE #1934: Motion won't load after creating two motions
When creating two motion after another the second one will not load anymore.

bastianjoel:
resolves #1934 

There was an error that single autoupdate connections would not be deleted from the connection pool after they were resolved. This resulted in positive duplicate checks and sending outdated data from the auotupdate service worker.


luisa-beerboom (approved):

emanuelschuetze (approved):
this works also for amendments. Thanks!

PR #1936, opened 2023-01-18
Improved participant edit view permission checks

### LINKED ISSUE #1932: Client sends group A fields in participant user update, even if the scope forbids it
Reproduction:
1. Create user that is in two meetings
2. As meeting admin of only one of these two meetings:
    1. Open the meeting-internal participant-edit view for this user
    2. First problem: The orga-level inputs aren't greyed out
    3. Try to change something (it doesn't have to be an orga field)
    4. Second problem: Missing OML permission error

Apparently the client is sending all applicable pieces of model-data, even those that haven't been changed at all or those that he shouldn't send due to missing perms.
The client also keeps sending user.set_present requests along with the update, even if presence hasn't been changed

luisa-beerboom:
Closes #1932 

Haven't found the time to filter out unchanged input values BUT the edit bug _should_ be fixed now!

### DIFF
--- a/client/src/app/gateways/presenter/get-user-scope-presenter.service.ts
+++ b/client/src/app/gateways/presenter/get-user-scope-presenter.service.ts
@@ -4,7 +4,7 @@ import { Id } from 'src/app/domain/definitions/key-types';
 import { Presenter } from './presenter';
 import { PresenterService } from './presenter.service';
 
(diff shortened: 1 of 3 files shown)


emanuelschuetze (approved):
Successfully tested!

rrenkert (approved):

PR #1956, opened 2023-01-24
Implemented account list active meeting filter + Orga statistics rework

### LINKED ISSUE #1953: Display active accounts in active meetings
In the Legal notice / Privacy policy / Info view there is (at least in the organization) a display for the number of active accounts/meetings. It only shows if there is a limit for active accounts/meetings.

- [x] Show a version of these displays that lacks the progress bar and the limit (i.e. one that only shows the absolute numbers) in case there is no limit

Also either:
- [ ] Show a new display that shows the number of active accounts that are in active meetings, or better:
- [x] Expand the account list so that it has a filter for accounts that are in active meetings. (Perhaps orga-admin-exclusive?)

luisa-beerboom:
Closes #1953

### DIFF
--- a/client/src/app/site/pages/meetings/view-models/view-user.ts
+++ b/client/src/app/site/pages/meetings/view-models/view-user.ts
@@ -85,6 +85,10 @@ export class ViewUser extends BaseProjectableViewModel<User> /* implements Searc
         return this.user.meeting_ids || [];
     }
 
+    public get isInActiveMeeting(): boolean {
+        return this.meetings.some(meeting => meeting.isActive);
+    }
+
     // Will be set by the repository
     public getName!: () => string;
     public getShortName!: () => string;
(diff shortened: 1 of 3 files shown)


bastianjoel (approved):

emanuelschuetze (changes requested):
A user who is in two meeting (active + archived) was not filtered correctly if filter for archived meeting is seleted -> use is not displayed! (works for filter for active meeting)

luisa-beerboom:
> A user who is in two meeting (active + archived) was not filtered correctly if filter for archived meeting is seleted -> use is not displayed! (works for filter for active meeting)

That is unsurprising since the filter only differentiates between users which have active meetings and those that don't have any active meetings at all.

There is currently no "is in archived meetings" filter.
Since this seems to be a functional requirement I will build one in, it will be a separate filter though.

bastianjoel (approved):

emanuelschuetze (approved):
good performance (tested with 10k accounts)

PR #1957, opened 2023-01-24
chip select component

### LINKED ISSUE #530: Own component for mat-chips which offer select-options
We use mat chips with drop downs on regular basis. Sometimes they offer a "clickable" (hand) indicator, sometimes they are just clickable and dont indicate their status.
We should use an own component for these small mat-chips with select-options.

We use them, i.e, as assignment phase indicator, motion state, motion motion category, motion rec (always motion detail)

bastianjoel:
resolves #530


bastianjoel:
Could also be used in `assignment-detail.component.html`

luisa-beerboom (approved):

emanuelschuetze (approved):

PR #1964, opened 2023-01-25
Repaired getUploadFileJson function error catching

### LINKED ISSUE #1962: `json-import-file-utils.ts` catch block broken
There are calls of `this.translate.instant` in the catch block of the `json-import-file-utils.ts`'s exported `getUploadFileJson` function that cause a very unhelpful `TypeError: this is undefined` to be thrown instead of the intended more helpful error message `Import data needs to have the JSON format`.

This causes problems in the meeting import as the user won't be able to see why the import isn't working, worst case they won't even realize it isn't working because there is no snackbar and the progress bar just keeps going.

So:
- [ ] Fix the catch block so that it will throw the correct error

luisa-beerboom:
Closes #1962 

Meeting and workflow imports should now throw a sensible error message if the json is not formatted correctly.

To test:
1. Take a json file that should  work
2. Break it (f.e. by adding a `,` after the last element of any array that is included in it)
3. When using it in the user import, pressing the upload button should throw an error "Import data needs to have the JSON format" along with showing with an equivalent snackbar message

### DIFF
--- a/client/src/app/infrastructure/utils/import/json-import-file-utils.ts
+++ b/client/src/app/infrastructure/utils/import/json-import-file-utils.ts
@@ -1,3 +1,4 @@
+import { Injectable } from '@angular/core';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 import { TranslateService } from '@ngx-translate/core';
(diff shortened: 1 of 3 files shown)


jsangmeister (changes requested):

client/src/app/site/pages/meetings/pages/motions/pages/workflows/components/workflow-import/workflow-import.component.ts
```diff
@@ -17,7 +17,8 @@ export class WorkflowImportComponent {
         private repo: MotionWorkflowControllerService,
         private location: Location,
         private snackbar: MatSnackBar,
-        private translate: TranslateService
+        private translate: TranslateService,
```
jsangmeister, line 20: You can remove these dependencies here

jsangmeister (changes requested):
Looks much better!

client/src/app/infrastructure/utils/import/json-import-file-utils.ts
```diff
@@ -1,10 +1,45 @@
...
+                    throw new Error(WRONG_JSON_IMPORT_FORMAT_ERROR_MSG);
+                }
+                resolve(result);
+            });
+            reader.readAsText(file.mediafile);
+        });
+        return json;
+    }
+}
+export async function getUploadFileJson<ExpectType>(
```
bastianjoel, line 38: It seems to me like this function can be completely removed. I couldn't find any more references to it.

bastianjoel (changes requested):

jsangmeister (approved):

emanuelschuetze (approved):

PR #1976, opened 2023-01-27
add validator for organization reply address

### LINKED ISSUE #1969: Trim whitespace in email fields
Whitespaces in email addresses should be trimmed.
In some places this is already done, but we should unify this for all email inputs.
Specially for the reply address in email settings it is missing and no mail can be sent at all if the address contains a leading or trailing whitespace.

bastianjoel:
resolves #1969

### DIFF
--- a/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.html
+++ b/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.html
@@ -50,6 +50,16 @@
 
                 <ng-template #input ngProjectAs="[matInput]">
                     <input
+                        *ngIf="setting.type === 'email'"
+                        matInput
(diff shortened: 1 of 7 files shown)


luisa-beerboom (commented):
Adding an email validator is a good idea in general (honestly I wonder why that wasn't done from the start), but I think that we may still want to trim white spaces to counteract a certain mobile bug that causes white spaces to be added behind the input when copy/pasting into a field

bastianjoel:
I've added a directive that trims the pasted input.

emanuelschuetze (changes requested):
I can still add blanks in meeting reply address (meeting participants settings), e.g. "info@example.com ".
`[{"action":"meeting.update","data":[{"users_email_replyto":"info@example.com ","id":126}]}]`

bastianjoel:
I can't replicate that. When I'm adding a space at the end of the email address a error message is shown below the input and I can't save.

emanuelschuetze (changes requested):
You fixed the orga settings component only! My test was in meeting participant settings. Please fix also this email input field.

luisa-beerboom (approved):

emanuelschuetze (approved):

PR #1977, opened 2023-01-27
fix empty scrolling tables and scrolling tables with a lot of data

bastianjoel:
When there were no elements within a scrolling table the search bar was scrollable. 

Also when there is too much data there were issues with the cdk scroll view. For that reason the height is now manually calculated by the dom element positions. 
This also reverts the changed behavior of these tables introduced by #1931.


luisa-beerboom (approved):

emanuelschuetze (changes requested):
Headbar of e.g. motion table still scrollable

emanuelschuetze (approved):

PR #1980, opened 2023-01-27
Clean up translations & introduce ngx-translate-extract fork

jsangmeister:
- replaced `ngx-translate-extract` with own fork, based on the previous one (https://github.com/bartholomej/ngx-translate-extract/tree/ng13, which was already an angular13-compatible fork of https://github.com/biesbjerg/ngx-translate-extract, since this seems to have become stale)
  - see https://github.com/bartholomej/ngx-translate-extract/compare/ng13...jsangmeister:ngx-translate-extract:openslides?expand=1 for the applied changes, if someone wants to review it. `dist` directory was pushed to enable direct installability via npm and can be skipped while reviewing
  - installed git in container to install directly from github
  - with the new versions, some more strings are extracted which 
- cleaned up `marked-translations.ts`
  - removed duplicate entries with new extraction fork
  - removed deprecated entries, e.g. from OS3
- added marker function to a lot of strings to ensure extractability
- fixed file generation scripts, cleaned up scripts in general

@emanuelschuetze there are now more untranslated strings than before, I guess they were just never reported as errors in the client. They have to be translated accordingly.


emanuelschuetze (changes requested):
as discussed, please use OpenSlides repo instead your personal repo

jsangmeister:
@emanuelschuetze done, but the name of the package still has the `@jsangmeister` prefix since I'm not sure if using an organization prefix is a common practice there. Should I change it there, too? Also, I could publish the package on npm so we don't have to manually install ist via github anymore if we want that.

luisa-beerboom (approved):

bastianjoel (approved):

emanuelschuetze (approved):

PR #1985, opened 2023-01-30
Rework permission matrix

jsangmeister:
test in conjunction with https://github.com/OpenSlides/openslides-backend/pull/1610

Permission matrix now has an explicit "Save"-Button and all checkboxes are clickable and the resulting permissions are calculated correctly.

- fully reworked the `generate-permission` script to also produce permission parents (could have theoretically also calculated them from the children, but IMO this would have been unnecessary overhead. Also, I wasn't really happy with the structure of the script anyway, I hope the code is clearer now)
- added group bulk updated and updated backend (https://github.com/OpenSlides/openslides-backend/pull/1610) to be able to send all changes at once to the backend (without having to transform them into the format of the `set_permission` action first, which would have been tedious and unnecessary work)
- added `Set.update` to the prototype (why is the javascript set so limited?...)
- removed obsolete methods


jsangmeister:
I noticed that when you enter the permission matrix form the participant list the client becomes unresponsive for about a second. The behaviour was already present before these changes, even though the effect appears to be stronger on my local instance then on the demo. I could not find the root of this behaviour quickly, so I left it as-is, but maybe we want to open an issue for this and investigate it further.

luisa-beerboom (approved):

luisa-beerboom (approved):
Linting fails

emanuelschuetze (approved):
works good now!

PR #1988, opened 2023-01-31
fix mass deletion selection reset

### LINKED ISSUE #1973: Mass deletion doesn't clear list selection count
Reproduction:
1. Create 10 users
2. Go to the account list
3. Enter multiselect mode
4. Select all users that were created through step 1
5. The list will show "10 selected" next to the arrow
6. Delete them all via the multiselect menu
7. The list will still show "10 selected" next to the arrow

This is probably an error with the list component in general.
Please fix it and ensure that the apparently remaining legacy selections do not break any further use of the multiselect.

bastianjoel:
resolves #1973
This PR adds a unfiltered dataSource to the scrolling tables for checking whether a entry needs to be deleted from selection. 

This reverts "keep selection in scroll tables if data filtered (#1581)"

### DIFF
--- a/client/src/app/ui/modules/list/components/view-list/view-list.component.html
+++ b/client/src/app/ui/modules/list/components/view-list/view-list.component.html
@@ -18,6 +18,7 @@
 
         <os-scrolling-table
             [dataSource]="dataListObservable"
+            [fullDataSource]="dataFullListObservable"
             [rowHeight]="vScrollFixed"
             [hiddenColumns]="hiddenColumns"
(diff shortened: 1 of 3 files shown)


luisa-beerboom (approved):
Style-wise this looks good to me. Since you said that #1581 is reverted, I have to wonder though: Does this re-implement the functionality or will we have to re-open that issue?

bastianjoel:
#1581 was a bit of a workaround. This fixes also that original issue in a more clean way.

emanuelschuetze (approved):

PR #1991, opened 2023-01-31
Remove tag_ids from payload for amendments.

### LINKED ISSUE #1990: Delegate can't create amendment if the motion has a tag
**Description**: If a delegate (or other user groups that have the permission to create amendments) wants to create an amendment to a motion with tags, this is not possible. An error message appears. The problem also occurs, if the motion had a tag which was deleted. After a tag is set to a motion, it's not possible to create an amendment anymore.

Screenshot_20230131_163604

**Reproduction:**
1. open a meeting as superadmin
2. open motions
3. create a tag in motions
4. create a motion and set the tag to it.
5. grant delegates the permission to create amendments in the group permission matrix
6. go to particpiants and create a new user with the group delegate
7. open the meeting in another browser and log in as the new user (delegate)
8. open motions and the motion
9. create a new amendment via plus icon
10. error message appears. 

**What should happen:** Amendment creating should not interact with tags. Delegates or other groups should be able to create amendments that have or have not tags.

rrenkert:
closes #1990


emanuelschuetze:
I've to merge it now. @luisa-beerboom: Feel free to review later.

PR #1999, opened 2023-02-02
[For testing] Moved poll.vote_count into own subscription

luisa-beerboom:
The field will now only be requested by people who can see it.
There will be a certain interval wherein the progress bar will always show "0/x", even if people have already voted.

The autoupdate-worker sometimes seems to be bundling this new subscription with the `participant_is_present_list` subscription. I think this bundling happens only in the client though, it should therefore (hopefully) have no bearing on the results.

Furthermore:
The new subscription is currently requested in the `operator.service` right after the new permissions for the meeting are calculated. If this new subscription is found to be helpful we might want to rethink whether the placement of the request code is appropriate before merging.


rrenkert:
Done some perfromance tests and it seems like there is no difference to the subscription including poll.vote_count for delegates.
Will perform some more tests...

PR #2009, opened 2023-02-06
refactored html-to-pdf services

### LINKED ISSUE #2004: Refactor `HtmlToPdfService`s
The client currently has two services `HtmlToPdfService` and `MotionHtmlToPdfService`.

The fact that the `HtmlToPdfService` is basically a copy of the `MotionHtmlToPdfService` with some motion-specific functionalities removed, quickly becomes obvious as soon as one starts comparing the two.

There is a LOT of boilerplate code and extremely long functions here!

TODO:

- [ ] Make code compliant with the DRY principle, one could f.E:
    - turn the `MotionHtmlToPdfService` into a subclass of the `HtmlToPdfService`, or 
    - create an abstract base class that is extended by both of them
- [ ] Refactor extremely long functions into multiple smaller functions (may also help with re-using code)

luisa-beerboom:
Closes #2004

Also moved some more general code from the motion-html-to-pdf.service to the html-to-pdf.service

### DIFF
--- a/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
+++ b/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
@@ -34,7 +34,7 @@ export class HtmlToPdfService {
     /**
      * Normal line height for paragraphs
      */
-    private lineHeight = 1.25;
+    protected lineHeight = 1.25;
 
     /**
      * space between paragraphs
--- a/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
+++ b/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
@@ -1,6 +1,20 @@
 import { Injectable } from '@angular/core';
 import { ExportServiceModule } from 'src/app/gateways/export';
 
+export interface ChildNodeParagraphPayload {
+    child: Element;
+    parent: Element;
+    styles?: string[]
+}
+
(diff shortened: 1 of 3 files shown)


client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
```diff
@@ -21,17 +34,17 @@ export class HtmlToPdfService {
     /**
      * Normal line height for paragraphs
      */
-    private LINE_HEIGHT = 1.25;
+    protected LINE_HEIGHT = 1.25;
```
bastianjoel, line 37: `LINE_HEIGHT` can be overwritten within the `convertHtml` method. I would only expect constant values to be uppercase.

client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
```diff
@@ -21,17 +34,17 @@ export class HtmlToPdfService {
...
      * space between paragraphs
      */
-    private P_MARGIN_BOTTOM = 4.0;
+    protected P_MARGIN_BOTTOM = 4.0;
 
     /**
      * Space above H
      */
-    private H_MARGIN_TOP = 10.0;
+    protected H_MARGIN_TOP = 10.0;
```
bastianjoel, line 47: These properties could be `readonly`.

bastianjoel (changes requested):

bastianjoel (approved):

PR #2035, opened 2023-02-15
Update jitsi API and config

### LINKED ISSUE #1872: Can't connect to conference rooms that use newer versions of jitsi
Maybe check for updates with jitsi library

m-schieder:
Fix #1872


emanuelschuetze (approved):
tested with webconf-staging, webconf and meet.jit.si

Issue #2032, opened 2023-02-15
Feature: enhance pdf export of agenda


MSoeb:
**Description:** 
Currently the pdf export file of the agenda is quite rudimentary and miss some content. For example, it does not include comments or polls. It would be nice to enhance it. 

**Feature description:** 
It would be nice to have an export dialog field just like the one in motions. So the user could decide, what should be exported. 
The dialog field should have options for at least title, agenda number, comment, duration, polls and content of an agenda topic. The fields should be de-/selectable.

bastianjoel:
Agenda poll export is tracked by #2030

emanuelschuetze:
comment from #2038: PDF export works good now for single pdf which can be generated from topic detail view.
It would be great to export multiple topic items in one pdf (via multiaction) OR/AND via an agenda export dialog for all agenda items

MSoeb:
Issue will be solved by #5060 and can therefor be closed.

PR #2038, opened 2023-02-17
agenda item poll PDF export

### LINKED ISSUE #2030: Feature: Export option for agenda polls
Description: Currently it is not possible to generate a pdf export file for agenda polls. This is not coherent with motion votes and elections.

What should happen:
From my perspective agenda polls should have a pdf export option, just like the ones in motion votes or elections. So the users could generate protocols of them.

bastianjoel:
resolves #2030 

- [x] Agenda export
- [x] Ballot paper export


client/src/app/site/pages/meetings/pages/agenda/modules/topics/services/topic-pdf.service/topic-pdf.service.ts
```diff
@@ -0,0 +1,183 @@
...
+                    margin: [0, 15, 0, 0]
+                });
+
+                pollTableBody.push([
+                    {
+                        text: ``,
+                        style: `tableHeader`
+                    },
+                    {
+                        text: this.translate.instant(`Candidates`),
```
luisa-beerboom, line 116: `Candidates` in a topic poll? Topic polls have text options so it should just be `Options`

luisa-beerboom (changes requested):

luisa-beerboom (approved):

emanuelschuetze:
Works good as single pdf which can be generated from topic detail view.
It would be great to export multiple topic items in one pdf (via multiaction) OR/AND via an agenda export dialog for all agenda items (this could be a bit more work).

bastianjoel:
> Works good as single pdf which can be generated from topic detail view. It would be great to export multiple topic items in one pdf (via multiaction) OR/AND via an agenda export dialog for all agenda items (this could be a bit more work).

There is already another issue for that #2032

emanuelschuetze (approved):

PR #2050, opened 2023-02-21
add ability to disable autoupdate compression

### LINKED ISSUE #2044: Disable compression for debugging on productive instances
Currently, the client connects to the autoupdate-service with `compress=1`. In this mode, the autoupdate-service sends each line of an autoupdate-connection with compression.

For debugging the connection, it would be nice, if this could be disabled.

Is it possible to add an option in the client to disable this? It don's not have to be nice. Fore example, the client could not use compression, if there is a key in the localstorage (that I could manually set).

Would it be hard to implement this?

bastianjoel:
resolves #2044 

This adds a method `disableAutoupdateCompression()` to the window object to disable the compression of autoupdate connections.
The method closes all currently open autoupdate connections and reopens them without the `compress=1` parameter. Also all new autoupdate streams will be opened without the `compress=1` parameter after the method has been called. 

Besides that it is now possible to use the worker helper methods (this includes `printAutoupdateState()`) in production by setting the value of the local storage key `DEBUG_MODE`. This can be done by calling `window.localStorage.setItem('DEBUG_MODE', 1);` from the dev tools console. 

Note that the util commands are only available within the shared worker.

### DIFF
--- a/client/src/app/site/services/autoupdate/autoupdate-communication.service.ts
+++ b/client/src/app/site/services/autoupdate/autoupdate-communication.service.ts
@@ -92,7 +92,20 @@ export class AutoupdateCommunicationService {
             } as AutoupdateAuthChange);
         });
 
+        if (window.localStorage.getItem(`DEBUG_MODE`)) {
+            this.enableDebugUtils();
+        }
+
         this.registerConnectionStatusListener();
+    }
+
+    /**
+     * Enable the debug utilities of the shared worker
+     */
+    public enableDebugUtils(): void {
+        this.sharedWorker.sendMessage(`autoupdate`, {
+            action: `enable-debug`
+        });
     }
 
     /**
--- a/client/src/app/worker/autoupdate-stream-pool.ts
+++ b/client/src/app/worker/autoupdate-stream-pool.ts
@@ -31,6 +31,7 @@ export class AutoupdateStreamPool {
     private _authTokenRefreshTimeout: any | null = null;
     private _updateAuthPromise: Promise<void> | undefined;
     private _waitingForUpdateAuthPromise: boolean = false;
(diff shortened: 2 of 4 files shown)


ostcar:
Thank you very much.

When your are at it: would it also be possible to get a debug function to get the id of a object by it's sequential_number?

For example 'window.getID(COLLECTION, SEQUENTIAL_NUMBER)' and getting the meeting id from the url.

ostcar:
@bastianjoel My javascript knowlage is to low. Can you give me a detailed explanation, how to use the `disableAutoupdateCompression` function? I thought, I just have to type the name in the console. But when I do this, I get the message: `Uncaught ReferenceError: disableAutoupdateCompression is not defined`

bastianjoel:
You need to run this within the SharedWorker. 
Detailed steps to access it in Chromium are:
1. Open OpenSlides 
1. Set the local storage entry
1. Make sure you don't have any other tabs with the OpenSlides instance open and reload
1. In another tab open `chrome://inspect`
1. Go to Shared workers 
1. There should be an openslides-shared-worker listed - click inspect
1. Within that worker window you can use the fubctions

ostcar:
It works. Thank you. Do you know, if there is a way to do the same thing in firefox?

ostcar (approved):

bastianjoel:
You can access the shared worker on `about:debugging` but Firefox currently does not support inspecting network connections there.

bastianjoel:
This is only needed when running in production or when the user explicitly enables the shared worker within the `environment.ts`. In the second case I think the user knows what they are doing. So I think this is not really development related but more for production debugging. 
Anyway I also think it might be a good idea to document not only that but also the functionality introduced by this PR somewhere.

PR #2053, opened 2023-02-22
fix point of order submitet on cancel

### LINKED ISSUE #2052: Point of orders: Cancel with written text is not possible
**Description:** 
The point of order button does not work correctly.  Once text is entered in the dialog box textfield, the point of order cannot be canceled. When pressing the Cancel button, the point of order request is created anyway. Only with an empty text field can the action be canceled. 

**Reproduction:**
1. open a meeting
2. open settings > list of speakers. Check if enable point of oders is active. 
3. open agenda, create a topic and project it
4. open the Autopilot
5. click point of order button, enter text and click cancel

**What should happen:**
By clicking on the Cancel-button, the action should be canceled.

bastianjoel:
resolves #2052


rrenkert (approved):
Not tested

peb-adr:
@MSoeb and me tested it. SeemsGood
I'll merge

PR #2057, opened 2023-02-23
Implemented client side list election polls

### LINKED ISSUE #1967: Implement client-side list election polls
This will require the changes in OpenSlides/openslides-backend#1608 and any eventual autoupdate changes.

****CREATION****

In the assignment-poll dialog there should be an option allowing the user to switch between normal assignment polls and **list-elections**. (Something like a checkbox maybe)
For now those should use only one list per poll.

If, upon confirmation, that checkbox is toggled, the poll creation must be changed in the following ways:

1. Send the appropriate payload for a single-option list election (i.e. fill in the new field in the option scheme, instead of content_object_id)
2. The rest proceeds as usual

****Displaying the poll****
The candidate list should be sorted according to weight, numbered and displayed above the voting area. For the sake of comfort it should be expandable (closed by default?).

When the list is closed, instead of the title we may want to display a preview of the contained entries

Everything else should (hopefully) work just fine with the current system.

luisa-beerboom:
Closes #1967 

poll-projection will work, but only in combination with OpenSlides/openslides-autoupdate-service#637

### DIFF
--- a/client/src/app/domain/models/poll/generic-poll.ts
+++ b/client/src/app/domain/models/poll/generic-poll.ts
@@ -1,7 +1,13 @@
 import { Observable } from 'rxjs';
-import { BaseViewModel } from 'src/app/site/base/base-view-model';
+import { ViewTopic } from 'src/app/site/pages/meetings/pages/agenda';
+import { ViewAssignment } from 'src/app/site/pages/meetings/pages/assignments';
+import { ViewMotion } from 'src/app/site/pages/meetings/pages/motions';
+import { ViewPollCandidateList } from 'src/app/site/pages/meetings/pages/polls/view-models/view-poll-candidate-list';
+import { ViewUser } from 'src/app/site/pages/meetings/view-models/view-user';
 
 import { EntitledUsersEntry, PollClassType, PollMethod, PollPercentBase, PollState, PollType } from './poll-constants';
+
+export type PollContentObject = ViewAssignment | ViewMotion | ViewTopic;
 
 /**
  * The main interface to describe everything related to polls
@@ -21,7 +27,7 @@ export interface PollData {
     options: OptionData[];
     options_as_observable: Observable<OptionData[]>;
     global_option: OptionData;
-    content_object?: BaseViewModel<any>;
+    content_object?: PollContentObject;
     getContentObjectTitle: () => string | null;
 }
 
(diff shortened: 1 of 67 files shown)


client/src/app/infrastructure/definitions/relations/relations.ts
```diff
@@ -894,12 +916,35 @@ export const RELATIONS: Relation[] = [
...
+        // ViewPollCandidateList -> ViewUser
+        {
+            ownViewModels: [ViewPollCandidateList],
+            foreignViewModel: [ViewOption],
+            ownField: `option`,
+            many: false,
+            generic: false,
+            structured: false
+        }
+    ] as Relation[]),
```
bastianjoel, line 947: The grouping into an array and spreading is not necessary.

bastianjoel (changes requested):

bastianjoel (approved):

emanuelschuetze:
@luisa-beerboom as discussed

client/src/app/domain/models/poll/generic-poll.ts
```diff
@@ -20,6 +21,7 @@ export interface PollData {
     options: OptionData[];
     options_as_observable: Observable<OptionData[]>;
     global_option: OptionData;
+    content_object?: BaseViewModel<any>;
```
bastianjoel, line 24: Is it possible to be more specific about the type here?

bastianjoel: Not completely sure but I think this could be
```suggestion
    content_object?: BaseViewModel<ViewUser | ViewPollCandidateList>;
```

bastianjoel (changes requested):

client/src/app/domain/models/poll/generic-poll.ts
luisa-beerboom: I agree that I may want to be a bit more specific here, but in this place it would be `content_object?: ViewMotion | ViewAssignment | ViewTopic`, and then in ln. 47 it should be `content_object?: ViewMotion | ViewUser | ViewPollCandidateList`
I'll look into it

bastianjoel (approved):

bastianjoel (approved):

emanuelschuetze (approved):
Looks good for me.
@MSoeb Please have a look

emanuelschuetze (approved):

PR #2063, opened 2023-02-27
fix poll valid votes font size

### LINKED ISSUE #2037: Font and icon size in election result view
The election result view card is a bit untidy: different font sizes and different icon sizes. In the mockup you'll see that all font sizes are the same, exept nicht-namentlich. All icons scale down to the font size. Furthermore, the card is less high. The mockup solution of the election result view card is more harmonic. 
electioresultharmony

bastianjoel:
resolves #2037

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.html
+++ b/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.html
@@ -39,7 +39,7 @@
                             <os-icon-container *ngIf="row.value[0].icon" [icon]="row.value[0].icon" [size]="iconSize">
                                 {{ row.votingOption | pollKeyVerbose | translate }}
                             </os-icon-container>
-                            <span *ngIf="!row.value[0].icon">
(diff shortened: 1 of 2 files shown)


client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.html
```diff
@@ -18,13 +18,13 @@
                         </td>
                         <td>
                             <div class="result-cell-definition">
-                                <span *ngIf="row.value[0].showPercent">
+                                <span *ngIf="row.value[0].showPercent" [class]="'text-' + iconSize">
```
luisa-beerboom, line 21: You've used this exact string concatenation a whole five times in this PR, why not write a getter for it?

luisa-beerboom (changes requested):

emanuelschuetze (changes requested):
Please reduce margin-left to get poll result left-justified with ballot title/subtitle

luisa-beerboom (approved):

bastianjoel:
The chart should have a padding so it is aligned with LOS

Issue #2002, opened 2023-02-02
Reduce `voted` calls


jsangmeister:
The `voted` route is called too often, although I did not investigate this in detail. But the route is called when updates come in which have nothing to do with the poll.

_Originally posted by @jsangmeister in https://github.com/OpenSlides/openslides-client/issues/1995#issuecomment-1412429531_

ostcar:
I also would like this solution, but it is hard to implement. The reason is, that even every user sees different content for the field. This is a requirement no other field from the auto update service has. So it does not fit in the current system. I also don't think we should change the autoupdate service for this one case.

I don't think we should use the icc service for that. It has a different purpose and should not be coublet with the vote service.

But it should be possible to create a long living url in the vote service. A user could send a request, maybe with a meeting id, and gets all poll Ids, where he can vote. (All started polls where he can vote for him self or for a delegate).

When a poll is changed (started, stopped, entitled groups, user gets present, etc.) an update is send in json line format. I think this should be possible.

jsangmeister:
Sounds also good. Could this have an impact on the performance of `poll.{start|stop}`? I imagine the vote service would have to iterate through all subscriptions of a poll's meeting upon starting/stopping it and send an update to all of them, which could be thousands. Since `poll.stop` is already a contender for a rather slow action, we should check this before just using that solution.

bastianjoel:
> Sounds also good. Could this have an impact on the performance of `poll.{start|stop}`? I imagine the vote service would have to iterate through all subscriptions of a poll's meeting upon starting/stopping it and send an update to all of them, which could be thousands. Since `poll.stop` is already a contender for a rather slow action, we should check this before just using that solution.

The subscription can be closed after a user has voted. For that reason I think this should not be an issue.

PR #2065, opened 2023-02-27
request has voted state for started polls only

bastianjoel:
This improves the behavior described in #2002. Also the "click here to vote" banner hides faster in the current tab with this changes. 
With this PR only running polls are requested from the vote service. 

To finish #2002 the voted service should return long polling responses. For now instead of requesting the voted status with totally unrelated actions the status is requested every 8-10 seconds. 

Besides that change the following changes are planned in this PR
- [x] Periodically check if a user has voted on a started poll
- [x] Do not display vote buttons when vote status not loaded
- [x] Display vote banner for delegated votes
- [x] Cleanup subscriptions for closed polls


luisa-beerboom (approved):

bastianjoel:
There is currently a console error when on organization view caused by the vote banner service.

rrenkert (approved):

PR #2077, opened 2023-03-07
run prettier within eslint

bastianjoel:
We are currently running `prettier` manually within pre defined folders. In the past and also currently this resulted in files missed by the prettier checks. Also the cleanup command breaks the output of the from within the script called commands which is quite annoying when working on linter errors. 

With this PR `eslint-plugin-prettier` is used to run prettier checks.


client/.eslintrc.js
```diff
@@ -77,6 +81,14 @@ module.exports = {
...
-        }
+        }/*,
+        {
+            files: ['*.html'],
+            excludedFiles: ['*inline-template-*.component.html'],
+            extends: ['plugin:prettier/recommended'],
+            rules: {
+                'prettier/prettier': ["error", { parser: 'angular' }]
+            }
+        }*/
```
jsangmeister, line 92: Please remove from this PR and either add a comment to the respective issue or open a PR for it

client/package.json
```diff
@@ -20,17 +20,17 @@
     "test-silently": "npm run test -- --watch=false --no-progress --browsers=jsdom",
     "test-live": "npm run test -- --watch=true --browsers=Chromium",
     "stats": "webpack-bundle-analyzer dist/client/stats.json",
-    "lint": "ng lint",
-    "lint-write": "ng lint --fix",
+    "lint": "ng lint && npm run prettify-check",
+    "lint-write": "npm run prettify-write && ng lint --fix",
```
jsangmeister, line 24: The order "linting -> formatting" was on purpose and should be kept: The automatic fixes from the linter are not always compatible with prettiert. So, linting has to run first, and prettier afterwards. To circumvent the issue that you can't see the output, the solution which you deleted was implemented (but maybe did not work correctly lately).

Also, I don't think `npm run lint` should execute the prettify commands.

jsangmeister (changes requested):

jsangmeister (approved):

PR #2101, opened 2023-03-16
Improve fieldset management

bastianjoel:
With this PR the default fieldsets will be moved to their models. 

Also a cli tool was added to automatically generate the default fieldsets and list missing properties defined by [models.yml](https://github.com/OpenSlides/openslides-backend/blob/main/global/meta/models.yml)

## Changes to `get{name}SubscriptionConfig` functions
The following changes will be made to all of these
- [x] `HidingConfig` will no longer be set there. When a subscription is not needed has nothing to do with the subscription but instead with its usage. 
- [x] Types will be added

## Review notes
The `DEFAULT_FIELDSET` properties were generated automatically and therefore do not need to be reviewed.


client/src/app/site/base/base-view-model.ts
```diff
@@ -9,6 +9,7 @@ import { BaseModel } from '../../domain/models/base/base-model';
 
 export interface ViewModelConstructor<T extends BaseViewModel> {
     COLLECTION: string;
+    DEFAULT_FIELDSET?: string[];
```
luisa-beerboom, line 12: I don't think you need this in both the models AND view-models

luisa-beerboom (changes requested):

client/src/app/site/base/base-view-model.ts
bastianjoel: I removed it from ViewModel but I still need this for models. I think the naming might be misleading as this property should just contain all fields one can request from autoupdate.

PR #2111, opened 2023-03-20
[WIP] use smaller fieldset

bastianjoel:


client/src/app/site/pages/meetings/services/active-meeting.subscription.ts
```diff
@@ -1,35 +1,52 @@
+import 'src/app/site/services/model-request-builder';
```
bastianjoel, line 1: Without this import I get an error in BaseComponent on runtime. I do not know why and did not put too much effort into it. We might want to open an issue for that.

PR #2115, opened 2023-03-23
Update translation loader

### LINKED ISSUE #2010: (German) Translation problem when sending multiple emails
The success snackbar message when multiple emails have been sent is `%num% emails were send successfully.`
The `%num%` is replaced with the corresponding number after the string has been translated, creating a new message `2 emails were send successfully`.

Problem:
The German translation for this is saved as `{0}um% E-Mails wurden erfolgreich versandt.`
This will not be recognized by the number replacement command, and therefore `{0}um% E-Mails wurden erfolgreich versandt.` will be exactly what the user sees.

Please fix this.

jsangmeister:
Fixes #2010 by skipping the unnecessary convertion to JSON and instead directly parsing the PO files. This will save us future analogous problems. The PO files are a little bigger (109KB vs 88 KB for the german translation), but still so small that the performance impact is irrelevant.

@bastianjoel @luisa-beerboom btw, you don't always have to both review my PRs (you can of course, if you want to) - I'm just never sure who to assign :D

And lastly, to resolve conflicts in `package-lock.json`, it should suffice to rerun `npm install`, or are there additional steps to take?


bastianjoel (approved):
Looks good but bundle maximum size is exceeded.

jsangmeister:
I can't seem to figure out why, though. `gettext-parser` has (according to `npm run stats`) only 15.96KB. Other PRs, e.g. https://github.com/OpenSlides/openslides-client/actions/runs/4479230574/jobs/7872940617?pr=2113, give the bundled size as 2.68MB vs 3.24MB in this PR, which is an increase of about 560KB (or about 20%). Do you have any idea where this comes from? If not, I'm inclined to just up the limit.

bastianjoel:
I think this might be caused by `gettext-parser` being a CommonJS dependency and requiring additional dependencies. Is there maybe another library?

bastianjoel:
`@types/gettext-parser` should be a dev dependecy as far as I know.

bastianjoel:
I replaced `gettext-parser` with `pofile` which should be much lighter.

bastianjoel:
closes #2131

PR #2123, opened 2023-03-24
hide created/modified date in motion list if not filtered by it

### LINKED ISSUE #2092: Show motion timestamp in list only if sort option is active
Hide timestampe by default.
Only if user sorts by creation/modification time show timestamp in list view

bastianjoel:
resolves #2092


emanuelschuetze (changes requested):
Please show created timestamp if sort by creation date

emanuelschuetze (approved):

rrenkert (approved):

PR #2131, opened 2023-03-27
Added po2json-tempfix

### LINKED ISSUE #2010: (German) Translation problem when sending multiple emails

emanuelschuetze:
Fixed #2010 

run through sed after autotranslating po2json


bastianjoel:
We might want to use #2115 instead

PR #2136, opened 2023-03-28
restart autoupdate connections on reload

### LINKED ISSUE #2127: Handle tab reload in shared worker
In different (e.g. laptop in sleep mode; network change; but mostly unreproducible...) scenarios browsers lose connection to the auto update service.
If multiple tabs are open in the browser the shared worker does not recognize the reload of a single tab.

To fix connection problems using reload in a single tab, the shared worker should do a reconnect for the current subscriptions if the tab reloads.

bastianjoel:
resolves #2127


rrenkert (approved):
Tested, works as expected.

PR #2142, opened 2023-03-29
fix motion justified text

### LINKED ISSUE #2135: Motion text editor: justification of text format didn't work
Description: Editing a text segment or the whole text of a motion in justification format doesn't work. The editing in the text editor works fine, but it isn't saved. The motion is still in the original text format. 

Text editor: text format changed to justify
Screenshot_20230328_163113

Motion view after save
Screenshot_20230328_163139

Reproduction:
1. create a motion with at least 3-4 lines.
2. edit the text format to justification and save it

What should happen:
Text format can be changed to justification (justify).

bastianjoel:
This PR is attempting to resolve #2135. 

While this fixes the motion text within the client this breaks the pdf export of justified text completely. When investigating this I was not able to find a way to manually add line breaks within justified text in pdfmake. 
For that reason it is possible that the issue is not fixable.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/line-numbering.service/line-numbering.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/line-numbering.service/line-numbering.service.ts
@@ -475,17 +475,21 @@ export class LineNumberingService {
      * @param {number} length
      * @param {number} highlight
      */
-    public textNodeToLines(node: Node, length: number, highlight: number | null = -1): Element[] {
+    public textNodeToLines(node: Node, length: number, highlight: number | null = -1, justify = false): Element[] {
         const out: any[] = [];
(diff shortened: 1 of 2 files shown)


bastianjoel:
The following results in a text with a line break after `sure`. This might be helpful for a workaround. 
```
text: [
		'Another paragraph, this time a little bit longer to make sure\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t',
		'Another paragraph, this time a little bit longer to make sure, this line will be divided into at least two lines',
]
```

PR #2158, opened 2023-03-31
Fixed motion list default sorting

### LINKED ISSUE #2145: Sort motions by call list (default does not work)
* add 3 motions A, B C
* open call list (3 dot menu) and sort e.g.: C, B, A
* open motion settings: set "sort motions by call list"
* now click "clear site data" (Chrome) to remove all local storage defaults for sorting
* open motion list => motion list is not sorted by call list (open sort menu -> nothing is checked)

sort motions by number works as expected

luisa-beerboom:
Closes #2145

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/services/list/motion-list-sort.service/motion-list-sort.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/services/list/motion-list-sort.service/motion-list-sort.service.ts
@@ -37,7 +37,7 @@ export class MotionListSortService extends BaseSortListService<ViewMotion> {
      * Define the sort options
      */
     protected motionSortOptions: OsSortingOption<ViewMotion>[] = [
-        { property: [`tree_weight`, `id`], label: `Call list` },
+        { property: `tree_weight`, label: `Call list` },
         { property: `number` },
         { property: `title` },
         { property: `submitters` },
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/pages/motions/services/list/motion-list-sort.service/motion-list-sort.service.ts
```diff
@@ -80,11 +85,19 @@ export class MotionListSortService extends BaseSortListService<ViewMotion> {
...
         return {
-            sortProperty: this.defaultMotionSorting as keyof ViewMotion,
+            sortProperty: this.getDefaultSortProperty(),
             sortAscending: true
         };
     }
 
+    private getDefaultSortProperty(): OsSortProperty<ViewMotion> {
+        if (this.defaultMotionSorting === `weight`) {
+            return [`tree_weight`, `id`];
```
jsangmeister, line 95: The `tree_weight` should always be unique I think, is the second sorting condition really necessary?

luisa-beerboom: Idk, I just applied the changes to the sort option properties from PR #1475 to the default properties, I thought maybe the `tree_weight` might not be set in certain conditions, I honestly found it stranger that the secondary sorting condition was `id` and not `sequential_number` or `number`.

Btw if it is only `tree_weight`, we may as well change the backend option `weight` to `tree_weight` for the `motions_default_sorting` setting, this would require the client settings to be changed, but at least it wouldn't require this extra if condition in the long term. What do you think @jsangmeister ?

jsangmeister: Ah, I see - please check if it works without the `id` and change it here and above if it does. I don't know about changing the default in the backend - I see the value, but at the same time `tree_weight` is rather client-specific and might confuse someone reading the models file, so I'm on the fence about it.

jsangmeister (approved):

emanuelschuetze (approved):

PR #2170, opened 2023-04-04
Fixed broken chip deselection in search selectors

### LINKED ISSUE #2161: file menu: remove group permission has a bug
**Description:** Remove a group permission right from a file has a weird bug. The x-button is not synched with the check boxes. Clicking the x-button don't remove the check box mark, but correctly removes the permission. If the user clicks now on the still marked check box, the permission will be granted again and the check box will be empty. 

Example - Step 1
Screenshot_20230331_125301

Step 2 - removed permission by clicking x
Screenshot_20230331_125343

Step 3 - unmark the check box -> permission granted again
Screenshot_20230331_125425

**Reproduction:**
1. open a meeting
2. open files menu and upload a file and grant Admin-group permission to see it
3. open files menu and click on the three-dot-menu behind the file. 
4. Remove the permission with the x-Button
5. remove the mark in the check box

**What should happen:**
X-button and check box marks should be synched. Both way of removing a permission should work correctly in synch not against each other.

luisa-beerboom:
Closes #2161

### DIFF
--- a/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.ts
+++ b/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.ts
@@ -251,9 +251,6 @@ export abstract class BaseSearchSelectorComponent extends BaseFormFieldControlCo
     public onChipRemove(itemId: Id): void {
         this.addOrRemoveId(itemId);
 
-        const value = this.selectableItems.find(item => item.id === itemId);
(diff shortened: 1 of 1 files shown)


bastianjoel:
This reverts the changes from  #2148

emanuelschuetze (approved):

emanuelschuetze:
@bastianjoel I cannot find any regressions - because you wrote from reverting #2148. Can you check and confirm that this PR works also for #2148?

bastianjoel:
The logic added by #2148 was removed because the change from this PR also fixes the previous issue.

PR #2172, opened 2023-04-04
fix poll tabledata read from method

### LINKED ISSUE #2171: Client gets unresponsive when accessing poll with a lot of votes
The client get unresponsive when accessing an election with published results and a chart.

bastianjoel:
resolves #2171


emanuelschuetze (approved):
not tested

PR #2174, opened 2023-04-04
fix single votes and entitled users tables

### LINKED ISSUE #2173: Single votes and entitled users tables break when scrolling
In polls when viewing the entitled users and scrolling in that list the list behaves weird.

bastianjoel:
resolves #2173


luisa-beerboom (approved):

luisa-beerboom:
Seems sensible but does this affect any other lists?

bastianjoel:
> Seems sensible but does this affect any other lists?

No

emanuelschuetze (changes requested):
I run into this typeerror if I open the entitled users table of an old motion poll:

```
TypeError: Cannot read properties of undefined (reading 'localeCompare')
    at entitled-users-table.component.ts:23:50
```
(same in current main)

bastianjoel:
I cannot replicate this. But as this is not related to the original issue I would suggest opening another issue for that.

emanuelschuetze (approved):

PR #2184, opened 2023-04-11
disable tinymce blockquoute

### LINKED ISSUE #2135: Motion text editor: justification of text format didn't work

bastianjoel:
resolves #2135

### DIFF
--- a/client/src/app/ui/modules/editor/components/editor/editor.component.ts
+++ b/client/src/app/ui/modules/editor/components/editor/editor.component.ts
@@ -1,6 +1,5 @@
 import { AfterViewInit, Component, forwardRef, Input } from '@angular/core';
 import { NG_VALUE_ACCESSOR, UntypedFormBuilder, UntypedFormControl } from '@angular/forms';
-import { marker as _ } from '@biesbjerg/ngx-translate-extract-marker';
 import { TranslateService } from '@ngx-translate/core';
 import { Deferred } from 'src/app/infrastructure/utils/promises';
 import { BaseFormControlComponent } from 'src/app/ui/base/base-form-control';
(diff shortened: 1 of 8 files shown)


client/src/app/ui/modules/editor/components/editor/editor.component.ts
```diff
@@ -36,6 +37,48 @@ export class EditorComponent extends BaseFormControlComponent<string> implements
...
+                    { title: _(`Pre`), format: `pre` }
+                ]
+            },
+            {
+                title: _(`Alignment`),
+                items: [
+                    { title: _(`Left`), format: `alignleft` },
+                    { title: _(`Center`), format: `aligncenter` },
+                    { title: _(`Right`), format: `alignright` },
+                    { title: _(`Justify`), format: `alignjustify` }
```
luisa-beerboom, line 78: Haven't looked too deeply into how this works with tinyMCE, but shouldn't the `alignjustify` item be removed if you want to remove the option?

luisa-beerboom (commented):
Aside from the more specific comment:
As far as I see, this implements blanket changes on the entire editor.
Wouldn't this cause the other places where we use the editor component to also not have blockquote?
Do we want to disable blockquote everywhere or only in places where it may interfere with pdf generation?

luisa-beerboom (approved):

rrenkert (approved):

PR #2187, opened 2023-04-11
Added functionality for clear motion block choice

### LINKED ISSUE #2153: motion block: blocks can't be removed from motions via multi selection
**Descritpion:** Removing motions blocks from motions via multi selection in the motion menu is not possible. The function and the clear button doesn't work. In return it is possible to add or change a motion block via multi selection.

Dialog field with button and selection that don't work.
Screenshot_20230330_161435

**Reproduction:**
1. open a meeting
2. open motions and create one motion
3. open motions > three-dot-menu > motion blocks: create two new blocks
4. head back to motion menu
5. open three-dot-menu > multiselect > choose the motion > click again on the three-dot-menu > set motion block > add the motion to a block
6. now  open three-dot-menu > set motion block > click on the button Clear motion block > nothing happens > choose the block of the motion > try again to clear it

**What should happen:**
Via multi selection it should be possible to remove a motion block.

luisa-beerboom:
Closes #2153

### DIFF
--- a/client/src/app/gateways/repositories/motions/motion-repository.service/motion-repository.service.ts
+++ b/client/src/app/gateways/repositories/motions/motion-repository.service/motion-repository.service.ts
@@ -342,8 +342,9 @@ export class MotionRepositoryService extends BaseAgendaItemAndListOfSpeakersCont
     }
 
     private getUpdatePayload(update: any, viewMotion: Motion & { workflow_id: Id }): any {
+        const ownMotion = this.getViewModel(viewMotion.id);
         const updatePayload = Object.keys(update).mapToObject(key => {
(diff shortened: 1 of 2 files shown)


bastianjoel (approved):

rrenkert (changes requested):
Maybe an edge case, but:
* Clear the motion block as described in the issue
* Keep the motion block dialog open
* Try to set a motion block
-> dialog closes without setting the block.

The data object of motion.update does not contain the block_id.

luisa-beerboom:
>     * Keep the motion block dialog open

???
If a Motion block is cleared through the dialog, the dialog will close?
What exactly are you doing @rrenkert ?

rrenkert (approved):

PR #2190, opened 2023-04-12
Trust media URLs in projector so they show up correctly

res2k:
Addresses #2183


jsangmeister:
Thank you for your input! Looks good on first glance.

bastianjoel (approved):
I think this might have been introduced by #2062 but I'm not entirely sure why this did not cause any problems back then. 
Anyway, looks good!

Issue #2106, opened 2023-03-18
ICC count-user-service


ostcar:
The `user-count` does not work on bigger meetings. I don't know the reason but fond a bug in this context.

When a client receives a [cound-user-request](https://github.com/OpenSlides/openslides-client/blob/cda27523d520f6add054090320f044c646820dc2/client/src/app/site/pages/meetings/pages/home/pages/meeting-info/services/count-users.service.ts#L18), he sends a response like 

```
{
    "name": "count-user-response",
    "message": {
        "token": "nochmal bitte2",
        "data": {
            "userId": 16438
        }
    },
    "channel_id": "BBGaSEjZ:16438:15970",
    "to_meeting": 966,
    "to_channels": [
        "BBGaSEjZ:16438:15934"
    ]
}
```

This tells the ICC-Service to sends this message to the channel `BBGaSEjZ:16438:15934` AND to all users in meeting `966`. But only the user that clicked on "count users" should receive this messages. So the attribute `to_meeting` has to be removed.

When I read the client code, it seems, that the [notify.service](https://github.com/OpenSlides/openslides-client/blob/main/client/src/app/gateways/notify.service.ts) currently does not support to skip the `to_meeting` attribute. Maybe the client `notify.service` has to be rewritten. The attributes `to_meeting`, `to_channels` and `to_users` are optional. If non are provided, the message is send to nobody. If more then one of the attributes are provided, the message is send to all of them.

luisa-beerboom:
The `to_meeting` in the request may be removed by PR #2201

PR #2201, opened 2023-04-14
Invite to call now only invites selected user

### LINKED ISSUE #2193: "Invite to conference room" invites all users
When a user from the speaker list is invited to the conference room via the three-point menu, all users receive the invitation. Only the one user who was invited should receive this invitation.

luisa-beerboom:
Closes #2193

It seems the client was sending the following sort of message to the ICC-notify route:
```
{
  "name":"invitationToCall",
  "message":{
     "inviter":"Administrator"
  },
  "channel_id":"8rLPRvPt:1:7",
  "to_meeting":2,
  "to_users":[6]
}
```

The effect that was seemingly desired with this kind of request, was to only send the message to the user when he is in the meeting.
But this form of request actually caused the message to be sent to all users in the meeting, as well as to the target user regardless of where he is. (This is the correct behaviour from the ICC services definition)

I've fixed this now, but for the future I think it would be a nice feature if the ICC/notify could also handle more exclusive requests, I think I'll write a proposal issue for this => OpenSlides/openslides-icc-service#100

### DIFF
--- a/client/src/app/gateways/notify.service.ts
+++ b/client/src/app/gateways/notify.service.ts
@@ -195,8 +195,7 @@ export class NotifyService extends BaseICCGatewayService<ChannelIdResponse | Not
         const notify: NotifyRequest<T> = {
             name: data.name,
             message: data.message,
-            channel_id: this.channelId,
-            to_meeting: this.activeMeetingIdService.meetingId!
+            channel_id: this.channelId
         };
         if (data.toAll === true) {
             notify.to_all = true;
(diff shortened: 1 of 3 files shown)


m-schieder (approved):

luisa-beerboom:
@emanuelschuetze When testing this please check if it also fixes  #2106

emanuelschuetze:
> @emanuelschuetze When testing this please check if it also fixes issue #2106

Tested: Does not work! (Super)Admin sends ICC request `count-user-request`, but normal users in this meeting get no ICC message. Counter shows 0 activ users.

emanuelschuetze (approved):
Jitsi invite action works as expected

emanuelschuetze:
@luisa-beerboom Please merge if you like - and fix icc issue in new PR

PR #2210, opened 2023-04-17
remove meeting user_ids

### LINKED ISSUE #2107: Do not use meeting/user_ids anymore
We have the definition, that a user is in a meeting, if he is in a group of a meeting: https://github.com/OpenSlides/OpenSlides/wiki/Users#user-association-to-meetings

Currently we have the field [meeting/user_ids](https://github.com/OpenSlides/openslides-backend/blob/5a4dd3448bbb7ec3fceb6fe25e70149fe9d8cd50/global/meta/models.yml#L1565-L1569). It contains all user ids, that at least in one group of the meeting.

This field has a problem. It is not of type `relation-list`. So it is not restricted. This means, if a requestuser can see only some users of the meeting, he still gets all user-ids of this meeting. You can test this, when you have a requestuser that has not the permission `user.can_see` and visit the meeting list or the commitee detail view.

The solution is easy. Do not use this field but calculate it. To calculate it, you have to fetch meeting/group_ids/user_ids:

```json
{
    "collection": "meeting",
    "ids": [
        1
    ],
    "fields": {
        "group_ids": {
            "type": "relation-list",
            "collection": "group",
            "fields": {
                "user_ids": null
            }
        }
    }
}
```

Then the user_ids of all fields have to be merged.

When you do it like this, the restricter removes the user_ids, that the request-user is not allowed to see.

After this is done in the client (and other services), the field can be removed in the backend.

bastianjoel:
resolves #2107 

Might need some testing especially if deleting and changing groups etc. still works. Currently only tested for participant list.

### DIFF
--- a/client/src/app/domain/models/meetings/meeting.ts
+++ b/client/src/app/domain/models/meetings/meeting.ts
@@ -243,7 +243,6 @@ export class Meeting extends BaseModel<Meeting> {
 
     // Other relations
     public present_user_ids!: Id[]; // (user/is_present_in_meeting_ids)[];
-    public user_ids!: Id[]; // Calculated: All ids all users assigned to groups.
     public reference_projector_id!: Id; // projector/used_as_reference_projector_meeting_id;
     public default_projector_$_ids!: string[]; // projector/used_as_default_$_in_meeting_id;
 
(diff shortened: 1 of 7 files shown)


luisa-beerboom (approved):

emanuelschuetze (changes requested):
If you remove meeting user_ids please remove also the particiapants number in meeting list view (if user is NOT allowed to see participants in meeting =! users.can_see).

For committee admins the field "Assigned accounts" in committee detail view can be removed - I think it was the reason for the meeting user_ids field. 

Important: the client performance for new calculation of participant numbers has to be checked for meeting list and committee detail with hight participant numbers!

PR #2215, opened 2023-04-20
Amendment list uses amendment workflow stati in filter

### LINKED ISSUE #2213: State filter in amendment list should use amendment default workflow
Use amendment default workflow (instead motion default workflow) for amendment list view.

luisa-beerboom:
Closes #2213


jsangmeister (approved):
not tested

PR #2224, opened 2023-04-21
Replaced special characters in motion diff

emanuelschuetze:
moved from OS3


bastianjoel:
```
    const textarea = document.createElement(`textarea`);
    textarea.innerHTML = html;
    html = textarea.value;
 ```
 Is supposed to replace special characters. If there are problems with the dash I would suggest just replacing the dash via regex.

emanuelschuetze:
I updated my PR and replace only the ndash character.

luisa-beerboom (approved):

Issue #1767, opened 2022-11-02
amendment bug with nested lists


MSoeb:
**Description:** If an amendmend has a nested list, all layers are crossed out red, if for example only the last layer is changed by the amendment.

Change in layer 4:
![Screenshot_20221102_124936](https://user-images.githubusercontent.com/105491030/199482624-ad6e3bf6-e062-4ea6-bc46-22696401892c.png)

Output: All crossed out. 
![Screenshot_20221102_124919](https://user-images.githubusercontent.com/105491030/199482604-2b7a0c38-1ec6-4711-90c2-e3dc5c978092.png)

**Reproduction:**
1. open a meeting
2. open motions
3. create a motion by clicking the plus-icon. add a 4 layer listed nest to the motion text.
4. create an amendment and change something in layer 4.
-> all layers are crossed out.

**Expected behavior:** Only the changed parts of a motion should be affected by the dif-view.

emanuelschuetze:
Another simple example to reproduce:

* Create new motion with nested list, e.g.
```
<ul>
<li>1</li>
<li>2
<ul>
<li>a</li>
<li>b</li>
<li>c</li>
</ul>
</li>
<li>3</li>
</ul>
```
* create a (paragraph-based) amendment > select 2nd paragraph (with nested list) > click NEXT
* add new sub-item to list (e.g. entry "d") > click CREATE
=> amendment shows full diff (= all items are red/green). Expected: only new line should be green

CatoTH:
@emanuelschuetze https://github.com/OpenSlides/openslides-client/pull/2226 should hopefully solve this.

Issue #2129, opened 2023-03-25
Bulleted lists in amendments destroy the amendment presentation


alveolus:
`…/…/motions/…`

### Bug description
If an additional paragraph containing an unsorted bullet list is added at the end of a motion via an amendment, the changes are not displayed correctly.

![Bildschirm­foto 2023-03-25 um 09 31 36](https://user-images.githubusercontent.com/23032204/227706619-25e08721-b8a9-4ba0-8513-555da1438f9b.png)

### Expected behavior
Additions should be inserted and displayed correctly, even if they contain enumerations. Like this, but with bullet points:

![Bildschirm­foto 2023-03-25 um 09 32 39](https://user-images.githubusercontent.com/23032204/227706616-0ea947ce-29d4-400d-a8c7-359916f51fe2.png)

### Additional context
OpenSlides: → 4.0.1-main-20230309-2f5789e

CatoTH:
Hi,

sorry for the delay. I created a Pull Request for this issue here: https://github.com/OpenSlides/openslides-client/pull/2226

PR #2226, opened 2023-04-23
Bugfix: appending lists in amendments produced broken diff (#2129)

CatoTH:
This solves two issues:
- https://github.com/OpenSlides/openslides-client/issues/2129
- https://github.com/OpenSlides/openslides-client/issues/1767

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-diff.service/motion-diff.service.spec.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-diff.service/motion-diff.service.spec.ts
@@ -865,7 +865,7 @@ describe(`MotionDiffService`, () => {
             expect(diff).toBe(expected);
         }));
 
-        it(`handles insterted paragraphs infront of list`, inject([MotionDiffService], (service: MotionDiffService) => {
+        it(`handles inserted paragraphs in front of list`, inject([MotionDiffService], (service: MotionDiffService) => {
             // Hint: line number should be moved into first element
             const before = `<ul><li><span class="os-line-number line-number-1" data-line-number="1" contenteditable="false">&nbsp;</span>Lorem ipsum</li></ul>`,
                 after = `<p>Add before UL</p><ul><li><span class="os-line-number line-number-1" data-line-number="1" contenteditable="false">&nbsp;</span>Lorem ipsum</li></ul>`,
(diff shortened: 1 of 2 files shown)


emanuelschuetze:
Does this PR also fixes #1767?

CatoTH:
@emanuelschuetze Now it does ;-)

luisa-beerboom (approved):

emanuelschuetze (changes requested):
#1767 is not yet fixed.

This example in amendment detail view looks good:
amendment-detail

But the created PDF shows wrong line ("3"):
amendment-pdf


And this example from initial issue message does not work:
amendment-issue1767


The other issue #2129 is fixed and looks good!

CatoTH:
@emanuelschuetze I fixed the second issue with the deeply nested lists (where a text replacement is done).

As for the PDF, this does not look like an issue withe the diff actually. The HTML generated and passed to `this.htmlToPdfService.convertHtml` in https://github.com/OpenSlides/openslides-client/blob/main/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts#L636 is as follows:

```
<h3>Zeile 5</h3><div class="paragraphcontext">
<ul class="os-split-after">
<li class="os-split-after"><span class="os-line-number line-number-2" data-line-number="2" contenteditable="false">&nbsp;</span>2 
<ul class="os-split-after">
<li><span class="os-line-number line-number-3" data-line-number="3" contenteditable="false">&nbsp;</span>a</li><li><span class="os-line-number line-number-4" data-line-number="4" contenteditable="false">&nbsp;</span>b</li></ul></li></ul></div>

<ul class="os-split-before"><li class="os-split-before">
<ul class="os-split-before">
<li><span class="os-line-number line-number-5" data-line-number="5" contenteditable="false">&nbsp;</span>c</li>
<li class="insert">d</li>
</ul></li></ul>
<div class="paragraphcontext"></div>
```

That looks correct to me - the  `insert` only is at `d` (4 in your screenshot). Maybe the PDF renderer can't handle that?

emanuelschuetze:
Rebased on top of current upstream/main branch.

Retested. Found 3 broken diffs in amendments, see screenshots:
amendment-01
amendment-02a
amendment-02b

CatoTH:
@emanuelschuetze So I added one more fix that addresses the second issue (where the "Ebene 3" is replaced by "Ebene 3b").
When changes are made at several lines at once, I have a feeling the effort in trying to fix the issues will be completely out of proportion to how often this will actually happen and the nuisance that the current situation is. There is just too much going on in the Diff (which not only needs to work around the HTML formattings, but also the line numbers)

emanuelschuetze (approved):

emanuelschuetze:
Rebased

PR #2237, opened 2023-04-25
set autopilot min height

### LINKED ISSUE #2024: Flickering Scrollbar during Split-Screen mode
**Description:** When OpenSlides is used on one screen in two browser windows and one window has the Autopilot open, flickering occurs on the scrollbar when content is projected. 

The red square area flickers:
Screenshot_20230214_112312_bearbeitet

**Reproduction:** 
Requirements: 
- A meeting with different motions and/or topics is needed to switch between projections. 
- two monitors 

1. open a meeting on monitor 1.
2. open the autopilot in a new tab, separate it from the browser and place it on the left side of monitor 2.
3. head back to monitor 1. open in the meeting the agenda or motions (depends on the content you have for projection.) in a new tab, separate it from the browser and place it on the right side of monitor 2
4. now go to right browser window on monitor 2 and project content via cam-icon until the scrollbar on the browser window with the Autopilot starts to flicker. 

**What should happen:**
By switching the projections, the scrollbar should not start to flicker if OpenSlides is used in Split-Screen mode.

### LINKED ISSUE #2259: It should not be possible to scroll over (head-)bars
Currently a bunch of views have problems where either the global headbar, the view headbar or the filter bar can be scrolled out of the view.

All of them should always remain fixed to their position.

Even if there is some sort of voting banner.

One possible approach to this would be to move all three possible bars into the same component (something like `HeadbarCollectionComponent`), control it via a global service (i.e. what kind of headbar should be shown, whether there should be a filter bar, what the filter bar should contain, etc) and place the new component where only the global headbar is currently placed.

bastianjoel:
This is to prevent the scroll bar from flickering by just always displaying it. Also a rework of the main container will be implemented.

resolves #2024 
resolves #2259

### DIFF
--- a/client/src/app/site/modules/global-headbar/global-headbar.service.ts
+++ b/client/src/app/site/modules/global-headbar/global-headbar.service.ts
@@ -1,0 +1,15 @@
+import { TemplatePortal } from '@angular/cdk/portal';
+import { Injectable } from '@angular/core';
+
+@Injectable({ providedIn: `root` })
+export class GlobalHeadbarService {
+    private _headbar: TemplatePortal = null;
+
+    public get headbar() {
+        return this._headbar;
+    }
+
+    public set headbar(portal: TemplatePortal) {
+        this._headbar = portal;
+    }
+}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Start dev setup
         run: make run-dev ARGS="-d"
       - name: Wait for dev setup
-        uses: iFaxity/wait-on-action@v1
+        uses: iFaxity/wait-on-action@v1.1.0
         with:
           resource: https://localhost:8000
           timeout: 300000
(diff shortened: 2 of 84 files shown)


jsangmeister:
Can you wrap this in media queries so it only triggers on the specific pixels where it is required to? Always displaying the scroll bar could also be archieved with `overflow-y: scroll`, right?

bastianjoel:
This now also resolves ##2259 partially. Head bars will not scroll away anymore as they are added to the global headbar instead of the page content.

client/src/app/site/modules/global-headbar/global-headbar.service.ts
```diff
@@ -0,0 +1,15 @@
+import { TemplatePortal } from '@angular/cdk/portal';
+import { Injectable } from '@angular/core';
+
+@Injectable({ providedIn: `root` })
+export class GlobalHeadbarService {
+    private _headbar: TemplatePortal = null;
+
+    public get headbar() {
```
luisa-beerboom, line 8: Return type

luisa-beerboom (changes requested):
Just one styling remark

emanuelschuetze (approved):

luisa-beerboom (approved):
@bastianjoel please fix the styling at a later date

PR #2238, opened 2023-04-25
optimized calender view for phone

Elblinator:
Only changes in scss file: 

Task -> solution
move time subtitles above each group -> changed flex direction (column) , and align-items (center)
use the full page width for meeting boxes -> automatically done because of changed flex-direction
use smaller font sizes -> if phone-display use 0.5em for font-size


luisa-beerboom (approved):

emanuelschuetze (changes requested):
In dark mode the font color is not correct:
darkmode

PR #2239, opened 2023-04-25
fix high cpu load on motion create

bastianjoel:
resolves an issue where there is high cpu load when accessing motion create. 

This happened in meetings with a lot of agenda items because the item sort function would be called on every interaction.


bastianjoel:
Changing to draft as this should be done as described in #2241

bastianjoel:
There is already a type on that getter so no change needed.

luisa-beerboom (approved):

emanuelschuetze (approved):

PR #2250, opened 2023-05-02
fix spinner not disappearing when blocking media access

### LINKED ISSUE #2247: Endless spinners when entering the conference room
If you want to enter the conference room and the browser does not have access to the camera and/or micro, an endless spinner appears. This can only be removed by reloading.
 
Instead we want to get a message that no access to the micro or camera is possible. If we only get access to the micro, we should still join the conference room.

bastianjoel:
resolves #2247


emanuelschuetze (approved):
Tested. It was only a Firefox issue: Firefox shows now the media access error message as expected. Chromium works, too.

PR #2256, opened 2023-05-03
fix missing autoupdate for amendment change recommendations

### LINKED ISSUE #2225: Missing autoupdate request for change recommendation
* create motion (text, e.g. "A")
* create amendment (new text, e.g. "A1")
* create change recommendation in amendment (new text, e.g. "A1a")
* set amendment to 'accepted'
* set motion to 'accepted'
* go to motion -> text shows "A1a"
* now reload => text shows "A1" (change recomendation missing)
* open amendment and go back to motion -> text shows again "A1a"

Same problem with amendment list -> reload => no change recommendation shows in preview

bastianjoel:
resolves #2225

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/motions.subscription.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/motions.subscription.ts
@@ -174,7 +174,8 @@ export const getMotionDetailSubscriptionConfig: SubscriptionConfigGenerator = (.
             { idField: `lead_motion_id`, fieldset: [`text`] },
             {
                 idField: `amendment_ids`,
-                fieldset: [`text`, `modified_final_version`, { templateField: `amendment_paragraph_$` }]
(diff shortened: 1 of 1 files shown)


rrenkert (approved):

emanuelschuetze (changes requested):
Change recommendations are NOT loaded ...
* ... if you reload the amendment list view -> wrong diff without change recommendation
* ... if you export PDF in motion list view -> wrong diff without change recommendation (tested with/without settings option to show amendments in motion list)

bastianjoel:
The problem within the PDF export does not seem to be related to a missing subscription. Therefore I would suggest opening another issue for that.

emanuelschuetze:
> The problem within the PDF export does not seem to be related to a missing subscription. Therefore I would suggest opening another issue for that.

New issue created: https://github.com/OpenSlides/openslides-client/issues/2257

emanuelschuetze (approved):

PR #2269, opened 2023-05-09
amendment list pdf export with nested lists fix

### LINKED ISSUE #2265: amendment list pdf export with nested lists doesn't work
**Description:** The amendment list can't be exported with nested lists that include more than one layer. The pdf export includes only one layer. All other layers are ignored.

example: Amendment with nested list with more than one layer
Screenshot_20230509_070119

PDF export without all layers 
Screenshot_20230509_070237

**Reproduction:**
1. open a meeting
2. open _motions_
3. create a motion with a text line and a nested list
4. create an amendment for the nested list with more than one addtional layers
5. head back to _motions_ menu: open the three-dot-menu and choose _amendments_
6. in _amendments_: open three-dot-menu and choose _amendment list (PDF)_
7. -> expect nested list layer one alle layers will be ignored

**What should happen?**
All layers should be correctly exported in the PDF file.

bastianjoel:
resolves #2265

### DIFF
--- a/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
+++ b/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
@@ -309,8 +309,18 @@ export class HtmlToPdfService {
      * Can be overwritten by subclasses for more specific functionality.
      */
     protected createListParagraph(data: CreateSpecificParagraphPayload): any {
-        const children = this.parseChildren(data.element, data.styles);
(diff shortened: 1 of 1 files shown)


luisa-beerboom (approved):
For now this is fine, but please make a cleanup issue to remove the ts-ignore and refactor the html-to-pdf.service properly

rrenkert (approved):

emanuelschuetze (approved):

PR #2274, opened 2023-05-10
enable prettier for html files

### LINKED ISSUE #2078: apply prettier rules to html files
We currently do not have any code style checks for html files. As we already use prettier for all other files I suggest applying it also for html files. 
This isn't hard to do but we might want to do this in a moment when we do not have a lot of pull request open as this will result in merge conflicts. 

Currently 243 errors are found when enabling `prettier/recommended` as following: 
```
        {
            files: ['*.html'],
            excludedFiles: ['*inline-template-*.component.html'],
            extends: ['plugin:prettier/recommended'],
            rules: {
                'prettier/prettier': ["error", { parser: 'angular' }]
            }
        }
```

bastianjoel:
resolves #2078 

Seems like this does not have any major conflicts with feature/remove-template-fields so I would suggest merging this.


rrenkert (approved):

bastianjoel:
Conflicts with this PR should be resolved as following (assuming your are currently on the branch you want to update and the remote is in sync):

Merge until commit before this change
```
git merge 4d16f75cc8ca89f16337c1ef220cf23c3a9b3dc6
```


Merge this change, ignore conficlts
```
git merge --strategy-option ours 219a13dc8a71718840c48c5e661882464ed8b392
```

Cleanup, then commit changes
```
npm run cleanup
(git add, commit, ...)
```
Merge rest
```
git merge main
```

PR #2295, opened 2023-05-15
Improve email not sent error messages

### LINKED ISSUE #2261: Give correct error messages on wrong email config
**Describe the bug**
I have set up the mailserver according documentation. When I then try to send an email to an newly created user within openslided I get the message "No emails were send. The user has no email, so the invitation email could not be send." altrough the user definitely have a email.

**How to Reproduce**

1. Go to Accounts.
2. Add an Account with an email.
3. On the profile page of the account on the tree dots klick "send invitation email"

**Expected behavior**
The email should be sended. No error should be thrown.

**System information**

- OpenSlides-Version: 4.0.1-main-20230424-19b6cce
- Device: MacBook Pro 16" 2021.
- OS: macOS, current Version
- Browser: Safari & Firefox
- Browser-Version: current versions

**Additional context**
```
(Excerpt from docker-compose.yml)

   environment:
      << : *default-environment
      OPENSLIDES_BACKEND_COMPONENT: action
      EMAIL_HOST: "smtp-relay.sendinblue.com"
      EMAIL_PORT: "587"
      EMAIL_HOST_USER: "*************"
      EMAIL_HOST_PASSWORD: "***********"
      EMAIL_CONNECTION_SECURITY: "SSL/TLS"  # This can also be "NONE" or "STARTTLS".
      EMAIL_TIMEOUT: "5"  # Timeout in seconds for blocking operations like the connection attempt.
      EMAIL_ACCEPT_SELF_SIGNED_CERTIFICATE: "false"  # Or "true" of course.
```

**Screenshots**

Fabian_Müller_-_OpenSlides

bastianjoel:
resolves #2261 

requires https://github.com/OpenSlides/openslides-backend/pull/1757


luisa-beerboom (commented):
What about the case when the backend sends "has no valid email address"? I think any error that may potentially be caused by the user that sees the message should probably be specified...

jsangmeister:
How about we introduce error codes for this in the backend? This string matching does not seem like the best option.

emanuelschuetze (changes requested):
I get this TypeError if I try to send emails and change `EMAIL_CONNECTION_SECURITY` to `"SSL/TLS"` (with quotes) to reproduce an error:
```
 TypeError: Cannot read properties of null (reading 'message')
    at http.service.ts:87:35
```

luisa-beerboom (approved):

emanuelschuetze (approved):

PR #2305, opened 2023-05-17
Fixed wrong vote delegation updates

### LINKED ISSUE #2304: Autoupdate processing problem when a delegation is deleted. Reproduction
1. Set a delegation between two users A and B in the edit dialogue
2. Open detail view of one of the two users and delete the delegation there
3. Return to the list: One of them will still show the delegation, even though it is gone.
4. Reloading the page fixes it
- This happens as well when other users receive the update, this can cause the polls to show delegation voting options that shouldn't exist

luisa-beerboom:
Closes #2304

SUMMARY:
Template field update handling (either in autoupdate or in client) broken. Since this will not matter once the template field removal happens,  I added a temporary quick fix.

MORE VERBOSE:
It seems that the problem came from the fact that the autoupdate doesn't send autoupdates for the empty specified template field values. This meant that, while the main template field variable was updated to not include the meeting id (signalling that there is no data for this meeting), the old meeting variable (`vote_delegated_$2_to`) still held the old value.

I caught that by changing the functions that look up the data per template attribute, so that they check whether there even should be data for this meeting.
This is a bit inefficient, but I figured it would be quicker than re-writing the autoupdate, and it shouldn't really matter too much, given that we'll likely soon finish the template field removal.

### DIFF
--- a/client/src/app/domain/models/users/user.ts
+++ b/client/src/app/domain/models/users/user.ts
@@ -156,11 +156,16 @@ export class User extends BaseDecimalModel<User> {
     }
 
     public vote_delegated_to_id(meetingId: Id): Id {
-        return (this as any)[`vote_delegated_$${meetingId}_to_id`];
+        return this.vote_delegated_$_to_id?.includes(`${meetingId}`)
+            ? (this as any)[`vote_delegated_$${meetingId}_to_id`]
+            : undefined;
     }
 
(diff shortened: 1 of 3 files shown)


luisa-beerboom:
@rrenkert please organize a code review, as I have no idea who will even be there in the next few days

m-schieder (approved):

emanuelschuetze (approved):

PR #2317, opened 2023-05-23
Made autopilot title link point to new tab with mediafiles

### LINKED ISSUE #2260: Projected image has wrong link in autopilot
**Description:** If an image is projected in the Autopilot, the title links to the dashboard and not to the image. 

For example: The title link is bild_500x200.jpg
Screenshot_20230508_102853

**Reproduction:**
1. open a meeting
2. open **files** > upload an image
3. click on the three-dot-menu behind the image and choose _Project_
4. go to the Autopilot
5. Click on the title -> Dashboard opens

**What should happen?**
I'm a bit ambivalent, what solution would be the best. 
- First option: Open the image in the current tab. It would have the same behavior like opening a motion for example. 
- Second option: Open the image in another tab. It would have the same behavior like clicking on the projector. 

Both solutions are fine. But to be more consistence to the title link behavior in general, I would prefer the first option: The image should open in the same tab.

luisa-beerboom:
Closes #2260

### DIFF
--- a/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.html
+++ b/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.html
@@ -6,7 +6,7 @@
 <div class="content-container">
     <!-- Title Card -->
     <mat-card class="os-card" *ngIf="!!title">
-        <a [routerLink]="viewModelUrl || null" [state]="{ back: 'true' }">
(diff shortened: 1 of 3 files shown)


bastianjoel:
I made a change to this so the mediafile will not be added to the local model store when projected on the reference projector. 
This way it is not necessary anymore to request `owner_id`.

bastianjoel (approved):

emanuelschuetze (approved):

PR #2319, opened 2023-05-23
Rewrote star icon setting title

### LINKED ISSUE #2283: description for activating the Star-Icon needs improvement
**Description:** The description of the function to active the star-icon in the list of speakers needs an improvement. In the current state it doesn't reflect that this function will activate the new option for speakers. It suggests that another account, like admin, could mark speakers with the icon. But that's wrong. The admin could do it even without this setting. In fact, the speakers could mark their speechs with it.

Screenshot_20230511_145750

**Where to find:**
1. open a meeting
2. open seetings > list of speakers > Enable star icon to mark speaker (e.g. for contribution)

**What should happen:**
The german description and all others should be improved.

Old: Stern-Icon zum Markieren von Redner*innen aktivieren (z.B. zur Einbringung eines Antrags)
New: Stern-Icon Markierung für alle Redner*innen aktivieren (z.B. zur Erkennung der Einbringung eines Antrags)

luisa-beerboom:
Closes #2283

### DIFF
--- a/client/src/app/site/pages/meetings/services/meeting-settings-definition.service/meeting-settings-definitions.ts
+++ b/client/src/app/site/pages/meetings/services/meeting-settings-definition.service/meeting-settings-definitions.ts
@@ -323,7 +323,7 @@ export const meetingSettings: SettingsGroup[] = fillInSettingsDefaults([
                     },
                     {
                         key: `list_of_speakers_can_set_contribution_self`,
-                        label: _(`Enable star icon to mark speaker (e.g. for contribution)`),
+                        label: _(`Enable star icon usage for speakers`),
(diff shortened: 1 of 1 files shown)


jsangmeister (changes requested):
My suggestion in https://github.com/OpenSlides/openslides-client/issues/2283#issuecomment-1558903928 was "by speakers", since " for speakers has the same problem as the current formulation - it sounds like the setting controls whether speaker can be marked with the icon at all. "By speakers" clarifies that the speakers are the active part and can afterwards set the star icons themselves.

jsangmeister (approved):

emanuelschuetze (approved):

PR #2320, opened 2023-05-23
fix overflow problems

### LINKED ISSUE #2316: Workflow and Group restriction menu: UI Bug - horizontal Navigation bar is gone
**Description:** The workflow detail view and thee group restrictions menu have a serious UI bug. The horizontal scrollbar isn't available anymore. The content is cut off at the right side and can't be used. 
 
**Example:**
Screenshot_20230523_154653

**Reproduction:**
1. open a meeting
2. open motions
3. open the 3-dot-menu, choose workflow, choose complexe workflow
4. minimize the browser window to trigger the problem (if it is not visible)

**What should happen:**
A horizontal scrollbar should appear to see all status and the configuration/save menu.

bastianjoel:
resolves #2316


rrenkert (approved):
The flickering scrollbar is back again!

PR #2324, opened 2023-05-24
Search selectors: Replaced virtual scrolling with normal scrolling

luisa-beerboom:
Closes #2310 

Please test if this causes any sort of performance issues.

They may be especially possible when you have
- A lot of options in the search selector
- Multiple search selectors at once
- Search selectors in a long list view

Search selectors are include in the following places:
- [ ] List of speakers
- [ ] Agenda content object form (i.e. where visibility and parents can be set) and topic detail edit view (same as agenda)
- [ ] Attachment control (i.e. selector for mediafile attachments, f.E. for motions)
- [ ] Poll create/edit form (for groups)
- [ ] Assignment detail (Candidate and tag selection)
- [ ] Chat group dialogue (For setting read- and write-groups)
- [ ] Comment section list (For setting read- and write-groups)
- [ ] History list (Model selection)
- [ ] Access group settings for mediafiles (both upload and edit)
- [ ] Meeting settings: group-selection
- [ ] Multiple in motion detail view (submitters, supporters, category, workflow, extension fields)
- [ ] Multiple in participant detail view, participant create wizard and participant "quick-edit dialogue" (groups and delegations) (please pay special attention to the quick edit)
- [ ] Add to meetings wizard
- [ ] Account detail edit (Management level)
- [ ] Committee detail edit (tags, managers forwardings)
- [ ] Meeting create/edit (duplicate from, participants, admins, tags)
- [ ] File list move dialogue (target directory)
- [ ] Mediafile upload (target directory)

If this doesn't work performance-wise, our only option may be #2311

PR #2342, opened 2023-05-30
improve search selector

### LINKED ISSUE #2309: Improvements to Search-Value Selectors
**Describe the bug**
When adding people to the list of speakers, the selection dropdown remembers the last entered search query. When reopening the selection dropdown, newly entered text is appended to the last selection, so when quickly adding people to the list of speakers, one always has to delete the last query from the selection before finding other people, leading to a disruption of flow.

**How to Reproduce**
Enter the beginning of a name you want to add
image
Add the user
image
Reopen the dropdown
image
Start typing and don't find the relevant users
image

**Expected behavior**
When opening the selection dropdown, one can always start typing a new query without removing the old query first.

**System information**

- OpenSlides-Version: 4.0

bastianjoel:
resolves #2309

### DIFF
--- a/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.html
+++ b/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.html
@@ -23,7 +23,7 @@
             class="background-card"
             [formControl]="searchValueForm"
             placeholder="{{ 'Search' | translate }}"
-            (keydown)="$event.stopPropagation()"
+            (keydown)="onSearchKeydown($event)"
         />
     </div>
     <ng-container *ngIf="multiple && showChips">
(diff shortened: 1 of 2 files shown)


Elblinator (changes requested):
If a person was selected in the latest search the new query starts empty. 
If a person wasn't selected, but the focus went to another element (e.g. one of the speakers) then the query is not preemptied and still has to be deleted.

If the up/down-arrows are used to select a paticipant, the scrollbar does not follow the item in focus but stays at the place it previously was.

If a non-existent name (e.g. 'aaaaaa') is typed into the search query and deleted again, then the up/down arrows are not working anymore. They only start working again after exiting and reentering the search bar.
Sometimes the up/down arrows simply do not work (potentionally if special characters are used, or/and if only one participant is left, possibly more cases).

If we scroll though the names AND use the up/down keys: Then the placement of the focused item sometimes stays with the up/down keys (absolutely independent from the cursor) but sometimes the placement of the cursor marks an item as the focused item.

Double Tab/Shift Tab is equivalent to enter, aka. activates the item in focus

bastianjoel:
The input now gets cleared when closing the selection field. 

In terms of the selection issues I'm not entirely sure if we can fix this easily as this seems to be implemented within the material components. For the scrolling issue we also might want to wait for the outcome of #2324.

Elblinator (approved):

PR #2344, opened 2023-05-30
do not search for id fields in list search service

### LINKED ISSUE #2343: Do not search for ids in motion list view
The motion list view seems to search for ids currently. This is wrong as they are not displayed anywhere.

bastianjoel:
resolves #2343 

This now disables searching for ids in all list views.


luisa-beerboom (changes requested):
Shouldn't you at least put it back in for orga-level models that are navigated to via id? (meeting, committee, account)?

emanuelschuetze:
> @emanuelschuetze Should the search field be able to search for every visible item in a motion? -> if we search for '17', should the results include motions with a tag '17', with the state '17', etc.

This is not required because tags and states are selectable in dropdown filter menu. This PR should search by strings/numbers which are not in filter list.

emanuelschuetze (approved):
works good

luisa-beerboom (approved):

PR #2355, opened 2023-05-31
Rearrange tooltip positions so that they dont overlap

### LINKED ISSUE #2061: Rearrange impractical tooltips in lists
In multiple places within the client (primarily the list-views) there are problems when multiple icons are used in the same meta column.

The tooltip direction for the `IconContainerComponent` is currently hard-coded to always be at the bottom. When hovering over an icon, one will often be unable to easily switch to the tooltip of the item below by moving the cursor onto it.

This is because the hit box for the tooltip will completely cover that of the icon below. One has to purposefully move the cursor out of the general area to make the tooltip disappear and then carefully move it to the target item while making sure not to accidentally hover over the element above. This is obviously impractical.

Solve this problem by:
1. Creating an option `iconTooltipDirection` for the `IconContainerComponent` that can be used to change the direction of the tooltip.
2. Looking through the list-views and re-arranging the icon tooltips everywhere, where they get in each others way.
3. You may also consider if it may be wise to use a different default direction for the tooltips in general.

luisa-beerboom:
Closes #2061

### DIFF
--- a/client/src/app/site/modules/global-headbar/components/account-button/account-button.component.html
+++ b/client/src/app/site/modules/global-headbar/components/account-button/account-button.component.html
@@ -26,6 +26,7 @@
                         matTooltip="{{
                             ((isDarkModeActiveObservable | async) ? 'Light mode' : 'Dark mode') | translate
                         }}"
+                        matTooltipPosition="left"
                     >
                         <mat-icon>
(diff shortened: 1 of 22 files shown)


emanuelschuetze (approved):
Looks good for me.

PR #2358, opened 2023-06-01
Fix filter buttons

### LINKED ISSUE #2357: Can't remove current filters via buttons in filter-bar sometimes
Specifically:
Clicking on the `X` icon in the button doesn't do anything.
Clicking anywhere else on the button works as expected.

luisa-beerboom:
Closes #2357


Elblinator (changes requested):
After selecting filters the first click still doesn't deactivate the button. 
It doesn't matter whether the x or anywhere else on the button is clicked.

bastianjoel (commented):
This does not seem right and I cannot replicate the issue on current main.

Elblinator (approved):

Elblinator (changes requested):
X in participants is still not working immediately

PR #2373, opened 2023-06-05
fix cors issue in video player reachable check

bastianjoel:
We are currently checking if a stream is available before displaying it via a rxjs ajax method. This method is adding a `x-requested-with` header which makes a cors preflight request necessary. This requires additional configuration of the webserver the stream is hosted an. 
I don't think this is wanted here.


jsangmeister:
Please stick to https://github.com/OpenSlides/OpenSlides/wiki/Development-organization#pull-request-titles-and-commit-messages

PR #2375, opened 2023-06-06
Make list of speakers rearrangeable on mobile

### LINKED ISSUE #2306: List of speakers cannot be rearranged on mobile
**Describe the bug**
When resizing the window the sliders to rearrange the speakers disappear

Wide View:
image

Narrow View:
image

**How to Reproduce**
1. Add more than one person to the list of speakers
2. Resize the window so that the sidebar disappear

**System information**

- OpenSlides-Version: 4.0

bastianjoel:
proposal to resolve #2306 

This PR enables the rearrangement functionality for mobile devices. To deal with accidental rearrangements a delay was added. A user has to hold the rearrangement icon for a certain time before they can move it. This way when scrolling over the page the user will not accidentally move an entry. 

The Angular Material documentation mentions the used functionality to deal with these scenarios so I think this is the way to go here. https://material.angular.io/cdk/drag-drop/overview#delayed-dragging
Although we might need to adjust the delay time.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-content/list-of-speakers-content.component.html
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-content/list-of-speakers-content.component.html
@@ -110,7 +110,8 @@
             [live]="true"
             [input]="waitingSpeakers"
             [count]="canSee"
-            [enable]="canManage && (isSortMode || !isMobile)"
+            [enable]="canManage"
+            [dragDelay]="isMobile ? 500 : 0"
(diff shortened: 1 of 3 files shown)


Elblinator (changes requested):
There is no visual feedback that the user can move the item after they held the item long enough.

An easy solution would be to slightly change the colour of the movable tile

bastianjoel:
I will close this one for now as a visual feedback does not seem to be possible with angular materials API and we therefore might need a different solution.

Eich-hoernchen:
Tested it, the feedback looks good. 

I found another bug in the mobile view. On the small devices the More Vert icon falls of the tile.

image

Elblinator (approved):

PR #2437, opened 2023-06-20
Fix search service dark mode

bastianjoel:
The background of list elements did not change in dark mode and were not readable.


m-schieder (approved):
Tested

luisa-beerboom (changes requested):
I would still prefer to have this in the `global-components-style.scss`, so that it is in the same place as other styles that change depending on light/dark mode and so that it may be easily re-used for lists that require similar styling

bastianjoel:
Change request will be implemented in #2616

PR #2449, opened 2023-06-23
Implement point of order customization

### LINKED ISSUE #2447: Implement point of order customization
- It should be possible to set custom point of order texts in the LoS settings and assign them ranks (smaller number => bigger rank).
- There also should be a setting that allows one to switch on rankings
- If there are custom point of order texts:
   - If there is only one custom text, rename the button to the text and use the text as the point-of-order name. 
   - If there are multiple custom texts open a selection field that lets the user select one of the texts from the list

Backend issue OpenSlides/openslides-backend#1774

luisa-beerboom:
Closes #2447 

For now this only creates the _possibility_ for rankings in the meeting settings.
Everything that should truly happen in this feature still needs to be implemented.
The allocation lists template and validators also need to be edited to accommodate number inputs (with minimum validators, etc)


client/src/app/domain/models/point-of-order-category/point-of-order-category.ts
```diff
@@ -0,0 +1,29 @@
+import { Id } from '../../definitions/key-types';
+import { HasMeetingId } from '../../interfaces';
+import { BaseModel } from '../base/base-model';
+
+/**
+ * Representation of user group.
+ * @ignore
```
bastianjoel, line 7: Not sure if this should be here.

client/src/app/site/pages/meetings/pages/agenda/agenda.subscription.ts
```diff
@@ -118,6 +119,7 @@ export const listOfSpeakersSpeakerCountSubscription = {
     follow: [
         {
             idField: `speaker_ids`,
+            follow: [`point_of_order_category_id`],
```
bastianjoel, line 122: You should explicitly request the full fieldset.

client/src/app/site/pages/meetings/services/meeting-settings-definition.service/meeting-settings-definitions.ts
```diff
@@ -94,7 +112,30 @@ function fillInSettingsDefaults(settingsGroups: SettingsGroup[]): SettingsGroup[
...
+                        (Array.isArray(setting.key) ? setting.key : [setting.key]).map(key => ({
+                            key: key as string,
+                            transformFn: setting.transformFn,
+                            reverseTransformFn: setting.reverseTransformFn
+                        }))
+                    )
+            )
+            .reduce<{ [key: string]: SettingsItemTransformFunctions[] }>((dictionary, current) => {
+                const { key, ...rest } = current;
+                if (Object.keys(dictionary).includes(key)) {
```
bastianjoel, line 131:
```suggestion
                if (dictionary[key] !== undefined) {
```

Or maybe even
```suggestion
                if (Array.isArray(dictionary[key])) {
```

client/src/app/site/pages/meetings/services/meeting-settings-definition.service/meeting-settings-definitions.ts
```diff
@@ -380,6 +421,46 @@ export const meetingSettings: SettingsGroup[] = fillInSettingsDefaults([
...
+                                        return false;
+                                    }
+                                    const former = original.find(value => value.id === val.id);
+                                    return former.text !== val.text || former.rank !== val.rank;
+                                });
+                            const toCreate: { text: string; rank: number }[] = value
+                                .filter(val => !val.id)
+                                .map(val => ({ text: val.entry, rank: +val.allocation }));
+                            return { toDelete, toUpdate, toCreate };
+                        }
```
bastianjoel, line 461: I am not sure if this should be part of the field definition. Maybe introduce a new field type instead of the transform function. 
If that is too much work or would result in a even more hacky solution I would be fine with the current solution too.

bastianjoel (changes requested):

bastianjoel (approved):

Elblinator (approved):

emanuelschuetze (approved):

PR #2472, opened 2023-06-28
Make sort lists update only when relevant data is changed

### LINKED ISSUE #2458: Fix Agenda sort collapse all/expand all
while the agenda is being sorted it can happen, that the collapsed items expand all at once

**How to recreate:**

1. navigate to agenda > sort agenda
2. collapse all
3. let an autoupdate happen (e.g. change your own presence)
4. everything expands again

**wanted behaviour:**
the collapsed items should not expand because of an autoupdate, they should only expand when they are expanded again

luisa-beerboom:
Closes #2458 

How it should work now upon update:
1. If items were deleted they should be removed from the list, immediate sub-items should gain expansion and visibility level of the removed parent, _all_ sub-items should be moved left
2. If items are added they should be appended to the list
3. Other changes like name-changes will not affect the list in any way. (i.e. the sort view will also not update names while opened)

Notes for testing: Please especially check if
- Actual changes with the agenda item (for example change of topic title) still trigger an update
- Other sort lists are negatively/positively affected (for example the motion call list), as this performs changes in the base class

### DIFF
--- a/client/src/app/infrastructure/definitions/tree.ts
+++ b/client/src/app/infrastructure/definitions/tree.ts
@@ -16,6 +16,7 @@ export interface TreeNodeWithoutItem extends TreeIdNode {
 
 /**
  * A representation of nodes with the item atached.
+ * The id of the item should be the id of the node.
  */
 export interface OSTreeNode<T> extends TreeNodeWithoutItem {
     item: T;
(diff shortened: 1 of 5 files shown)


bastianjoel (approved):

emanuelschuetze (changes requested):
Open agenda sort view > Collapse all items > now trigger autoupdate in an other tab:
* set done flag for agenda item => sort view expand!
* add new topic item => sort view expand!
* switch from public to internal => sort view expand!
* add/edit some agend details (e.g. duration, comment) => sort view expand!

Similar problem in motion call list: collabse all motions > change metadata (e.g. state) in an other tab => call list view expand!

emanuelschuetze (approved):

PR #2477, opened 2023-06-29
In organization settings edit only send values that have been changed

### LINKED ISSUE #2476: Organization settings sends values for fields that haven't been changed
This can cause problems if a specific field is not being processed correctly, as an unacceptable value for it will be sent to the backend, even if it has not been changed, potentially causing all attempts to change other fields to fail as well.

While fixing this the bug, where the content of the attribute mapping field, if left blank, is not replaced with null, should also be fixed. This replacement should, of course, harmonize with the change detection.

luisa-beerboom:
Also made sure that the SSO config is replaced with null if the new value is an empty value

Closes #2476

### DIFF
--- a/client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
+++ b/client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
@@ -129,7 +129,7 @@ export class OrganizationSettingsComponent extends BaseComponent {
 
     public onSubmit(): void {
         const payload: any = this.orgaSettingsForm!.value;
-        payload.saml_attr_mapping = !!payload.saml_attribute_mapping
+        payload.saml_attr_mapping = !!payload.saml_attr_mapping
             ? JSON.stringify(JSON.parse(payload.saml_attr_mapping as string))
             : null;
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
code-review

rrenkert (changes requested):
After clearing the saml_attr_mapping reverting chenges does not work anymore:

ERROR TypeError: Cannot read properties of undefined (reading 'split')
    at OrganizationSettingsComponent.updateForm (organization-setting…component.ts:126:43)
    at OrganizationSettingsComponent.revertChanges (organization-setting…component.ts:107:18)

Same error appears on save, but does not prevent saving.

luisa-beerboom:
> After clearing the saml_attr_mapping reverting changes does not work anymore:
> 
> ERROR TypeError: Cannot read properties of undefined (reading 'split') at OrganizationSettingsComponent.updateForm (organization-setting…component.ts:126:43) at OrganizationSettingsComponent.revertChanges (organization-setting…component.ts:107:18)
> 
> Same error appears on save, but does not prevent saving.

What do you mean by "clearing the saml_attr_mapping"?

rrenkert:
> What do you mean by "clearing the saml_attr_mapping"?
I removed the content and saved the settings.

rrenkert (approved):

PR #2479, opened 2023-06-29
fix display for new side-nav-width

### LINKED ISSUE #2415: Fix display with new side-nav width
The side-navaigation now collapses if the display is 960px or less pixel wide.
At some point in time it collapsed at 959px. There are some leftover- artifacts which need adjusting.

Example of the first behaviour:
Screen if 961px (or more) wide: 
961px
Screen if 960px wide:
960px
Screen if 959px (or less) wide:
959px

The following views display this behaviour:
- inside a meeting: settings
- Dashboard
- Motions in blockview (NOT motion blocks)

Example of the second behaviour:
Screen if 960px wide:
960_NM
Screen if 959px wide:
959_NM

The following view displays this behaviour:
- motions 

Problem:
both behaviours are not adjusted to the now 960px tablet view.

Wanted Behaviour:
For the first:
the screen at 960px should look and behave like the screen at 959px.

For the second:
the distance between the object and the border should not make such a jump. The changed width of the object should not be  visible. 

To recreate these behaviours go to the view in question and shuffle between a display 961px, 960px and 959px wide.

Elblinator:
closes #2415

### DIFF
--- a/client/src/app/ui/modules/sidenav/components/sidenav/sidenav.component.ts
+++ b/client/src/app/ui/modules/sidenav/components/sidenav/sidenav.component.ts
@@ -30,7 +30,7 @@ export class SidenavComponent {
     }
 
     public get isOpen(): boolean {
-        return this.isMobile ? this.sideNav?.opened : this.desktopOpen;
+        return this.isMobile ? this.sideNav?.opened ?? false : this.desktopOpen;
     }
 
     public constructor(private vp: ViewPortService) {}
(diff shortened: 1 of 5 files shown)


client/src/app/ui/modules/sidenav/components/sidenav/sidenav.component.ts
```diff
@@ -30,7 +30,7 @@ export class SidenavComponent {
     }
 
     public get isOpen(): boolean {
-        return this.isMobile ? this.sideNav?.opened : this.desktopOpen;
+        return this.isMobile ? (this.sideNav ? this.sideNav.opened : false) : this.desktopOpen;
```
luisa-beerboom, line 33:
```suggestion
        return this.isMobile ? this.sideNav?.opened ?? false  : this.desktopOpen;
```

luisa-beerboom (approved):

rrenkert (approved):

PR #2480, opened 2023-06-29
Fix reappearing autoupdate error message

### LINKED ISSUE #2441: fix reappearing error-message
After an error message "Error talking to autoupdate service Try reconnect" was triggered once it reappears while staying in the dashboard and everytime the dashboard is entered again from a meeting.

Current behaviour:

1. navigate into any meeting -> agenda -> (any) topic
4. navigate directly to the dashboard 
5. the error message occurs.
6. the error message will reapprear while you are staying in dashboard, design or settings. The error will reappear IF no other page than these three was opened. It doesn't matter if you tried to resolve the error by pressing "try reconnect". The error message vanishes for every other page.
7. After activating this message once it will reappear everytime the dashboard is navigated to from any page in any meeting

Wanted behaviour: 
bestcase: the error should not be triggered at all 
worstcase: the "Try reconnect" should solve the error message for good and prevent it from reappearing

luisa-beerboom:
Closes #2441

### DIFF
--- a/client/src/app/site/services/model-request.service.ts
+++ b/client/src/app/site/services/model-request.service.ts
@@ -88,6 +88,15 @@ export class ModelRequestService {
         hideWhen,
         isDelayed = true
     }: SubscribeToConfig): Promise<void> {
+        const ids = modelRequest.ids;
+        const invalidIds = ids.filter(id => !this.isValidId(id));
+        if (invalidIds.length === ids.length) {
+            console.warn(`${subscriptionName}: No valid ids in configuration: Skipped subscription.`);
+            return;
+        } else if (invalidIds.length) {
+            console.warn(`${subscriptionName}: Skipped invalid ids in configuration.`);
+            modelRequest.ids = ids.filter(id => this.isValidId(id));
+        }
         const fn = async () => {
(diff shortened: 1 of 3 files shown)


client/src/app/site/services/model-request.service.ts
```diff
@@ -155,11 +144,35 @@ export class ModelRequestService {
...
         this._observableSubscriptionMap[requestFamilyName] = observable.subscribe(isTrue => {
             if (isTrue) {
                 this.closeSubscription(requestFamilyName);
             }
         });
     }
+
+    private isValidId(id: number): boolean {
+        return !Number.isNaN(id) || id > 0;
+    }
```
bastianjoel, line 177: I think it would be better to place this in `client/src/app/domain/definitions/key-types.ts` or `utils`.

Also:
```suggestion
    private isValidId(id: Id): boolean {
        return !Number.isNaN(id) || id > 0;
    }
```

bastianjoel (changes requested):

bastianjoel (approved):

rrenkert (approved):

PR #2517, opened 2023-07-12
Save sort order only on change

### LINKED ISSUE #2166: Save sort order only on change
Currently, when a list view is opened, the algorithm to check the sort order in `base-sort-list.service.ts` is as follows:
1. Try to fetch the saved sort order from the local storage
2. If not present, use the default sort order
3. Save the sort order to local storage

This should be changed so that step 3 is not executed anymore. The sort order should only be saved to the local storage if the user explicitely changes it. This means that a normal user who does not change her sorting order will always see the default sorting order, even when it is changed by the admin, which is the reason for this change.

luisa-beerboom:
Closes #2166


Elblinator (approved):
code-review

emanuelschuetze (changes requested):
Please follow all admin changes if user is in motion list and on default sorting

Elblinator (approved):
code-review

emanuelschuetze (changes requested):
admin changed default sorting -> for all users (also for users with custom sorting) the sorting changed to default.

client/src/app/site/base/base-sort.service/base-sort-list.service.ts
```diff
@@ -105,8 +122,44 @@ export abstract class BaseSortListService<V extends BaseViewModel>
...
+            option =>
+                this._defaultDefinitionSubject.value &&
+                this.isSameProperty(option.property, this._defaultDefinitionSubject.value.sortProperty)
+        );
+    }
+
+    public isSameProperty(a: OsSortProperty<V>, b: OsSortProperty<V>): boolean {
+        a = Array.isArray(a) ? a : [a];
+        b = Array.isArray(b) ? b : [b];
+        return a.length === b.length && a.every((value, index) => value === b[index]);
```
bastianjoel, line 136:
```suggestion
        return a.equals(b);
```

bastianjoel (changes requested):

Elblinator (approved):
last commit code-review

bastianjoel (approved):

bastianjoel (approved):

luisa-beerboom:
@emanuelschuetze please test if this also fixes #2650 
Given that the last commit should have introduced some stability, I believe it might actually fix it as well

Elblinator (approved):
code-review

emanuelschuetze (approved):

PR #2521, opened 2023-07-13
Fix spaces in motion pdf export

### LINKED ISSUE #2519: Fix missing space in pdf export
If a motion is exported without numbers: 
 words which are separated by a new line if numbers are visible, now are missing a space between them. 

This bug could be a result from this Issue #2486 
 
 **Example:**
 - How it is presented with numbers:  "Aenean [new line] commodo" 
 - How it is exported without numbers "Aeneancommodo"

**Wanted behaviour:**
the words should be separated by a space.

luisa-beerboom:
Closes #2519


Elblinator (changes requested):
The space is still missing :(

Elblinator (approved):
code and functinality

PR #2534, opened 2023-07-17
Fix default daterange value for meeting create

### LINKED ISSUE #2501: Meeting create: small improvement for date picking
**Description:** During a meeting create a user can set the date of the meeting. If the meeting date is only one day, the user needs click twice on the date to set it otherwise the meeting create won't save.

**Reproduction:**
precondition: a committee should exist.
1. open committes and open an existing one
2. clik on the plus-icon to create a new meeting
3. open the date field and try to set a date for a one day meeting. 

**What should happen**: If a user wants to create a one day meeting, one click on a specific date should be enough to set it. After clicking once on a date and clicking the Save button, the date field should have set the selected date as start and end date.

luisa-beerboom:
Closes #2501


Elblinator (approved):
code-review

PR #2535, opened 2023-07-17
Fix voting banner style

### LINKED ISSUE #2488: Voting banner: problem with font color and background color / small layout gap
**Description:** If an election, poll or vote is started, a voting banner appears at the top of the page. Two issue occur:

- [x] If the accent color is set to a value near white or white, the banner seems to be empty, because the color of the font is white. The font didn't "react" to the accent color.  
- [x] Addtionally a thin white line appears between the voting banner and the global headbar. 

**Example:**
Screenshot_20230704_123701

**Reproduction:**
1. open _design_ on organisations level > choose white or almost white as accent color
2. open a meeting
3. go to _participants_ and add yourself as admin to the meeting
4. open _motions_ > create a motion > create a new vote > set the group admin as entitled to vote > start vote > voting banner should appear

**What should happen:**
1. If the accent color is bright, the font color should switch to for example black. The font color should react on the accent color to avoid unreadble text.
2. The thin white line should not appear. Voting banner and global headbar should not have a gap.

luisa-beerboom:
Closes #2488

### DIFF
--- a/client/src/app/site/modules/site-wrapper/components/banner/banner.component.scss
+++ b/client/src/app/site/modules/site-wrapper/components/banner/banner.component.scss
@@ -12,13 +12,11 @@
     line-height: 20px;
     width: 100%;
     text-align: center;
-    border-bottom: 1px solid white;
 
     a {
         align-items: center;
         justify-content: center;
         text-decoration: none;
-        color: white;
         &.banner-link {
             width: 100%;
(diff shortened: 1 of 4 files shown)


Elblinator (approved):
code-review

Elblinator (approved):
function

PR #2542, opened 2023-07-20
Fix poll pdf exports

### LINKED ISSUE #2515: Analog motion votes: pdf export issue with majorities
**Description:** If a motion has an analog vote which contains only a majority selection and no values for yes/no/abstain, the pdf export will fail. The motion pdf export won't include the voting and the vote pdf export will includes strange values like -2 or -1.

**Example: Base content: two analog votes **
Screenshot_20230712_134638

**pdf export - result which includes only first result with all numbers. majority case is missing.** 
Screenshot_20230712_134742

**pdf export of analog voting: only majority -> result is not displayed. Instead of it, gibberish numbers are exported.**
Screenshot_20230712_134819

**Reproduction:**
1. open a meeting
2. open motions
3. create a motion
4. create a analog vote > select a majority for yes and save it
5. Export motion via three-dot-menu  > missing content
6. open created analog voting by clicking on it's title > open the three-dot-menu > export pdf > gibberish results.

**What should happen:**
The pdf export of analog vote results should be able to handle a single majority selection. The motion pdf and the voting pdf should include a correct result.

luisa-beerboom:
Closes #2515 

Also removed the single votes table from the pdf exports of analog polls, since it also isn't shown in the `poll-detail-view` and the information portrayed did not make sense, as was to be expected.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/base/base-poll-pdf.service.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/base/base-poll-pdf.service.ts
@@ -2,7 +2,7 @@ import { Directive } from '@angular/core';
 import { TranslateService } from '@ngx-translate/core';
 import { Id } from 'src/app/domain/definitions/key-types';
 import { BallotPaperSelection } from 'src/app/domain/models/meetings/meeting';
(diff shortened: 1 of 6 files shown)


Elblinator (approved):
code-review

Elblinator (approved):
function

PR #2543, opened 2023-07-20
Fix motion list multiselect tag dialog

### LINKED ISSUE #2508: Fix Set Tags > Add/Remove
The Remove and the Add button from "Set Tags" is doing the same: setting the selcted tag onto the motions

**Recreation::** 

1. create one motion and two tags
2. with multi select select a motion
3. use "Set Tags" 
4. and add both tags
5. now remove one of the tags, only the "removed" tag should remain
6. now add only the other tag, only the added tag should remain

**wanted behaviour:**
If tags are added they should not removed the Tags which were already selected.
If tags are delected they should be deleted.

luisa-beerboom:
Closes #2508


Elblinator (approved):
code-review

Elblinator (approved):
functionality

PR #2544, opened 2023-07-21
Add tag_ids and agenda_type to agenda item duplication

### LINKED ISSUE #2496: fix wrong and missing information of duplicated agenda items
If agenda items are duplicated the connected tags vanish and the visibility (public/intern/hidden) can be wrong in the duplicated item

**current behaviour:**

1. go into agenda 
2. create a new topic A (visibility e.g. public)
3. activate multi select and select topic A
4. add a tag and change the visibility (e.g. to hidden)
5. stay in multiselect mode and duplicate topic A 
6. now one topic A with visibility hidden and a tag, and a topic A with visibility public and without a tag exist.

**wanted bahaviour:**
if a topic is duplicated the item should have the same visibility and the same tags.

luisa-beerboom:
Closes #2496

### DIFF
--- a/client/src/app/domain/models/topics/topic.ts
+++ b/client/src/app/domain/models/topics/topic.ts
@@ -5,7 +5,6 @@ import { HasAgendaItemId } from '../../interfaces/has-agenda-item-id';
 import { HasAttachmentIds } from '../../interfaces/has-attachment-ids';
 import { HasListOfSpeakersId } from '../../interfaces/has-list-of-speakers-id';
 import { HasMeetingId } from '../../interfaces/has-meeting-id';
-import { HasTagIds } from '../../interfaces/has-tag-ids';
 import { BaseModel } from '../base/base-model';
 
 /**
@@ -30,7 +29,6 @@ export class Topic extends BaseModel<Topic> {
         `attachment_ids`,
         `agenda_item_id`,
         `list_of_speakers_id`,
-        `tag_ids`,
         `meeting_id`
     ];
 }
@@ -39,6 +37,5 @@ export interface Topic
         HasAgendaItemId,
         HasListOfSpeakersId,
         HasAttachmentIds,
-        HasTagIds,
         HasSequentialNumber,
         HasPollIds {}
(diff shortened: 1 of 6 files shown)


jsangmeister:
Well, we either restructure the client to use topic tags as well as agenda item tags or we change the backend to be able to set the agenda item tags directly, analogous to the other agenda item properties in https://github.com/OpenSlides/OpenSlides/wiki/topic.create#payload. Sooo the latter :D

The suggestion would then be to rename the `tag_ids` payload field to `agenda_item_tag_ids` and transfer these to the agenda item instead of the topic.

Furthermore, if the topic tags are not used at all, we could go ahead and remove them completely from the `models.yml`. I don't see any further ramifications in that if they are not used anywhere.

luisa-beerboom:
> Well, we either restructure the client to use topic tags as well as agenda item tags or we change the backend to be able to set the agenda item tags directly, analogous to the other agenda item properties in https://github.com/OpenSlides/OpenSlides/wiki/topic.create#payload. Sooo the latter :D
> 
> The suggestion would then be to rename the `tag_ids` payload field to `agenda_item_tag_ids` and transfer these to the agenda item instead of the topic.
> 
> Furthermore, if the topic tags are not used at all, we could go ahead and remove them completely from the `models.yml`. I don't see any further ramifications in that if they are not used anywhere.

Sounds good to me

jsangmeister:
Opened https://github.com/OpenSlides/openslides-backend/issues/1824 for this

jsangmeister:
https://github.com/OpenSlides/openslides-backend/pull/1832 was merged and `tag_ids` were removed from topics alltogether, so this can be continued and the field removed in the client as well

luisa-beerboom:
For testing:
This PR might have consequences when creating an agenda item from assignments, motions and motion block, please check if these still work.

luisa-beerboom:
> If a tag is added to a motion the tag is not carried over to the agenda, the tags from a motion should be in the agenda as well

Talked to @MSoeb 
This was not in OS3 therefore currently this is the correct behaviour and changes concerning this would be a new feature request.
Also it would honestly require quite a bit of planning and therefore it is not currently on the table

bastianjoel (approved):

Elblinator (approved):

PR #2549, opened 2023-07-24
Always show point of order category for speaker

### LINKED ISSUE #2547: Point of orders: categorie information is lost in list of speakers after deactivation of point of order
**Description:** When the point of order functions are switched off, the specification extensions disappear from the list of speakers in the autopilot. The reference frame, i. the extension with new specifications, is no longer visible. This means that it is no longer possible to attribute a speech to a specification. Information is lost.

**Example - specification and point of order is activated:**
Screenshot_20230724_102936

**Example - specification and point of order setting is disabled:**
Screenshot_20230724_103108

**Reproduction:** 
1. open a meeting
2. open settings > list of speakers: activate point of orders and specifications
3. add some specifications
4. open agenda > create a new item > project it
5. go to autopilot. > add yourself via point of order with specification to the list of speakers
6. open settings > list of speakers > point of order > deactived both settings 
7. go to autopilot and check the provided information

**What should happen:**
The specifications should always appear in the list of speakers, if a user has used it, even when the point-of-order functions are deactivated. This would allow to close the list of speakers for normal entries and point of order entries without losing information.

luisa-beerboom:
Closes #2547

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-content/list-of-speakers-content.component.html
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-content/list-of-speakers-content.component.html
@@ -363,18 +363,13 @@
         </mat-icon>
         <b
             [class.foreground-warn]="!!showcolor"
-            *ngIf="
-                pointOfOrderCategoriesEnabled &&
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
code-review

MSoeb (approved):

PR #2563, opened 2023-07-26
Fix agenda show deleted items

### LINKED ISSUE #2497: Deleted agenda items are still displayed in list when navigation back
1. Create two topic items
2. Delete one of the created items
3. Open the other topic item 
4. Navigate back

The deleted topic item is in the topic list again. When reloading the page the deleted entry disappears.

luisa-beerboom:
Closes #2497 

The problem is the following: 
1. **Open agenda item list** => `agenda_list` subscription is open
2. **Delete one item** => The autoupdate sends a single-field update with the new value for the meetings `agenda_item_ids`, which will cause the client to consider these fields to be deleted, but NOT the autoupdate shared worker, which will continue to store them
3. **Go to the topic detail view** 
4. **Go back from the topic detail view** => In the short moment where the parameters change, the detailview doesn't have an id and therefore subscribes to the `agenda_list_minimal` (as it now believes itself to be the new topic form, which needs the agenda list information) The shared worker will send the already stored data (see point two: it includes the deleted models) as the initial data for the subscription which will then be used by the client completely unfiltered
5. **See that the deleted models seem to have returned**

Honestly we should do something about this quirk of our autoupdate handling system, but I didn't really have any idea what exactly can be done so I "fixed" it this way _for now_


Elblinator (approved):
code and function

PR #2572, opened 2023-07-27
2556 wrong dialog text

### LINKED ISSUE #2556: Fix wrong text in dialog
Where to find: 
> one workflow > rename state > delete

The displayed title and content are wrong

(Wanted behaviour in the comment below)

Elblinator:
closes #2556 

additional change in the code for dialogs - these should not change any function or dialog text!!:

dialog text while 
- deleting one election in detail-view of this election
- deleting a projector (from projector -lsit-view and detail view)
- deleting tags from the orga level
- clearing a chat history


luisa-beerboom (approved):
Code review -> approved
Did some testing as well:
The workflow detail delete is basically a double-dialog. 
You open the first dialog, click on the delete-button, then a second dialog opens.
This is probably suboptimal from a design perspective, what do you think @MSoeb ?

PR #2607, opened 2023-08-09
Fix motion pdf export supporters not displayed

### LINKED ISSUE #2599: [RTF] Missing supporters in Motion export
**Description:**
If a motion has supporters and is exported. the supporters are never visible in the exported document

**Reproduction:**
1. in a meeting > settings > motions-settings > set "**number of (minimum)required supporters to a motion**" to one
2. allow your user to support motions -> participants > (three dots menu) groups > activate "**can support motions**"
3. allow to support a motion in a specific state -> motions > (three dot menu) workflows > one specific workflow > activate "allow support" (**remember the workflow and the state**)
4. set yourself to **present** 
5. create a new motion, and make sure that you are **not the submitter yourself** 
6. in the new motion put the state to the **state which can be supported**
7. support the motion
8. in motion-list > three dot menu > **export** > pdf and supporter users
9. the supporters in the pdf are missing

**Wanted behaviour:**
the supporters should be correctly exported and displayed in the pdf

bastianjoel:
resolves #2599

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/services/export/definitions.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/services/export/definitions.ts
@@ -4,7 +4,7 @@ import { ViewMotion } from 'src/app/site/pages/meetings/pages/motions';
  */
 export type InfoToExport =
     | 'submitters'
-    | 'supporters'
+    | 'supporter_users'
     | 'state'
     | 'recommendation'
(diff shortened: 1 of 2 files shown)


luisa-beerboom (approved):

Elblinator (approved):
Supporters are now exported, yay :)

PR #2616, opened 2023-08-10
Search interface improvements

bastianjoel:


Elblinator:
- [ ] Under the search field a grey smaller sentence: "To search press Enter or the magnifier"
- [ ] There should be a loading bar if it is still searching 
- [ ] The distance between the search field and the filtern should be bigger (see mock-up)
- [ ] Search field should be grey (see mock-up)
- [ ] If a the phone view is active (<=700px) then the filters should be each in their own line
- [ ] the amount of displayed results should be displayed

peb-adr:
@bastianjoel I will merge now, please extract important info and TODOs into a new issue as discussed

PR #2631, opened 2023-08-14
Fix missing projection number

### LINKED ISSUE #2443: projector-menu: fix missing number (Antragsbezeichner) and topic-number in projection waiting-list
In the projector detail view the topic-number in the waiting list is missing and additionally the number (Antragsbezeichner) vanishes if the website is reloaded by the user. 

Example of partially missing content:
topic-num

Example of missing content in waiting list after reload:
Topic-titles-projector-display

**Current behaviour:**
1. queue topics with number and motions with number (Antragsbezeichner) in the standardprojector waiting list
2. navigate into the projector menu > open detail view of standardprojector
3. reload the page
4. the numbers will vanish

**Wanted behaviour:**
all numbers should be available and should not vanish with a reload

bastianjoel:
resolves #2443 

This only fixes the not displayed motion numbers. As one does not project agenda items directly but instead topics or elections we cannot display the agenda item numbers currently.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/projectors/projectors.subscription.ts
+++ b/client/src/app/site/pages/meetings/pages/projectors/projectors.subscription.ts
@@ -7,7 +7,14 @@ import { DEFAULT_FIELDSET } from 'src/app/site/services/model-request-builder';
 
 import { ViewProjector } from './view-models';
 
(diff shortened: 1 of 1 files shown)


bastianjoel:
Note that this is not a rtf specific issue.

Elblinator (approved):
as stated in the description only fixes the motion numbers.

PR #2633, opened 2023-08-14
Fix user create

### LINKED ISSUE #2626: [RTF] Creating a new user in Accounts as orga-admin/account-admin don't work
**Description:** As an organisationsadmin oder accountadmin it is not possible to create manually or by csv file a new account. Superadmins instead can create new accounts.

By doing it manually the error "null has no properties".

By doing it via cv file the error "Fehler: You are not allowed to perform action user.create. Missing permission: Permission user.can_manage in meeting None" appears.

**Reproduction:** 
1. open an instance as superadmin
2. open accounts
3. create a new account as orgaadmin and another one as accountadmin > log out as superadmin
4. login as orgaadmin or accountadmin
5. open accounts
6. try to create a new account via plus icon as orgaadmin or accountadmin
7. try to import new users as a orgaadmin or account admin

What should happen:
Accounts with the permission to create new accounts should be able to do it.

bastianjoel:
resolves #2626

### DIFF
--- a/client/src/app/gateways/repositories/users/user-repository.service.ts
+++ b/client/src/app/gateways/repositories/users/user-repository.service.ts
@@ -159,7 +159,9 @@ export class UserRepositoryService extends BaseRepository<ViewUser, User> {
         const payload: any[] = data.map(partialUser => ({
             ...partialUser.user,
             ...partialUser.first_meeting_user,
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
code and function

Issue #2636, opened 2023-08-14
Account / participant deletion dialog - missing content


MSoeb:
**Description:** The account delete dialog and the participant delete dialog are missing important information where the user is added. 

- [ ] **For Account dialog:** Information about elections and motions is missing.
- [ ] **For participant dialog:** Information about elections, lists of speakers and motions is missing. 

**Account dialog:**
![Screenshot_20230814_160917](https://github.com/OpenSlides/openslides-client/assets/105491030/196d47bd-3a9b-40ae-b933-175204678a32)

**Participant dialog:**
![Screenshot_20230814_161030](https://github.com/OpenSlides/openslides-client/assets/105491030/0267bcee-232b-43b9-8b0d-874152fdbe9e)

**Reproduction:**
1. open a meeting 
2. open participants > create a new user
3. open motions > create a new motion with the user as submitter > project it Autopilot
4. open autopilot and add user to list of speakers > push play and stop button
5. open elections > create a new elections with the user as candidate > do election > publish it 
6. open participants > try do delete user from meeting 
7. head back to organization level
8. open Accounts > try to delete the new user account

**What should happen:**
Account dialog: The dialog should include referenced information about elections and motions where the user took part.
Participant dialog: Information about elections, lists of speakers and motions should be included. 

@emanuelschuetze: Please check, if the solution for participants is okay.

jsangmeister:
This also requires changes in the backend

jsangmeister:
Upon closer inspection, no backend changes are needed.
- The participant removal dialog does not need these informations since the user itself is not deleted, only removed from the meeting and the information is therefore not lost. The purpose warning in the account dialog is to prevent accidental deletions since then all elections etc. would just display "deleted user", which is not the case for the participant dialog.
- The informations about motions and assignments are not displayed due to a bug which I will fix.

PR #2640, opened 2023-08-15
Fix information in user delete dialog

jsangmeister:
fixes https://github.com/OpenSlides/openslides-client/issues/2636


jsangmeister:
- fixed wording: "account" vs. "participant"
- fixed header in case of participant removal
- hid left column in case of single user removal/deletion

emanuelschuetze (approved):

rrenkert (approved):

PR #2645, opened 2023-08-16
Testing action service

luisa-beerboom:
Wrote tests for all non-deprecated functions as well as the create function.
Edited and added a comment for the action service's `beforeActionFn` feature that makes it possible to turn off the ability to send actions.


bastianjoel (approved):

bastianjoel:
closes #2511

PR #2686, opened 2023-08-22
Fix add-to-meetings autoupdates

### LINKED ISSUE #2679: Missing autoupdates for add to meetings > select meetings
**Current behaviour:**
If the meetings are not pre-loaded while an account is added to meetings then the "select meetings"-list is empty

**Recreation:**
1. Navigate to accounts > accounts > add to meeting
2. reload page
3. open select meetings-list 
4. the list should be empty

**Wanted behaviour:**
the list should not be empty, even if the meetings are not preloaded

luisa-beerboom:
Closes #2679

### DIFF
--- a/client/src/app/site/pages/organization/pages/accounts/pages/account-detail/components/account-detail-main/account-detail-main.component.ts
+++ b/client/src/app/site/pages/organization/pages/accounts/pages/account-detail/components/account-detail-main/account-detail-main.component.ts
@@ -1,7 +1,5 @@
 import { Component } from '@angular/core';
-import { filter, firstValueFrom, map } from 'rxjs';
 import { BaseModelRequestHandlerComponent } from 'src/app/site/base/base-model-request-handler.component';
-import { getMeetingListSubscriptionConfig } from 'src/app/site/pages/organization/organization.subscription';
 
 import { getAccountDetailSubscriptionConfig } from '../../../../accounts.subscription';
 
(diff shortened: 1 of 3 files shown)


client/src/app/site/pages/organization/pages/accounts/pages/account-detail/components/account-detail-main/account-detail-main.component.ts
```diff
@@ -13,6 +14,7 @@ export class AccountDetailMainComponent extends BaseModelRequestHandlerComponent
         if (params[`id`] !== oldParams[`id`]) {
             const id = +params[`id`];
             this.subscribeTo(getAccountDetailSubscriptionConfig(id), { hideWhenDestroyed: true });
+            this.subscribeTo(getMeetingListSubscriptionConfig(), { hideWhenDestroyed: true });
```
bastianjoel, line 17: This subscription is potentially quite big therefore `hideWhenDestroyed` should not be added to prevent reopening this multiple times. Also it would be better to open this subscription only when the add to meeting page was opened. 

You could also add a new subscription config that subscribes less fields and does not subscribe to archived meetings.

bastianjoel (changes requested):

bastianjoel:
Adding via the multiselect in the account list does not work either. Maybe you could fix this here too.

bastianjoel (approved):
It would be better to replace the `os-list-search-selector` with `os-repo-search-selector` and directly pass the needed subscription there but for now I am okay with your solution.

Elblinator (approved):
function

PR #2698, opened 2023-08-24
Build in referring motions sorting in pdf

### LINKED ISSUE #2685: Fix sort of "referring motions" in pdf
**Current behaviour:**
If a motion with several referring motions is printed, the order of the referring motions follows the wrong order

**Reproduction:**
in meeting:
1. settings > motion settings > set a name for: name of recommender
2. create a one motion A in a complex workflow and 
3. create a Motion B with the number (Bezeichner) 5
4. create a Motion C with the number (Bezeichner) 3 
5. workflows > complex workflows > activate "Show recommendation extension field"
6. Let B and C refer to A 
7. Print A as pdf
8. the referring motions in A are printed as: 5,3 

**Wanted behaviour:**
the order in which the referring motions is sorted should be the number (Bezeichner) and not something else

luisa-beerboom:
Closes #2685

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
@@ -331,15 +331,7 @@ export class MotionPdfService {
         if (!infoToExport || infoToExport.includes(`referring_motions`)) {
             if (motion.referenced_in_motion_recommendation_extensions.length) {
                 const referringMotions = motion.referenced_in_motion_recommendation_extensions
-                    .sort((a, b) =>
-                        a.number
-                            ? b.number
-                                ? a.number.localeCompare(b.number)
-                                : -1
-                            : b.number
-                            ? 1
-                            : a.title.localeCompare(b.title)
-                    )
+                    .naturalSort(this.translate.currentLang, [`number`, `title`])
                     .map(motion => motion.getNumberOrTitle())
                     .join(`, `);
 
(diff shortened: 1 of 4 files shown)


client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
```diff
@@ -331,6 +331,15 @@ export class MotionPdfService {
...
+                    .sort((a, b) =>
+                        a.number
+                            ? b.number
+                                ? a.number.localeCompare(b.number)
+                                : -1
+                            : b.number
+                            ? 1
+                            : a.title.localeCompare(b.title)
+                    )
                     .map(motion => motion.getNumberOrTitle())
```
jsangmeister, line 343: Can't you just sort after this map? Then you don't have to differentiate whether the motions have numbers or not

luisa-beerboom: If I want motions without numbers to end up at the end of the list, I'll either have to do this or use strange string operations in order to guess whether what I'm getting is a number or a title

jsangmeister: ah, okay, did not know this was a requirement

Elblinator (changes requested):
The sort algorithm is not working correctly if the numbers have different lengths: 
referring_moitons

luisa-beerboom:
> The sort algorithm is not working correctly if the numbers have different lengths: referring_moitons

These are strings and can contain letters. Since these are strings "19" is smaller "3" which is smaller than "A12". It would be possible to sort in the strings that are pure numbers by their number value at the front and then put the strings containing letters behind that, but that'll be finicky and it's not as if it's done anywhere else

Elblinator:
if I sort the motion list via number (bezeichner)  then they are correctly sorted:
sort_bezeichner

jsangmeister:
> > The sort algorithm is not working correctly if the numbers have different lengths: referring_moitons
> 
> These are strings and can contain letters. Since these are strings "19" is smaller "3" which is smaller than "A12". It would be possible to sort in the strings that are pure numbers by their number value at the front and then put the strings containing letters behind that, but that'll be finicky and it's not as if it's done anywhere else

There are also sort algorithms which can cope with that (called "natural sort"). If the motion list is already correctly sorted, it seems like we already have a solution for this, but otherwise I'm sure there are packages implementing this and it is what users expect, after all.

Elblinator (approved):

PR #2718, opened 2023-08-28
Cleanup worker cache on delete

### LINKED ISSUE #2497: Deleted agenda items are still displayed in list when navigation back

bastianjoel:
resolves #2497 

The worker will now be notified if the client deletes models from the modelstore and will delete them within its internal state too. 
Note that in order to test properly this the shared worker needs to be enabled. 

This PR could have some side effects especially if there are misconfigured entries in `relations.ts` or other bugs that result in models being accidentally deleted.


luisa-beerboom (approved):

Elblinator:
If you delete a topic in topic detail, then the topic still only disappears with a reload or when you go into the detail view of this deleted topic again. But that is a different bug, created at a different place

Elblinator (approved):

Elblinator:
the additional problem is now in #2731

PR #2723, opened 2023-08-29
Improve user deletion dialog

### LINKED ISSUE #2656: Change of wording and appearance of Dialog
**Current appearance of delete dialog:** 
delete_dialog

**Wanted changes:**
- the sentence "These accounts will be removed [...]" should be changed to "These participants will be removed [...]"
- if some participants are removed AND some accounts are deleted then the sentence "These acocunts will be deleted:" should have a margin to the last item above the sentence

**Example appearance of change delete dialog:**
new_delete_dialog

jsangmeister:
fixes #2656

### DIFF
--- a/client/src/app/site/modules/user-components/components/user-delete-dialog/user-delete-dialog.component.html
+++ b/client/src/app/site/modules/user-components/components/user-delete-dialog/user-delete-dialog.component.html
@@ -21,7 +21,7 @@
                     </ng-container>
                 </cdk-virtual-scroll-viewport>
             </div>
-            <div *ngIf="hasUsersToRemove" class="user-list-wrapper">
+            <div *ngIf="hasUsersToRemove" class="user-list-wrapper spacer-top-20">
                 <p class="section-title flex-vertical-center">
                     {{ 'These participants will be removed:' | translate }}
                 </p>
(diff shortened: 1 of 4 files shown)


Elblinator (dismissed):
If I try to remove participants and delete accounts, following error occurs:
```
Error: Datastore service sends HTTP 400. The following locks were broken: 'group/7/meeting_user_ids', 'group/8/meeting_user_ids', 'meeting_user/8/meeting_id', 'meeting_user/8/user_id', 'meeting_user/9/meeting_id', 'meeting_user/9/user_id'
```
Only deleting or only removing works

Yes, delete (and Yes, remove) has an additional empty space: "Yes , delete"

The sentence "These participants will be removed:" needs a bigger margin if the list above is full

jsangmeister:
Fixed the layout comments. The  broken locks error already existed beforehand. The reason is that the request is being sent to the `handle_request` route instead of the `handle_separately` route like it should be. But here: https://github.com/OpenSlides/openslides-client/blob/36888d9edf2a4c2622821ccbb5570b9b2a133d75/client/src/app/site/pages/meetings/pages/participants/services/common/participant-controller.service/participant-controller.service.ts#L233
the parameter is correctly given and should also be correctly kept in the `concat` method, also I suppose that the error happens somewhere in there. Unfortunately I'm currently not able to pin it down in the short amount of time. @luisa-beerboom @bastianjoel maybe you can have a look?

jsangmeister:
Also, the playwright tests were failing before. We will see if they fail again, but I don't know why they failed in the first place, as I only changed layout and style-related things.

luisa-beerboom (approved):

jsangmeister:
Since this is a separate bug which already existed beforehand, I would suggest to merge this if you have no more remarks @Elblinator and open a new issue for the request error.

Elblinator (approved):

PR #2726, opened 2023-08-30
Improve shared worker reliability

### LINKED ISSUE #2706: Shared worker reliability improvements
Currently we face some issues regarding reliability of the shared worker. In some scenarios like waking up a computer from sleep mode the worker seems to get into an unresponsive state where we do not get any autoupdates. We are not entirely sure what exactly cases these issues but in the past it was already observed that especially chrome sometimes gets into a state where it is not possible to connect to a shared worker. Although the methods calls do not return errors. 

Therefore we want to implement a fault detection and improve fault recovery. 

## Fault detection
This will be done by introducing a health check and message acknowledgements to the shared worker. 
1. The shared worker service should check periodically if the worker is still available.
2. Some actions (especially long running ones like opening au connections) should get a quick acknowledgement. If this is not received within a certain time a recovery procedure should be run. 

## Fault recovery
Currently only the internal state of the shared worker is reset when reloading the page. This should be replaced by a real restart of the shared worker. 

To do so a new event will be sent from the shared worker indicating a termination. After that the clients need to reconnect to the worker and reopen their autoupdate connections. 

For this we have to take some problems into account. 
1. In the past we had concerns about this resulting in potentially a lot of autoupdate connections opened at the same time if the user has a lot of different tabs open. As a countermeasure a browser tab should only reopen its autoupdate connections [if it is](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) or when [gets in focus](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event). 
3. Restarts could happen at any time and are especially likely on app boot. This could have some side effects with the current au connection opening process. 
4. After restart a context switch is possible (directly running in window/shared worker). We might need to do experiments how long it takes for the shared worker to terminate and be accessible again. 
5. We might need mechanisms to prevent restart loops.
6. When the worker is running directly inside the window it is not possible and necessary to restart it.

bastianjoel:
resolves #2706

This PR also drops jsdom support for the karma tests as they do not work properly anymore.


luisa-beerboom (approved):

Elblinator (approved):
I tested it to the best of my ability, it would be better if someone else tests it as well

bastianjoel:
While connections are getting reopened a visual feedback should be shown so the user knows that the client might be out of sync.

emanuelschuetze (approved):

PR #2734, opened 2023-08-31
Revert daterange picker in meeting edit

bastianjoel:
reverts changes from #2361


rrenkert (changes requested):
* Using the clear button in one of the date pickers freezes the browser tab with the following error:
TypeError: Cannot read properties of null (reading 'start') at get empty [as empty] (datepicker.component.ts:18:28)

* Setting the start date after the end date is possible and saving has no effect and no error is visible to the user.

* And there is a visual artifact on entering or reloading the page:

date-settings

PR #2745, opened 2023-09-01
Make pdf and csv export respect the meetings export settings

### LINKED ISSUE #2591: Fix general-settings in meetings
**Description:**
Even when some fields in general-settings are changed, OpenSlides does not change behaviour.

**Problem-fields:**
- Separator for CSV-Export: independent from the decided separator, the csv Export is using commas
- Page-format for pdf-format: as far as I can tell, the exported pdfs are identical minus one problem; the title for a motion vanished in the A5 version:
Bildschirmfoto_2023-07-31_15-40-43

**Wanted behaviour:**
- the separator for CSV should follow the character decided on in general-settings
- The pdf-format  should be different dependent on the general settings page-format, it seems that Openslides already has rules to follow for A5 (search for pageSize in client), but I would suspect that they are currently not activating. 
- The title from motions should not vanish in the A5 version

luisa-beerboom:
Closes #2591

### DIFF
--- a/client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
+++ b/client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
@@ -458,7 +458,7 @@ export class PdfDocumentService {
         docDefinition,
         filename: filetitle,
         ...config
-    }: DownloadConfig & { pageMargins: [number, number, number, number] }): void {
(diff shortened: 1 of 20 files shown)


Elblinator (changes requested):
the Agenda is still not following the correct csv-separator and encoding type

Everything else is working great :)

bastianjoel (approved):

Elblinator (approved):

PR #2772, opened 2023-09-07
Make sort lists not update during drag/drop

### LINKED ISSUE #2622: LoS: Moving speaker bug
**Description:** If a user moves a speaker in the list of speakers while a delegate adds themselves to the list, an UI error occurs.

**Example of broken UI**
Screenshot_20230811_141444

**Reproduction:**
precondition: an admin and a user with permission to add himself to the LoS is needed. Also a second device is needed.
1. open the meeting as superadmin
2. open the agenda > create a topic
3. open participants > create four participants as delegates
4. open the autopilot > add three of them to the LoS
5. Open on SECOND device the meeting as a delegate
6. open autopilot
7. Now you need both devices: On first devices as admin open Autopilot, grab a delegate in the LoS and move him to another position. IMPORTANT - Hold this! - Switch to the other device and add the delegate to the LoS. - Switch back to devices one: Move the position of the delegate and release it. -> UI bug.

**What should happen:**
The interaction between adding a user to the LoS and moving a user in the LoS should not result in an UI bug.

luisa-beerboom:
Closes #2622 

Made the drag/drop events stop updates from appearing on the list while the dragging is taking place.

Items are moved to the exact list index where they were dragged upon regardless of where the other items move meanwhile.

Please test thoroughly as
1. It might look a bit weird if items are removed while dragging is taking place (especially when the item that is being dragged is removed)
2. I had to amend the multiselect logic as well, but I didn't test it, so please test that especially


client/src/app/ui/modules/sorting/modules/sorting-list/components/sorting-list/sorting-list.component.ts
bastianjoel, line 1: I would prefer the use of our `Mutex` class instead of using the `draggingFinished` with a `BehaviorSubject`.

bastianjoel (changes requested):

bastianjoel (approved):

Elblinator (changes requested):
the behaviour while adding a normal participant is correct. 

A new PoO should always be at the top of the list, even if a participant is sorted to the top while the PoO is added.
At the moment a PoO is at the second place if it was added while moving a participant to the top

luisa-beerboom:
As you didn't provide what _should_ actually happen, I did some conceptualizing.

I could try to implement this:

The program will look if there's points of order at the top of the list, if there are it will mark the area at the top of the list where there are only points of order, both for the list as it was when the dragging started (interval A), and as it should be if there was no dragging (interval B).
Then the new position of the dropped item is calculated as follows:
- Case 1: If the drop position is in interval A: Just drop it there, the user clearly intended to sort the item in between the points of orders anyway
- Case 2: If the drop position is not in interval B: Drop it there as it doesn't get in the way of any points of order anyway
- Case 3: If the drop position is in interval B, but not in A: Drop it right below interval B, as the operator did not clearly intend to move it in between the points of order.

@Elblinator would that work?

luisa-beerboom:
If the item that is moved is a point of order I would stick it onto the bottom of the priority area if it seems like it wasn't purposefully moved out of it

Elblinator (approved):
function

bastianjoel (approved):

PR #2773, opened 2023-09-07
Update build-and-test-dev-image to use localprod

bastianjoel:
This is an attempt to make the build-and-test-dev-image more reliable by using localprod which builds the client when building the images instead of when starting the setup which should prevent the timeout which currently occurs quite often. 

Also included is a performance improvement for initial loading when the service worker is not yet registered which was necessary to make the tests work. In that case waiting for the service worker to download a new application version is skipped as there is none installed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,11 +74,3 @@ jobs:
 
       - name: Run karma tests
         run: make run-karma-tests
-
-  build-prod-image:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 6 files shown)


bastianjoel:
In my opinion we could remove the `build-prod-image` job with this. 
Also the requirement for `build-and-check-dev-image` needs to be disabled and in case we want to keep it `build-prod-image` needs to be set to required.

rrenkert (approved):

PR #2774, opened 2023-09-08
Fix single option chart table style

### LINKED ISSUE #2499: Poll results need improvement
**Description:** Poll results have some layout issues if the browser is not used in full screen.

- The result text has an unsightly line break. The percent sign slips into a new line. 
- The graphic is too close to the results and should be displayed aligned with the texts, as in fullscreen mode.

Example:
Screenshot_20230706_122406

**Reproduction:**
1. open a meeting
2. open motions
3. create a motion
4. change state until vote is possible. -> vote

Hint: It would be good to have several users in the participant list. They can be used to generate poll results with three decimal places.

**What should happen:** The layout of the poll result should be improved. It should be possible to use smaller browser windows and see nice poll results. 

@Eich-hoernchen : Could you please make a suggestion to beautfiy this?

luisa-beerboom:
Closes #2499

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.scss
+++ b/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.scss
@@ -9,10 +9,10 @@
     display: flex;
     column-gap: 2em;
     flex-flow: row wrap;
+    justify-content: center;
 
     .result-table {
-        flex-basis: 150px;
-        flex-grow: 2;
+        width: 240px;
         th {
             text-align: right;
             font-weight: initial;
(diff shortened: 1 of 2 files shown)


bastianjoel (approved):

Elblinator (changes requested):
If all vote amounts are <10: the percent is in two lines between 556px and 465px width
If at least one vote amount is >=10: the percent is in two lines between 1335px and 965px width

The percentage should stay in one line

luisa-beerboom:
Fixed that. Also changed the motion detail view layout a bit so the table won't flow out of bounds if the card gets to small.

Elblinator (approved):
function

bastianjoel (approved):

PR #2777, opened 2023-09-08
Fix projector preview

### LINKED ISSUE #2702: Fix projector preview
**Current behaviour:**
The content which is projected is not visible in the preview of the projector

**Reproduction:**
1. project something onto the projector 
2. open the "edit projector" dialog
3. content is not displayed in the preview

**Wanted behaviour:**
The content should be displayed in the preview

luisa-beerboom:
Closes #2702

### DIFF
--- a/client/src/app/site/pages/meetings/modules/projector/components/slide-container/slide-container.component.ts
+++ b/client/src/app/site/pages/meetings/modules/projector/components/slide-container/slide-container.component.ts
@@ -145,7 +145,7 @@ export class SlideContainerComponent {
      */
     private updateScroll(): void {
         if (this.slideOptions.scrollable && !this.slideIsAutonomic) {
-            let value = this.scroll;
+            let value = this.scroll ?? 0;
             value *= -25;
             if (this.projector && this.projector.show_header_footer) {
                 value += 50; // Default offset for the header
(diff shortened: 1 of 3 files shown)


bastianjoel (approved):

Elblinator (changes requested):
the projection is too far up if "show header and footer" is activated

luisa-beerboom:
> the projection is too far up if "show header and footer" is activated

Two things:
1. I can't see/replicate this, so I would probably need a better replication description or a picture or something
2. All I did was fix the projection data in order to make the preview work again, I did not change anything about the preview projector itself, which is the same component as the normal projector. Unless this "projection being too far up"-thing happens _only_ in the preview and _not_ in the normal projector, I'd say that is a separate bug.

@Elblinator  Please take a closer look at this and act accordingly

Elblinator:
This indeed only happens in the preview: 

the actual projector(same in projector-detail):
projector

projector preview:
projector_preview

projector preview without the header:
projector_preview_without

It happens with an election, a topic, the agenda, a LoS

**Additionally** png files in the preview are only a black field

Elblinator:
the files still don't work

Elblinator (approved):
function

bastianjoel (approved):

PR #2778, opened 2023-09-08
Notify via icc if message was read

### LINKED ISSUE #2461: Change chat "new" messages behaviour
If OpenSlides is opened in several tabs (Tab A and Tab B) and the chat has a new message the new message is displayed as new in all tabs. If the message is viewed in tab A, tab B does not update the chat and still displays the messages as new. 

**Wanted behaviour:**
If the message is viewed in one tab every other open tab should then stop displaying the message as new

bastianjoel:
A non client side solution to resolve #2461.

This uses the icc service to update the last read status. 

It would be possible to reduce the icc request by validating if actually a message was read before updating the last read times. On the downside this would reduce the likelihood devices (not tabs or windows on the same machine) are synced. Tell me if this is wanted.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/chat/services/chat-notification.service.ts
+++ b/client/src/app/site/pages/meetings/pages/chat/services/chat-notification.service.ts
@@ -1,7 +1,9 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, Observable } from 'rxjs';
 import { Id } from 'src/app/domain/definitions/key-types';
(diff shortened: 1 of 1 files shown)


luisa-beerboom (approved):

rrenkert (changes requested):
Does not work for me with two browser windows. Seems the second client does not receive the icc message.

Elblinator:
this should be fixed in a new PR and with a bigger/clearer change

PR #2783, opened 2023-09-11
Make motion multiselect csv/pdf export sort by current sorting

### LINKED ISSUE #2780: Motion export via multiselect should sort data
If a motion export is done via multiselect, the sorting is standard, it should be according to the selected list sorting.

luisa-beerboom:
Closes #2780


bastianjoel (approved):

emanuelschuetze (changes requested):
* should also work for amendment-list
* change amendment-list default sorting to main-motion-and-linenumber

emanuelschuetze (changes requested):
amendment list does not work -> always the same order in pdf export (by id)

Elblinator (approved):
Code

PR #2788, opened 2023-09-12
Fix vote delegation

### LINKED ISSUE #2710: Missing autoupdate for delegated votes
**Current behaviour:**
If the user is not an admin and the participants are not preoloaded, then every delegated person shows the same error: You do not have the permission to vote.

**Recreation:**
1. create two (delegate) participants (A and B)
2. delegate B's vote to A
3. start a poll 
4. login as A and open the poll
5. you should see the Error: You do not have the permission to vote.
6. If you can vote for B, reload the page
7. in order to vote for B, open the participants-list once

**Wanted behaviour:**
The error should not occur, and delegates should be able to use the delegated votes without preloading the participant-list

luisa-beerboom:
Closes #2710

### DIFF
--- a/client/src/app/site/services/operator.service.ts
+++ b/client/src/app/site/services/operator.service.ts
@@ -665,8 +665,13 @@ export class OperatorService {
                         follow: [
                             {
                                 idField: `vote_delegations_from_ids`,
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
function

PR #2790, opened 2023-09-12
Fix supporter edit in motion edit

### LINKED ISSUE #2669: Fix missing user_ids for supporters
**Description:**
If supporters are added by an admin (not by the supporter themselves) then the the supporter's names but different names are displayed

**How to recreate:**
In a metting: 
1. settings > motions > supporters > Number of (min) required supporters [..] has to be one or bigger
2.  motions > New motion > select some participants as supporters, save motion
3. look at the lsit of supporters from the motion

**Send request from an updated motion:**
motion update

Ths as a supporter displayed account is not in the meeting and has the id: 2923

**Wanted behaviour:**
The correct supporters should be displayed in a motion

luisa-beerboom:
Closes #2669


Elblinator (approved):
function

bastianjoel (approved):

PR #2801, opened 2023-09-14
Fix scrolling to change recommandation

### LINKED ISSUE #2800: Fix jump marker for motions
**Current behaviour:**
If a motion has a change recommendation (different from amendment) then in the summary of changes the link are supposed to scroll to the line in question. This doesn't work at the moment

**Recreation:**
1. create a motion with a long text 
2. add in a line a change of recommendation
3. at the top press on the link "Line <number> (Change recommendation)
4. nothing happens

**Wanted behaviour:**
if you press the link you should jump to the line the change recommendation is indicating

Note: This is a very similar behaviour to the "Go to line" behaviour from motions

bastianjoel:
resolves #2800


Elblinator (approved):
function

PR #2805, opened 2023-09-14
Calculate sorted motion supporter list

luisa-beerboom:
Closes #2799 

Please check the perfomance especially

PR #2809, opened 2023-09-15
Revert "Calculate sorted motion supporter list (#2805)"

peb-adr:
This reverts commit 8828b280437f4c312890c6f241b82d7effcd72c0.


peb-adr:
I don't know the details, but unfortunately it seems #2805 did not work as intended, so it won't be taken in for this update

PR #2810, opened 2023-09-15
Update angular

### LINKED ISSUE #2345: Replace libraries that use ViewEngine
`ViewEngine` libraries are deprecated with angular 16. To make an update easier we should update or replace libraries that currently use it. 

The following libraries are affected:
- @biesbjerg/ngx-translate-extract-marker https://github.com/biesbjerg/ngx-translate-extract-marker.git
- ngx-material-timepicker https://github.com/Agranom/ngx-material-timepicker.git (update available)
- ~~ngx-papaparse https://github.com/alberthaff/ngx-papaparse (update available)~~

~~Timepicker replacement:~~
- ~~https://github.com/tonysamperi/ngx-mat-timepicker (does not support material legacy components)~~

Extract marker:
This library is so small that we should just implement its functionality ourselves in a helper function.

### LINKED ISSUE #2755: Angular 15/16 update
We are currently still on version 14 and should update in the near future. 

Requires #2345

bastianjoel:
Legacy components seem to work quite well by now. I suggest migrating them later step by step. 

resolves #2755 
resolves #2345


bastianjoel:
Before merging the `ngx-translate-extract` libraries need to be replaced by the new packages.

Elblinator (changes requested):
1. In the lower and bigger part of the global headbar: the font is not OS-Font but Roboto
2. On the Orga-level in the edit meeting view: the language setting is not disabled, even though it should be disabled
3. If the date-range of a meeting is changed (on orga- and meeting-level) then the ui is not correctly displayed

PR #2811, opened 2023-09-15
Fix committee meeting list user count

### LINKED ISSUE #2791: Removing Participant from Meeting is not refected in User count of meeting inside the committee view of meetings
Talked with Luisa this seems to be a client problem.
Steps to reproduce:
- Add an existing User to a meeting
- "delete" the user form meeting
- the participant list inside meeting is empty again
- go back to the committee showing all its meetings the user is still counted

luisa-beerboom:
Closes #2791


Elblinator (commented):
additional bug: 
the number displayed in meetings is wrong if a participant is removed from the meeting; if the account is deleted from the meeting then the number is correct)

bastianjoel (approved):

PR #2815, opened 2023-09-18
Fix participant-create-wizard

### LINKED ISSUE #2787: Prevent possibility to edit "new participant"
**Current behaviour:**
If an admin adds an **account** to a meeting and changes the personal information (without having the right to do so) then the following Error is thrown and the account is not added as participant:
```
Error: You are not allowed to perform action user.update. Missing permissions: OrganizationManagementLevel can_manage_users in organization 1 or CommitteeManagementLevel can_manage in committee 10 or Permission user.can_manage in meeting 56
```

**Recreation:**
1. create an admin (**not** super-/orga-/accountadmin) who is only admin in one meeting (meeting A)
2. create a participant (C) in meeting B
3. With the admin in meeting A try to add C (select C in step 2)
4. go to step 3 and change your present status from present to not present to present 
5. and change/add some information
6. press save 
7. the error is thrown

if you were not able to change/add any information in 4. then do not add C and reload the participant-list and try again

**Wanted behaviour:**
a normal admin should not be allowed to change any information from an already existing account, _except if the admin is an admin in ALL meetings from this account_.
The personal information-fields of such accounts should be disabled

**What the add page most of the time looks like:**
add_account

(edited after revelation that the bug needs an autoupdate to make changes possible, for the whole discussion see #2815 )

luisa-beerboom:
While trying to replicate #2787 (which I didn't manage to do at first) I instead found that the meeting-specific information form didn't appear for a meeting admin trying to add a user from another meeting.

I fixed that one.

Also fixes #2787

### DIFF
--- a/client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
@@ -283,7 +283,7 @@ export class ParticipantCreateWizardComponent extends BaseMeetingComponent imple
 
     private async checkScope(): Promise<void> {
         if (this._accountId) {
-            this._isUserInScope = await this.userService.isUserInSameScope(this._accountId);
+            this._isUserInScope = await this.userService.hasScopeManagePerms(this._accountId);
             if (this._isUserInScope && this.account?.id !== this._accountId) {
                 this.account = new User(
                     (await this.modelRequestService.fetch(getParticipantDetailSubscription(this._accountId)))[`user`][
--- a/client/src/app/site/services/operator.service.ts
+++ b/client/src/app/site/services/operator.service.ts
@@ -15,6 +15,7 @@ import { Deferred } from '../../infrastructure/utils/promises';
 import { GroupControllerService } from '../pages/meetings/pages/participants';
 import { ActiveMeetingService } from '../pages/meetings/services/active-meeting.service';
 import { NoActiveMeetingError } from '../pages/meetings/services/active-meeting-id.service';
+import { MeetingControllerService } from '../pages/meetings/services/meeting-controller.service';
 import { ViewMeeting } from '../pages/meetings/view-models/view-meeting';
 import { AuthService } from './auth.service';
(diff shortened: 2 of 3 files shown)


bastianjoel (approved):

Elblinator (commented):
The meeting specific information is still missing. 
Additional information: 
the meeting specific information is only missing in the dev-environment
I could not reconstruct this behaviour in production-environment

to trigger #2787 you need an autoupdate. 
e.g. be at the broken mask and then set yourself present or not present

luisa-beerboom:
> The meeting specific information is still missing. Additional information: the meeting specific information is only missing in the dev-environment I could not reconstruct this behaviour in production-environment

It is fixed for me in firefox and chromium, what exactly are you doing?

> 
> to trigger #2787 you need an autoupdate. e.g. be at the broken mask and then set yourself present or not present

Doesn't work for me. The personal information form flickers a bit, but there's no permanent enabling. Does this only happen in prod or smth?

Elblinator:
> > The meeting specific information is still missing. Additional information: the meeting specific information is only missing in the dev-environment I could not reconstruct this behaviour in production-environment
> 
> It is fixed for me in firefox and chromium, what exactly are you doing?
> 
> > to trigger #2787 you need an autoupdate. e.g. be at the broken mask and then set yourself present or not present
> 
> Doesn't work for me. The personal information form flickers a bit, but there's no permanent enabling. Does this only happen in prod or smth?

**preparation:**
1. make services-to-main
2. checkout fix-participant-create-wizard
3. make run-dev
4. autoupdateOnSharedWorker: false

**in localhost:**
1. create participant X in meeting A
2. login as (admin) Y, Y is an admin in meeting B, 
3. put yourself on present
4. start creating a new participant, select participant X in step 2
6. put yourself from present to not present to present
7. the following is my view:
personal_page

luisa-beerboom:
Can't replicate this, unless X is also in another meeting. Please mention such things.

bastianjoel (approved):

rrenkert (commented):
The form is no more editable but on every change of the present status the form fields flicker from disabled to enabled and back to disabled.

luisa-beerboom:
> The form is no more editable but on every change of the present status the form fields flicker from disabled to enabled and back to disabled.

That is a pre-existing bug. It should not be considered as part of this PR. Please create a separate issue.

rrenkert (approved):

PR #2818, opened 2023-09-18
Fix motion-with-amendment dialog

### LINKED ISSUE #2676: Change delete-dialog for motion with amendment
**Current behaviour:**
If a motion with amendments is deleted then the normal deletion dialog is presented and (if deleted) the amendments lose their motion without warning.

**Delete-dialog form motion with and without amendment:**
delete-motion

**Wanted behaviour:**
If a motion with amendments is deleted, then the participant should be warned that the motion has amendments and that they will lose their parent item. The amendments should not be deleted together with the motion.

**Example for possible dialog:**
delete-motion-warning

Elblinator:
closes #2676


bastianjoel (approved):

emanuelschuetze (changes requested):
we need a better list of amendments - in a separate paragraph and short the list by using number of amendments (instead of number AND title). See screenshot.

See example with current PR:
amendment-dialog1

Desired formatting:
amendments-dialog2

PR #2824, opened 2023-09-19
Minor cleanup

jsangmeister:
also fixed the missing chess functionality

some deduplication still necessary

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -53,7 +53,7 @@ module.exports = {
                     }
                 ],
                 '@typescript-eslint/quotes': ['error', 'backtick', { 'avoidEscape': false }],
-                '@typescript-eslint/no-unused-vars': ['error', { 'argsIgnorePattern': '^_' }],
+                '@typescript-eslint/no-unused-vars': ['error', { 'argsIgnorePattern': '^_', 'varsIgnorePattern': '^_' }],
                 'simple-import-sort/imports': 'error',
                 'simple-import-sort/exports': 'error',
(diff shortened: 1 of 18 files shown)


FinnStutzenstein (approved):

emanuelschuetze (approved):

m-schieder (approved):

client/.eslintrc.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
                     }
                 ],
                 '@typescript-eslint/quotes': ['error', 'backtick', { 'avoidEscape': false }],
-                '@typescript-eslint/no-unused-vars': ['error', { 'argsIgnorePattern': '^_' }],
+                '@typescript-eslint/no-unused-vars': ['error', { 'argsIgnorePattern': '^_', 'varsIgnorePattern': '^_' }],
```
bastianjoel, line 56: Is there a reason why this was added?

jsangmeister: Because of this line: https://github.com/jsangmeister/openslides-client/blob/e0954aa511eac613198b467eaeade5c480f3bd82/client/src/app/ui/modules/sidenav/modules/easter-egg/components/base-game-dialog/base-game-dialog.ts#L118

Maybe there is a better way to structure this functionality where it would not be needed to ignore the var, but I think this is fine and generally allowing unused underscore variables as well.

bastianjoel: While I think in that case `return this.startGame(notify.message)[1];` might have been okay too unpacking arrays is indeed a valid use case I did not thought about for that.

PR #2838, opened 2023-09-28
Fix set submitter containing non meeting users

### LINKED ISSUE #2568: Prevent superadmin to see non participants in motion > multi select > set submitters
**Description:**
as superadmin: in motions/amendments > multi-select > set submitters 
if the list of submitters is opened all accounts are displayed and every account can be seleced, even accounts which are not participants of the current meeting.

**Current behaviour:**
as superadmin: in motions/amendments > multi-select > set submitters 
every account can be added (and afterwards be removed) as a submitter, without adding them to the meeting as participants.

**Wanted behaviour:**
The list for the potential submitters should be the participant's list and not the accounts's list. 
the superadmin should not see accounts who are not in the meeting

bastianjoel:
resolves #2568


luisa-beerboom (approved):
Fix linting

emanuelschuetze (approved):

PR #2853, opened 2023-10-04
Fix countdowns over an hour

### LINKED ISSUE #2852: Countdown can not display durations more than a hour
Currently it is not possible to create a projector countdown that has a duration over a hour. The remaining hours are not displayed.

bastianjoel:
resolves #2852 

This still needs some discussion or info on how to display larger countdowns. With this PR only minutes are shown. Currently the padding to the left is too small with more than three digit long minutes. We might need to refactor the css part of the countdown to fix that.


Elblinator (approved):
it is noted and so far correct that the number is not centre of the box anymore (if min is bigger 99). There is already an Issue regarding this: #2940

Elblinator (approved):
code :D

PR #2856, opened 2023-10-05
Fix double click needed to remove filter

### LINKED ISSUE #2357: Can't remove current filters via buttons in filter-bar sometimes

bastianjoel:
resolves #2357 

I consider this a workaround. It seems when blurring focus of a checkbox from the filters a mousedown event is fired which prevents the click event on the filter from getting fired. I could not figure out why this happens. Especially why only on the participant list.


bastianjoel:
An additional observation was that the `Delegation of vote` filters were not removable at all via the `X` whereas the other filters just needed 2-3 clicks.

luisa-beerboom (approved):

Elblinator (approved):

PR #2861, opened 2023-10-06
Fix move in call list functionality

### LINKED ISSUE #2695: Fix function from motions > multiselect > move in call list
**Current behaviour:**
If any motion is selected via multiselect and then you try so move it in the call list errors are thrown

**Reproduction:**
in a meeting:
1. create two motions 
2. via multiselect > select one of the motion 
3. activate move in call list 
4. select the other motion as reference 
5. and then press either "set as parent" or "insert after"
6. for set as parent the error:_Error: data.tree[0].children[0].children must contain at least 1 items_ 
7. for insert after the Error: _Error: data.tree[0].children must contain at least 1 items_

**Wanted behaviour:**
the function move in call list should change the sorting from the call list. 
set as parent should make the via multiselect selected motions subitems from the in the list selected motion
insert after should move them behinf the selected motion (they should be on the same level)

luisa-beerboom:
Closes #2695

### DIFF
--- a/client/src/app/ui/modules/sorting/modules/sorting-tree/services/tree.service.ts
+++ b/client/src/app/ui/modules/sorting/modules/sorting-tree/services/tree.service.ts
@@ -177,6 +177,7 @@ export class TreeService {
 
     /**
      * Removes the `item`-property from any node in the given tree.
+     * Deletes empty children-arrays.
      *
      * @param tree The tree with items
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

Elblinator (approved):
function 

please do not merge and please do not assign someone
it will be merged in the near future

PR #2862, opened 2023-10-06
Update motion workflow export

### LINKED ISSUE #2555: Add the fields `set_workflow_timestamp` and `allow_motion_forwarding` to the motion workflow export
if a workflow is exported the two options set timestamp and allow forwarding of motions are not exported. 

These two options should also be exportetd

luisa-beerboom:
Close #2555

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/workflows/services/workflow-export.service/workflow-export.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/workflows/services/workflow-export.service/workflow-export.service.ts
@@ -28,7 +28,9 @@ export class WorkflowExportService {
             `show_state_extension_field`,
             `show_recommendation_extension_field`,
             `merge_amendment_into_final`,
-            `weight`
+            `weight`,
+            `set_workflow_timestamp`,
+            `allow_motion_forwarding`
         ];
         const json = [];
         const getNextWorkflowJson = (workflow: ViewMotionWorkflow) => {


bastianjoel (approved):

Elblinator (changes requested):
allow_motion_forwarding and set_workflow_timestamp are only exported if it was activated once 
It should always be exported even if it was never true

luisa-beerboom:
The reason why this happens is because all other workflow fields have defaults set in the backend. This causes the value to be automatically set in the backend upon workflow creation, meaning that those fields will have a value, even if they were never changed.

Therefore even the fields that are off by default technically have a value at the start, except for the two new fields which have no default set and will therefore initialized without a value (which will be interpreted as "false" in the client).
Since those two fields have no value, they will not be registered by the export.

There are multiple ways in which this may be changed.
1. I could build an extra conditional into the client and add the two fields, if they are empty, or
2. The backend could set defaults for these two fields (which may require a migration to fix this problem on legacy workflows)
3. The backend could delete the defaults of all boolean fields that are false by default (which would at least create consistency in that this behaviour would now happen to all these fields)

Given that the workflow.import and all workflow-related functionality will still work perfectly fine with how the system currently is, however, I question whether it is really necessary to change anything here.

What do you think would be best @Elblinator @rrenkert

rrenkert:
Please implement option 2.

Elblinator (approved):
Given the new course of action the client PR is working fine. Before merging however the back-end should be adjusted

see https://github.com/OpenSlides/openslides-backend/issues/1930

bastianjoel (approved):

jsangmeister:
Do we really want to add a migration for this? It feels like a huge overhead when the client could just fill in empty boolean fields with `false` by default, considering how long our migrations take, especially on big instances.

PR #2866, opened 2023-10-09
Mediafile support recursive folder download

### LINKED ISSUE #2586: Enhance download files function
**Description:**
If _download folder_ is selected not everything is downloaded.

**Current behaviour:**
1. If  _download folder_ is selected only files are downloaded and no folders and the items in the folder
2. If only a folder is selected in multi-select and downloaded, then an empty zip-file is downloaded

**Wanted behaviour:**
_download folder_ should download every available file and folder.
if a folder is specifically selected and then downloaded the folder and it's content should actually be downloaded

bastianjoel:
resolves #2586


Elblinator (approved):
function

Elblinator (approved):
first test approved

rrenkert (approved):

PR #2868, opened 2023-10-09
Fix icc connections not reopened on resolve

### LINKED ISSUE #2836: Client loses connection to ICC service after some time
When in a meeting, the client does not receive messages from the ICC service anymore. This can be tested, e.g., by opening a meeting in two windows with two different users and waiting some time (a few minutes?). Then, open the C4 dialog and search for players in both windows. Both windows will not receive the invitation and will continue searching.

Note that the ICC service is broken in the current demo alltogether, so this has to be tested locally or on nightly.

bastianjoel:
resolves #2836

### DIFF
--- a/client/src/app/gateways/base-icc-gateway.service.ts
+++ b/client/src/app/gateways/base-icc-gateway.service.ts
@@ -33,6 +33,10 @@ export abstract class BaseICCGatewayService<ICCResponseType> {
 
     private connectionClosingFn: (() => void) | undefined;
 
+    private get iccEndpointName(): string {
+        return ICC_ENDPOINT + `_` + this.serviceDescription;
+    }
+
     /**
      * Constructor to create the Service.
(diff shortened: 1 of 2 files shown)


luisa-beerboom (approved):

jsangmeister (changes requested):
Does not fix the problem for me. If I open two tabs in local dev and wait a few minutes, the connection is still lost.

bastianjoel:
I cannot replicate this with my local dev setup but I found another issue. 

When entering offline mode and then reconnection the notify service says that it reconnects but no notify connection is opened. Instead two connections to the applause endpoint are open. Maybe this is related to your issue. 

@jsangmeister  Could you verify if two connections to applause and no connection to the notify endpoint are open in your case?

luisa-beerboom (approved):

jsangmeister (approved):
Cannot reproduce my problem anymore

Elblinator (approved):
function

PR #2872, opened 2023-10-10
Show warning msg before try to delete operator

### LINKED ISSUE #2705: Meeting admin can try to delete himself
A meeting admin who has only one meeting can try to delete himself both via multiselect and via detail view.
This won't work as it will be blocked by the backend, but the error message is not shown since the action uses the handle_separately route and the client is therefore seemingly not equipped to recognize the error.

We should at least address the problem that the error isn't shown.

Also:
If the user is an Account-Admin he can remove himself from his own meeting, even if he is an admin there as well.
If he is the only admin left the backend will however refuse with a rather cryptic error message `data.group_ids[1] must be integer`

reiterl:
Closes #2705

### DIFF
--- a/client/src/app/gateways/presenter/get-user-related-models-presenter.service.ts
+++ b/client/src/app/gateways/presenter/get-user-related-models-presenter.service.ts
@@ -16,6 +16,7 @@ export interface GetUserRelatedModelsUser {
         motion_submitter_ids: Id[];
     }[];
     committees?: GetUserRelatedModelsCommittee[];
+    name?: string;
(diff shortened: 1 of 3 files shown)


bastianjoel (changes requested):

bastianjoel (approved):

Elblinator (changes requested):
There is still no error thrown if a meeting admin is deleting themselves
The client thinks that the action was successful
failed_succesfully

reiterl:
Okay, so this pr is not the solution.

PR #2873, opened 2023-10-10
Prevent add ballot if deleted users

### LINKED ISSUE #2683: Prevent error if an election with a deleted User is started
**Current behaviour:**
If a ballot with only a deleted User as candidate is created nothing happens visually and an error is thrown in the console: Error: Poll creation aborted because the minimum amount of votes was set higher than the number of available options

**Recreation:**
In a meeting:
1. navigate to elections > election
2. create a new user 
3. create, start and publish ballot
4. navigate to participants delete the user
5. navigate back to election 
6. in the candidate list should be only a deleted user
7. press the new ballot button
8. Error is thrown in the console

**Wanted behaviour:**
If only a deleted user is in the candidate list, then the "new ballot"-button should be disabled, and the deleted user (which already cannot be voted for) should be greyer than the rest

reiterl:
Closes #2683


Elblinator (approved):
function

Issue #2738, opened 2023-09-01
[META] Global search improvements and bugs


bastianjoel:
These are the remaining open points from #2616

Things which are important for short-term:
- [x]  The amounts displayed results should be under the filters and over the results (good first issue)
- [x] where the files are from is not indicated
- [x] max amount of results is 100; should there be a hard-cap for results found? (Might get better with https://github.com/OpenSlides/openslides-search-service/issues/52) @rrenkert 
- [x]  if you press on some results (tested with topics and motions) then the meeting id in the addressee-bar is _undefined_
- [x] pressing on the results is followed with long waiting periods (it is possible to prevent that with reloading the page
- [x] the search should be able to cover parts of words in titles (motions, elections affected) not only whole words (https://github.com/OpenSlides/openslides-search-service/pull/62)
- [x] special symbol _&_, Ä, Ö, Ü, ß etc. is not searchable (https://github.com/OpenSlides/openslides-search-service/pull/62)
- [x] Amendment changes should be searchable (requires search service change) (https://github.com/OpenSlides/openslides-search-service/pull/63)

Things which are important long-term:
- [x] colour of radio-buttons and check-box are not theme-coloured (good first issue)
- [x] you search from the orga-level, the filters are gone, the filters should be present even for the orga-level
- [x] the searched word is not bold (_good first issue_)
- [ ] participant should be searchable via their group
- [x] in the results "From" should be replaced with "by"
- [x] sequential number should only be displayed if it is displayed in the motions as well, if not possible then just remove it completely (good first issue)
- [x] out of place "<mark> [...] </mark>" in the crumb-menu if I search for meeting titles (try "Landesbezirkskonferenz") (_good first issue_)
- [x] Motion reason should be searchable
- [x] Linebreaks in search results broken

Other:
- [ ] If a the phone view is active (<=700px) then the filters should appear in a collapsible UI element. (good first issue)
- [x] Search field should be grey (see mock-up) (good first issue)
- [x] published ballots and votes and also votes from topics should be findable (_good first issue_)

MSoeb:
I've edited some priorities in the listing.

Elblinator:
marked "Search field should be grey (see mock-up) (good first issue)" as solved because this should be solved on a more general level and not specifically the global search

Elblinator:
If a deli is not allowed to see participants -and correctly cannot filter for them- then participants are still displayed in the results

bastianjoel:
[...]

PR #2879, opened 2023-10-11
Update global search to show marks bolder

reiterl:
Resolves "the searched word is not bold (good first issue)" of #2738

Global search result title are no longer `marked`.
`mark` tags are now bolder.


bastianjoel (approved):

Elblinator (dismissed):
Some of the titles are still not bold (Agenda, Motion, Election, File

bastianjoel:
The marks should not be visible in titles as far as I undestood.

Elblinator:
After reconsidering the desired Thanks for pointing that out 
I'll update the change request accordingly

Elblinator (changes requested):
If an account is found, then the part of the account which matched with the searched word is bold. 
The titles of the results should not be bold

reiterl:
Add code to prevent marking in account case.

bastianjoel (approved):

Elblinator (approved):
function

PR #2883, opened 2023-10-12
Add perm check to remove from motion block

### LINKED ISSUE #2877: Hide wrongly displayed actions
**Current behaviour:**
 on a small display: In a motion-block every user (independent of their group-rights) can see the action "Remove from motion block"

**Reproduction:**
in a meeting with a motion block with at least one motion inside: 
1. login as a deli (or another participant without the right to manage motions)
2. navigate to the motion-block
3. shrink your screen until you have the mobile-view
4. inside the three dot menu you see the action "Remove from motion block"

**Wanted behaviour:**
If a person does not have the rights to Remove motions from a motion block, then the action should NOT be displayed, even in the mobile-view

reiterl:
Closes #2877


bastianjoel (approved):

Elblinator (approved):
function

PR #2887, opened 2023-10-12
Fix conflicting amendment projection

### LINKED ISSUE #2880: Fix projection of motion
**Current behaviour:**
If a motion which has two amendments which change the same line is projected the projection is wrong

**Recreation:**
1. Create a motion with at least 4 paragraphs (each paragraph should have around 3 sentences)
2. create one amendment and change the first sentence
3. change the state of the amendment to zugelassen/accepted
4. repeat step 2 and 3 (make sure to change the same sentence)
5. project the motion as the diff version
6. the whole motion is displayed twice 

**Projected motion:**
projected_motion

**Current projector view (note that the motion in the second picture is scrolled down):**
projector

projector2

**Wanted projector view:**
projector_wanted

**Wanted changes written down:**
- the unchanged text should only be displayed once, and it should be displayed at the correct place (e.g. line 3 displayed in line 3)
- changed text should have a margin of 22 at the top and bottom

bastianjoel:
resolves #2880


luisa-beerboom (approved):

Elblinator (approved):
function

PR #2891, opened 2023-10-13
Possible fix for the projection edit dialog

### LINKED ISSUE #2878: Fix Projection edit-dialog
**Current behaviour:**
If some options were activated as projection defaults then they cannot be deactivated again

**Recreation:**
1. create a projector
2. open the edit-dialog from the projector
3. as projection defaults activate: Agenda, Ballots, Current list of speakers, List of speakers, Motion blocks, Motion votes, Countdowns, Messages and Topics
4. save
5. open the dialog again and try removing any of the defaults

**Wanted behaviour:**
It should be possible to remove every option as a projector default

reiterl:
Closes #2878 

There are two sets of keys used. (eg topic vs topics).
I tried to fix the problem with mixing them in the payload method.


bastianjoel (approved):

Elblinator (approved):
function

PR #2895, opened 2023-10-16
Update getTitle of speaker with deleted user

### LINKED ISSUE #2639: Account deletion: speech in LoS is deleted
**Description:** 
If a user does a speech during a meeting, it is recorded in the topic LoS. If the user account will be deleted, the reference in the LoS will also be deleted without a reference like anonymous user with XYZ speech time.

**Reproduction:**
1. open a meeting
2. open Agenda > create a new topic > project it to Autopilot
3. open participants > create a new user
4. open Autopilot > add new user to LoS > Start and Stop speech > now a reference of the speech is listed
5. open the organization level
6. open Accounts > delete new created user
7. head back to the meeting > Autopilot > reference is no more available

**What should happen:**
The speech reference should be available even after account deletion. The reference should be "deleted user + metadata of speech". Without that, the protocol of the meeting will lost important data. 

Another alternative for a reference will be to display "Anonymous user" instead of "deleted user".

reiterl:
Closes #2639

Needs pr https://github.com/OpenSlides/openslides-backend/pull/1921 to be closed

### DIFF
--- a/client/src/app/gateways/repositories/speakers/speaker-repository.service.ts
+++ b/client/src/app/gateways/repositories/speakers/speaker-repository.service.ts
@@ -20,7 +20,8 @@ export class SpeakerRepositoryService extends BaseMeetingRelatedRepository<ViewS
 
     public getVerboseName = (plural = false) => this.translate.instant(plural ? `Speakers` : `Speaker`);
 
-    public getTitle = (viewSpeaker: ViewSpeaker) => (viewSpeaker.user ? viewSpeaker.user.getFullName() : ``);
+    public getTitle = (viewSpeaker: ViewSpeaker) =>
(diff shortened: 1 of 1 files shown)


bastianjoel (commented):
I am not sure about that. It could be that the user is not available because it was not loaded. In that case this would be misleading. On the other hand I don't have a better idea on how to do that. 
@luisa-beerboom do you have a better idea?

luisa-beerboom:
> I am not sure about that. It could be that the user is not available because it was not loaded. In that case this would be misleading. On the other hand I don't have a better idea on how to do that. @luisa-beerboom do you have a better idea?

Just need to make sure he is loaded then

reiterl:
I looked at the `os-autopilot-main` subscriptions. Here the subscription seems to be in place.

Elblinator (approved):
function

PR #2901, opened 2023-10-17
Fix meeting admin update as committee admin

### LINKED ISSUE #2896: Meeting edit: Not possible to add yourself as Admin
**Description:** Currently it is not possible as a organization- or committee-admin to add yourself to a meeting as an admin via meeting edit view. Performing the action results in an error message "Fehler: You are not allowed to perform action meeting.update. Missing Permission: meeting.can_manage_settings". The message indicated, that this is a backend issue, but it is not. The payload sends wrong data, which leads to the error message. The payload sends a false request with empty date data.  

Screenshot_20231016_153747

**Reproduction:**
1. open an instance as superadmin
2. open accounts: create a new account
3. open committee: create a new committee, a meeting in it and add the new user as a committee admin to the created committee (NOT the meeting)
4. login as the new user
5. open committee: edit the meeting and try to add yourself to the meeting as admin

Hint: The search field to add a person as an admin didn't work well. Sometimes the enter field didn't load. This should be fixed in another issue. 

**What should happen:**
It should be possible that a committee admin could add itself as a meeting admin. 

But take care: Committee admins which did not have the permission to see all Accounts should only get the opportunity to add themselves to the meeting and nobody else. After they added themselves to a meeting, they could add more users as admins based on the participants of the meeting. 

Organization Admins instead could add themselves or somebody else as a meeting admin, because they can see all users.

bastianjoel:
resolves #2896

### DIFF
--- a/client/src/app/gateways/repositories/meeting-repository.service.ts
+++ b/client/src/app/gateways/repositories/meeting-repository.service.ts
@@ -137,6 +137,13 @@ export class MeetingRepositoryService extends BaseRepository<ViewMeeting, Meetin
             update.start_time = this.anyDateToUnix(update.start_time);
             update.end_time = this.anyDateToUnix(update.end_time);
         }
+        if (meeting && meeting.start_time === undefined && update.start_time === null) {
+            delete update.start_time;
+        }
+        if (meeting && meeting.end_time === undefined && update.end_time === null) {
+            delete update.end_time;
+        }
+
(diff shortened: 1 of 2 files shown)


luisa-beerboom (approved):

Elblinator (approved):
function

luisa-beerboom (approved):

PR #2909, opened 2023-10-18
PDF: Do not show referring motions when it is turned off in the motions settings

### LINKED ISSUE #2908: Motion PDF Export: referring motion exported even deactivated function
Description: If a user has set referrings between motions, the referrings are still exported even though the function to see referring motions is deactivated.

What should happen:
If the function 'show referring motions" is deactivated, the motion pdf export should exclude referrings.

m-schieder:
Fix #2908


MSoeb:
Small issue with the multiple export: If Show referring motions is disabled, the multi motion export still shows the possibility the select Referring Motions. If the user select it with the disabled function Show referring motions, the export correctly didn't include the referred motions. 

It would be good, if the multi selection menu could reflect the enabled/disabled function in the way, that the field Referring motions is enabled or disabled according to the function Show referring motions.

Elblinator (approved):

PR #2911, opened 2023-10-19
Add validators to email and default_vote_weight in some dialog

### LINKED ISSUE #2899: Disable Save button and show hint if email or default vote weight in <new account>/<new participant> is faulty
**Current behaviour:**
new account:
If the <new account> mask is opened and a wrong e-mail address or vote weight is put into the field the save button is not disabled and if the new account is saved the error  ```Cannot read properties of undefined (reading '0')``` is displayed but the mask with every input stays on screen
new participant:
If in step 1 or step 3 in the <new  participant> mask a wrong email-address is assigned to the new participant then the save button is not disabled and if pressed the error ```Error: email must be valid email.``` it thrown and the are navigated back to the participant-list (the input for the new participant is being lost)

**Correct input (as far as I understand it):**
Correct vote weight: number >= 0.000001

Correct Email: <word>@<word>.<word>
Description in words: 
A correct mail-address consists out of three strings with to separators:
- a string 
- "@"
- a string
- "."
- a string

**Wanted behaviour:**
If there is a wrong input in these masks then two things should happen:
1. the save button should be disabled
2. the input-field which has the faulty input should get coloured in the warn colour and a text should be displayed unter the line:
   - for a wrong email: "Please enter a valid email address. "
   - for a wrong vote weight: "Please select a vote weight bigger than or equal to zero."

reiterl:
Closes #2899

use the Angular email validator to validate the email.


bastianjoel (approved):

Elblinator (changes requested):
The save button is still not disabled if a wrong mail-address is entered (e.g. email@email) and the Error ```Cannot read properties of undefined (reading '0')``` is thrown on account-level and the following Error is thrown on participant-level ```Error: email must be valid email.```

reiterl:
@Elblinator  wants to look for a useful email validation in backend und frontend.

Elblinator:
@reiterl 
The mail-validation from the backend is sufficient and follows the current guide: https://www.icann.org/en/announcements/details/new-gtld-dotless-domain-names-prohibited-30-8-2013-en


https://github.com/OpenSlides/openslides-backend/blob/7bbbcc614b295c8fd44262c0e6e347ee2d8656bd/openslides_backend/action/mixins/send_email_mixin.py#L65

reiterl:
Two changes:

- use a custom email validator for emails with regex from the backend
- updated default vote weight validator to min of 0.000001

Elblinator (approved):
function

bastianjoel (approved):
While the implementation looks okay I would like to add that the backend (and now also frontend) regex does not allow everything that is possible. 

The regex used by the Angular validator can be found here with some explanation. 
https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address

Also leaving out the TLD might be useful when sending mails to local mailboxes. But I don't know if that is really a use case here.

PR #2919, opened 2023-10-19
Account import improvements

### LINKED ISSUE #2894: [META] Account Import issues
**Description:** The account import has some issues, which needs to be fixed:

A) The preview doesn't include the vote weight value. This only affects the preview, not the actual import. The data will be correctly imported. It is only not displayed in the preview. 

B) If the Account import file has values which leads to warnings in the import preview, the warning message needs to be improved to focus more on the fact what actually happens. - The text part "Accounts with warnings (will be skipped)" indicates a skipping of the whole account during import. This is simply not true. Only the coloumn is excluded.

C) If a warning happens, a warning sign is displayed in the preview. The warning sign is quite generic for the warning in the password coloumn if the data also includes a SAML value. - Good would be a warning, which reflects the case. Here, that it is not possible to provide a password with a SAML value. - If this could be improved, it would be good. 

D) In the import description is a small ?-sign which could be clicked to get more information. This info-box should be improved for gender options. So the user can easily see, what could be used for import. 

**Reproduction:**

Case A) 
1. open the organiszation level
2. open Accounts
3. Import a user with a vote weight and check the preview of the data

Case B and C
1. Do an import of user data with username AND SAML until preview appears. In the preview box the text part "Accounts with warnings (will be skipped)" should appear. Also a warning sign should be visible in the username coloumn. Hover over it to see the warning message.

Case D)
1. Open Account > Import. Description is visible above.

**What should happen:**
- [x] A) The vote weight values should be included in the preview. 
- [x] B) Change text part to Accounts with warnings (Coloumns will be skipped)
- [x] C) Check of the warning message could be improved for this case or if it's a generic one for all coloumns. If it could be improved change it to "Import with SAML doesn't allow password."
- [x] D) Add all possible gender options to the help-box. This includes: female, male, diverse, non-binary

luisa-beerboom:
Closes #2894

### DIFF
--- a/client/src/app/site/pages/organization/pages/accounts/pages/account-import/services/account-import.service/account-import.service.ts
+++ b/client/src/app/site/pages/organization/pages/accounts/pages/account-import/services/account-import.service/account-import.service.ts
@@ -31,7 +31,7 @@ export class AccountImportService extends BaseBackendImportService {
         created: _(`Accounts created`),
         updated: _(`Accounts updated`),
         error: _(`Accounts with errors`),
-        warning: _(`Accounts with warnings (will be skipped)`)
+        warning: _(`Accounts with warnings (columns will be skipped)`)
     };
 
     public constructor(
--- a/client/src/app/site/pages/organization/pages/accounts/pages/account-import/services/account-import.service/account-import.service.ts
+++ b/client/src/app/site/pages/organization/pages/accounts/pages/account-import/services/account-import.service/account-import.service.ts
@@ -31,7 +31,7 @@ export class AccountImportService extends BaseBackendImportService {
         created: _(`Accounts created`),
         updated: _(`Accounts updated`),
         error: _(`Accounts with errors`),
-        warning: _(`Accounts with warnings (affected cells will be skipped)`)
+        warning: _(`Accounts with warnings: affected cells will be skipped`)
     };
 
     public constructor(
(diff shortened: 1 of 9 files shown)


bastianjoel (approved):

Elblinator (changes requested):
A) vote weight as short as possible
B) Please change textt to "Accounts with warnings: the affected cells will be skipped: 
D)"
 1) one space to much in helper-box
 2) Please change in the preview "Gender, possible options: male, female, diverse, non-binary" to only "Gender"

client/src/app/site/pages/organization/pages/accounts/pages/account-import/services/account-import.service/account-import.service.ts
```diff
@@ -31,7 +31,7 @@ export class AccountImportService extends BaseBackendImportService {
         created: _(`Accounts created`),
         updated: _(`Accounts updated`),
         error: _(`Accounts with errors`),
-        warning: _(`Accounts with warnings (will be skipped)`)
+        warning: _(`Accounts with warnings (affected cells will be skipped)`)
```
Elblinator, line 34: warning: _(`Accounts with warnings: affected cells will be skipped`)

Elblinator (changes requested):

bastianjoel (approved):

Elblinator (approved):

PR #2947, opened 2023-10-25
Update sort and filter at meeting switch

### LINKED ISSUE #2870: Motions: Filter stays active after meeting switch
**Description:** Selected filters in motions stays active after switching to another meeting. The selection seems to be set to  global default value over all accessible meetings. 

**Reproduction:**
Precondition: At least two meetings with the same user base.
1. open meeting A > motions > set a filter
2. head back to organisation level and open the other meeting B > motions. The selected filter is still active.

**What should happen:**
The selected filter should only be active inside the meeting where it is set and not global in all other meetings.

reiterl:
Closes #2870 

Small cleanup

### DIFF
--- a/client/src/app/site/base/base-sort.service/base-sort-list.service.ts
+++ b/client/src/app/site/base/base-sort.service/base-sort-list.service.ts
@@ -1,4 +1,4 @@
-import { Directive, Injector, ProviderToken } from '@angular/core';
+import { Directive, inject, Injector, ProviderToken } from '@angular/core';
 import { TranslateService } from '@ngx-translate/core';
 import {
     auditTime,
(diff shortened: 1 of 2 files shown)


bastianjoel (approved):

Elblinator (approved):
function

client/src/app/site/base/base-sort.service/base-sort-list.service.ts
```diff
@@ -177,6 +179,8 @@ export abstract class BaseSortListService<V extends BaseViewModel>
 
     private initializationCount = 0;
 
+    private activeMeetingIdService = inject(ActiveMeetingIdService);
```
jsangmeister, line 182: Shouldn't this be injected via the constructor like all other services?

reiterl: I had problems with the use of constructor injection here, as I need to update  5+ super calls then. I talked to @bastianjoel  about this and he said I can use this method here.

bastianjoel: `inject` was introduced in Angular 14. I assume that this was not done before because the feature was not available when these classes were written. 
In my opinion, placing service injections in the constructor within base classes (or any other classes that will be extended) makes these classes unnecessarily harder to maintain as you always need to update all references when adding or removing a service the class is using.

jsangmeister: I agree! I just asked for consistency reasons. I'm all for switching all classes to the new injector style, if possible - the long constructor chains have always bugged me immensly. Is there maybe an automatic migration tool to get rid of these?

PR #2949, opened 2023-10-25
Change Arrow behaviour in motion-detail

### LINKED ISSUE #2424: Change of arrows in motion-view
Where: in the motion-view while being in the detail-view of one motion 

In the desktop view two arrows can be used directly to switch between the different motions: 
motion-arrows

in the tablet and phone view these arrows move into the three-dot-menu:
motion-arrows-small

Wanted behaviour: 
The arrows should not move to the three dot menu, regardless of the display size
In the tablet-view the arrows should be like the arrows in the desktop-view
In the phone view the arrows should stay where they are but lose the describer (in the pictures 15-Ä01 and R01-Ä02)

Elblinator:
Closes #2424

I took the liberty to change the appearance in a different way. 
Could the testers please state if the new behaviour is still inline with the wanted behaviour (to make the handling on the phone easier)


bastianjoel (approved):

MSoeb:
In general the buttons works as they should. But there is small layout issue with small mobile devices in portrait mode. The buttons are quite large in portrait mode. @Elblinator let's check this. Maybe we can optimize this.

bastianjoel (approved):

Elblinator:
verbal rewiev from @MSoeb 
Tablet-view is fine.
In the phone view: only the arrow should be displayed

bastianjoel (approved):

MSoeb (changes requested):
image

First: It works as it should. Small improvements needed. 

- [ ] The distance/space between the navigation buttons in mobile view is a to wide. They should be placed a bit nearer. 
- [ ] The desktop view has a slightly different navigation layout, which made the describer "jump". This feels strange. Does this happens because of the wider space on mobile view? (@Elblinator I will show it to you)

bastianjoel (approved):

bastianjoel (approved):

MSoeb (approved):
Works as intended.

MSoeb (approved):
Works as intended.

PR #2953, opened 2023-10-26
Add missing sorting for filter properies in filter list

### LINKED ISSUE #2939: Motion Block: Titles with spaces leads to wrong filter sorting
**Description:** If a motion block has a title with one or more spaces, the motion block will be listed in the motion filters on the highest position.

**Example: Motion block title "   Z" -->**
Screenshot_20231023_152254

**Reproduction:**
1. open a meeting
2. open motions > three-dot-menu > motion blocks: Create a block with the title "Test" and another one with the title "   Z". Take care of spaces. The listing in the motion block menu should be fine.
3. head back to motion menu > click on filters and check the order of the motion blocks. wrong order of blocks. 

**What should happen:** 
To avoid this behavior, spaces at the beginning of a motion block title should not be allowed.

reiterl:
Closes #2939

Sort the filterProperties by the label which is trimed

### DIFF
--- a/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
+++ b/client/src/app/site/base/base-filter.service/base-filter-list.service.ts
@@ -277,9 +277,6 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements 
             .pipe(
                 map(viewModels => {
                     if (viewModels && viewModels.length) {
-                        const sortFn = (a, b) => {
-                            return a.label.trim().localeCompare(b.label.trim());
-                        };
                         const filterProperties: (OsFilterOption | string)[] = viewModels
                             .filter(filterFn ?? (() => true))
                             .map((model: any) => ({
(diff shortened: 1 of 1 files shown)


client/src/app/site/base/base-filter.service/base-filter-list.service.ts
```diff
@@ -293,7 +296,8 @@ export abstract class BaseFilterListService<V extends BaseViewModel> implements
                                               condition: child.id
                                           }))
                                         : undefined
-                            }));
+                            }))
+                            .sort(sortFn);
```
bastianjoel, line 300: No need to declare the sort function somewhere else. 

```suggestion
                            .sort((a, b) => a.label.trim().localeCompare(b.label.trim()));
```

bastianjoel (changes requested):

bastianjoel (approved):

Elblinator (approved):
function

PR #2954, opened 2023-10-26
Use gramatically correct email snackbar messages

luisa-beerboom:


peb-adr:
Is this enough for translations to also keep working correctly?
Won't they break now as the string will no longer be matched?

luisa-beerboom:
@peb-adr  The translations will be updated by someone else later

Elblinator (approved):

PR #2960, opened 2023-10-27
Add tests to openslides-status service

reiterl:
I'm not sure, that the reset method is correct. It doesn't change the isStable to false.

### DIFF
--- a/client/src/app/site/services/openslides-status.service.spec.ts
+++ b/client/src/app/site/services/openslides-status.service.spec.ts
@@ -1,8 +1,9 @@
 import { TestBed } from '@angular/core/testing';
+import { Deferred } from 'src/app/infrastructure/utils/promises';
 
 import { OpenSlidesStatusService } from './openslides-status.service';
 
-xdescribe(`OpenSlidesStatusService`, () => {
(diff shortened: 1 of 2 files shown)


client/src/app/site/services/openslides-status.service.spec.ts
```diff
@@ -13,4 +14,20 @@ xdescribe(`OpenSlidesStatusService`, () => {
...
+    it(`observe isStable`, () => {
+        let actualValue: boolean | undefined;
+        service.isStableObservable.subscribe(value => {
+            actualValue = value;
+        });
+        expect(actualValue).toBe(false);
+        service.setStable();
+        expect(actualValue).toBe(true);
+        service.reset();
+        expect(actualValue).toBe(true);
```
bastianjoel, line 31: This seems odd.

reiterl:
The reset method is not used, so I cut it and updated the tests.

client/src/app/site/services/openslides-status.service.spec.ts
reiterl: I removed the reset method.

PR #2963, opened 2023-10-30
Fix a problem in the poll service

### LINKED ISSUE #2914: Fix per default selected Groups in polls
**Current behaviour:**
If a poll is started and something else than the default group/s is selected for this poll, then these groups are remembered the next time you start a poll, in contrast to the actual default group/s.

**Reproduction:**
1. in a meeting: 
2. navigate to settings > Elections
3. select Admins for "Default groups with voting rights"
4. Navigate to Election-list > one election 
5. open the dialog of "new ballot" and select delegates and staff as "Entitled to vote"
6. create the ballot 
7. start creating a new Ballot
8. Now the groups delegates and staff are preselected in "Entitled to vote"

(The votes in motion and Topics have the same behaviour)

**Wanted behaviour:**
If a new Ballot/Vote is being created the groups which are selected as the default group in the settings should be preselected in "Entitled to vote"

reiterl:
Closes #2914 

An array is an reference type and can be changed. So copy the array helps.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/agenda/modules/topics/modules/topic-poll/services/topic-poll.service.ts
+++ b/client/src/app/site/pages/meetings/pages/agenda/modules/topics/modules/topic-poll/services/topic-poll.service.ts
@@ -75,7 +75,7 @@ export class TopicPollService extends PollService {
         const poll: Partial<Poll> = {
             title: this.translate.instant(`Vote`),
             onehundred_percent_base: this.defaultPercentBase,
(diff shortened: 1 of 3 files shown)


Elblinator (changes requested):
Please also add default groups for motion-votes and agenda-votes

reiterl:
Add the fix to the two other poll services.

Elblinator (approved):
function

bastianjoel (approved):

PR #2979, opened 2023-11-06
Move motion block recommendation text after icon

### LINKED ISSUE #2930: Motion blocks: Follow recommendations for all motions text is gone
Description: If a user has opened a motion block and want to follow all recommendations, the function is hiding behind an icon with two hooks. For unexperienced users or users of OS3 this is an unclear or strange behavior.

OS4:
Screenshot_20231020_142619

OS3:
Screenshot_20231020_142711

Reproduction:
1. open a meeting
2. open motions > create a motion 
3. open settings > motions > add a recommender
4. head back to motions and add the recommender to the motion
5. open the motion block menu and create a motion block. 
6. add the motion to the block and open the block

What should happen:
Like in OS3 beside the icon with two hooks should be the text "Follow recommendations for all motions" visible. Also the text and the icon should be one button.

reiterl:
Closes #2930


Elblinator (approved):
The tooltip is now gone, which is okay, given that the icon is never shown without the text now

PR #2983, opened 2023-11-06
Add sorting to workflow next states (comma-list, menu)

### LINKED ISSUE #2929: Workflow: Sorting changes in workflow doesn't affect drop-down-menu in next state
**Description:** If a user sorts the the states in workflows, the sorting doesn't affect the drop down menu in next state. 

**Original sorting**
Screenshot_20231020_135623

**last states changed**
Screenshot_20231020_135653

**menu without correct sorting**
Screenshot_20231020_135743

**Reproduction:**
1. open a meeting
2. open motions > three-dot-menu > workflows > simple workflow > open three-dot-menu > sort > change a state position
3. head back to simple workflow > click on a next state menu and check the order

Important: Take care, some states didn't all other states to be a follow up. In this case they are correctly not displayed. 

**What should happen:**
The new sorting should be inlcuded at the correct position in the next state menu, if the option is available as the next state. 

Note: Not 100% sure, if this a bug or an enhancement. But I prefer the bug interpretation because of the inconsistency.

reiterl:
Closes #2929

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/workflows/components/workflow-detail/workflow-detail.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/workflows/components/workflow-detail/workflow-detail.component.html
@@ -106,7 +106,7 @@
                             <div class="inner-table">
                                 <div *ngIf="!state.next_states.length">-</div>
                                 <div *ngIf="state.next_states.length">
-                                    <os-comma-separated-listing [list]="state.next_states">
+                                    <os-comma-separated-listing [list]="sortedNextStates(state)">
                                         <ng-template let-next>{{ next.name }}</ng-template>
                                     </os-comma-separated-listing>
                                 </div>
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
function

client/src/app/site/pages/meetings/pages/motions/pages/workflows/components/workflow-detail/workflow-detail.component.html
```diff
@@ -106,7 +106,7 @@ <h2 *ngIf="workflow">
                             <div class="inner-table">
                                 <div *ngIf="!state.next_states.length">-</div>
                                 <div *ngIf="state.next_states.length">
-                                    <os-comma-separated-listing [list]="state.next_states">
+                                    <os-comma-separated-listing [list]="sortedNextStates(state)">
```
bastianjoel, line 109: This can cause performance issues because this gets calculated on every change detection cycle and the underlying method performs a sort. 

However because the next states might be quite small in most cases I approve.

bastianjoel (approved):

PR #2994, opened 2023-11-08
Make user option title load

luisa-beerboom:
Closes #2984

PR #2995, opened 2023-11-09
Make user option title load - alternative

luisa-beerboom:
I tried another solution based on #2994  so we can test if this works in production

@m-schieder can you create a test instance for this?


rrenkert (approved):
Tested.

bastianjoel (approved):

PR #3010, opened 2023-11-13
Allow more refined Re-add button behaviour

### LINKED ISSUE #2991: Fix re-Add last speaker
**Current behaviour:**
If someone is on the LoS with a normal contribution AND a point of order and then one of them is finished, then neither the normal contribution nor the point of order can be re-added -> because the "User is already on the list of speakers"

**Reproduction:**
in a meeting:
1. add yourself as a participant
2. navigate onto a LoS 
3. add a point of order and a normal contribution
4. start and stop the point of order 
5. the re-add button will be disabled
Addition:
6. start the normal contribution
7. the re-add button is not disabled anymore 
8. if activated the following Error will be thrown ```Error: User ## is already on the list of speakers```

**Wanted behaviour:**
(these descriptions are if the speaker is the same person
A normal contribution should not prevent a point of order to be re-added
A point of order should not prevent a normal contribution to be re-added
If a point of order/contribution is started then the re-add button should (correctly) be disabled

Elblinator:
closes #2991 

The back-end is still throwing errors.
Before merging this PR the backend has to allow the readding


rrenkert:
Backend issue is closed and the PR is merged. Please test.

MSoeb (approved):
Works as it should.

PR #3012, opened 2023-11-14
Synchronize commitUpdate

### LINKED ISSUE #3004: Motion text not loading when opening multiple tabs
Sometimes the motion text is not loading when opening multiple motions quickly after another. 

It seems like the motion detail view gets a broken motion observable with incomplete/broken motion data.

bastianjoel:
resolves #3004


Elblinator (approved):
function

luisa-beerboom (approved):

PR #3013, opened 2023-11-15
Global search improvements

bastianjoel:
- Display results found via change reco or poll
- Fix mediafile folder urls
- Display preview and mark match in amendments
- Display the motions reason if a result was matched by it


luisa-beerboom (approved):

Elblinator (changes requested):
Folders are not detectable: {"message":"Media-Server: Not Found."}

Amendments are not detectable at all.

bastianjoel:
Sounds like you are using an old version of the search service.

Elblinator (approved):

PR #3021, opened 2023-11-17
Add tests for view port service

reiterl:
Just two small unit tests.


bastianjoel (changes requested):
I don't think this is very useful. Maybe use something like https://www.npmjs.com/package/karma-viewport to actually test if this reacts to viewport changes. 

While you are at it you could also change `isMobileSubject` to an `Observable<boolean>` instead of a `BehaviorSubject` so that it is not changeable from outside the service.

PR #3034, opened 2023-11-21
Fix vote weight 0 not displayed

### LINKED ISSUE #3033: Vote_weight: Fix wrong display and export
**Current behaviour:**
If a participant was saved with vote_weight: 0 in the past (which was allowed in the past, but not anymore) then the vote weight is displayed as as "vote weight: 1"

**Wanted behaviour:**
the vote weight should be correctly displayed

bastianjoel:
resolves #3033

### DIFF
--- a/client/src/app/site/pages/meetings/view-models/view-user.ts
+++ b/client/src/app/site/pages/meetings/view-models/view-user.ts
@@ -198,8 +198,7 @@ export class ViewUser extends BaseViewModel<User> /* implements Searchable */ {
     }
 
     public vote_weight(meetingId?: Id): number {
-        const mUserVoteWeight = this.getMeetingUser(meetingId)?.vote_weight;
-        return mUserVoteWeight ?? this.default_vote_weight;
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/view-models/view-user.ts
```diff
@@ -198,7 +198,8 @@ export class ViewUser extends BaseViewModel<User> /* implements Searchable */ {
     }
 
     public vote_weight(meetingId?: Id): number {
-        return this.getMeetingUser(meetingId)?.vote_weight || this.default_vote_weight;
+        const mUserVoteWeight = this.getMeetingUser(meetingId)?.vote_weight;
+        return mUserVoteWeight || mUserVoteWeight === 0 ? mUserVoteWeight : this.default_vote_weight;
```
luisa-beerboom, line 202:
```suggestion
        return mUserVoteWeight ?? this.default_vote_weight;
```

Wouldn't this work too?

bastianjoel: Yes

luisa-beerboom (approved):

PR #3036, opened 2023-11-21
Fix multiselect motion move in agenda

### LINKED ISSUE #3019: Motion-list > multi-select: Prevent Error in snackbar
**Current behaviour:**
If a several motion are added to the agenda and at least one motion is already in the agenda then a cryptic error is thrown: ```Error: Update of agenda_item/3395: You try to set following required fields to an empty value: ['content_object_id']```

**Reproduction:**
in a meeting:
1. have at least one topic in agenda
2. create two motions, put one of them in the agenda
3. navigate to motion-list 
4. activate multi-select and select the two motions
5. move them into the agenda (select the topic as parent)
6. error is thrown

**Wanted behaviour:**
If one of the motions is already in the agenda, then this motion should be skipped and the rest should be continued

bastianjoel:
resolves #3019


Elblinator (changes requested):
if a motion is added via multiselect then the visibility is always hidden. -> the visibility should follow the pre determined default visibility (see settings > default visibility for new agenda items)

bastianjoel:
Not moving the motion when it is already in the agenda would change the current behavior. I think this should be discussed and has nothing to do with this fix so please open another issue for that.

Elblinator (approved):
function

luisa-beerboom (approved):

PR #3042, opened 2023-11-23
Implement motion import via backend

### LINKED ISSUE #2871: [Imports] Implement motion import via backend
See OpenSlides/openslides-backend#1713

The motion export should be extended to allow for exports in correct backend `motion.import` format.
We want to keep the old format as well for the sake of customers who were using it for some purpose. Especially for the users, the proposal was to keep the old functionality and add username columns.

There was also a discussion whether OpenSlides should recognize if the user tries to import a previously exported amendment and remove it from the payload pre-emptively. For that purpose we want to export if the motion is an amendment (see description in Wiki) as a new boolean column `motion_amendment` in the export.

luisa-beerboom:
Closes #2871 

needs OpenSlides/openslides-backend#1923


bastianjoel (approved):

Elblinator (changes requested):
category_name and category_prefix are not exported (missing in the CSV-file)
-> if they are correctly in the CSV-file then the category_name is correctly displayed in the preview and the category_prefix has only the blue mat-icon block icon. 
If the prefix should also be displayed as the letter then that's currently still missing

UI: 
unknown  submitter/supporter: 
1. the warning icon/autorenew icon are not aligned with the names, both should be aligned in the center
2. the icons and the names which belong together are two or three spaces away from each other (the space should be smaller) and there is no space in between the icon and the name which do not belong together (the space should be bigger)

If tags are unknown then the name and the icon are also not aligned

luisa-beerboom:
> category_name and category_prefix are not exported (missing in the CSV-file) -> if they are correctly in the CSV-file then the category_name is correctly displayed in the preview and the category_prefix has only the blue mat-icon block icon. If the prefix should also be displayed as the letter then that's currently still missing

Clear cache and try exporting again, are they still not exported?

luisa-beerboom:
> -> if they are correctly in the CSV-file then the category_name is correctly displayed in the preview and the category_prefix has only the blue mat-icon block icon.
> If the prefix should also be displayed as the letter then that's currently still missing

That was due to a backend bug. It should be fixed with the current version

Elblinator (changes requested):
1. please add a space between the warn-icon and the user to the right of the icon 
see warn-icon and "admin"
abc

2. the warn-icon behind the category prefix is missing 
 is that also in the backend bug you mentioned above?


additional question:
Now in the export: you can select submitter_verbose, submitter_name, supporter_name and supporter_verbose.  Was the change from only Submitter/Supporter to _name and _verbose on purpose?

luisa-beerboom:
> 2\. the warn-icon behind the category prefix is missing
>         is that also in the backend bug you mentioned above?

There's not supposed to be any warn icon there.
The `category_prefix` field is an extension for the `category_name` field. It will show no errors as all errors related to it are to be attributed to the main field, which is `category_name`

luisa-beerboom:
> additional question:
> Now in the export: you can select submitter_verbose, submitter_name, supporter_name and supporter_verbose. Was the change from only Submitter/Supporter to _name and _verbose on purpose?

Yes, that was on purpose:
The export previously only worked with full names (i.e. in the worst case smth along the lines of "Prof. Dr. Max Mustermann (er/ihm · NRW · Nr. 123456)")
That is pretty much impossible for the backend to interpret, but  might still be necessary to export for people who export for different reasons than for importing, so we kept that around in the verbose fields.
The best choice for the backend was a simple list of usernames, so we added the ability to export that.
For the sake of avoiding mistaken exports due to misunderstood features we have allowed the verbose fields in the import and, if filled, the backend loosely compares the length of the verbose to that of the username fields, complaining when it looks like the user may have added users in the wrong column

Elblinator (approved):
function

jsangmeister:
ping @bastianjoel 

and @luisa-beerboom probably needs to rebase

Elblinator (changes requested):
There is still a UI Bug left
All titles from the motions should be centered
 (this example motion import was created with the initial data from the instance)

bastianjoel (approved):

Elblinator (approved):
function 

But @MSoeb (or @emanuelschuetze) should look over the functionality as well

bastianjoel:
@MSoeb @emanuelschuetze What is the status here?

bastianjoel:
Please retest after merge. Changes were not trivial.

Elblinator (changes requested):
I would like an answer from @Msoeb or @emanuelschuetze of that needs to be discussed here or what they want differently before any of my things get fixed. 

Found problems
- If I try to export the motion example file the following Error is thrown in the console: 
```
backend-import-list.component.html:113 ERROR 
TypeError: headers.join is not a function
    at MeetingCsvExportForBackendService.dummyCSVExport (csv-export-for-backend.service.ts:107:36)
```

This following error is not specific for this PR but I want to document it here anyway: 
- Open the motion export dialog the following Error is thrown in the console:
```
Error: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'null'. Current value: 'false'. Expression location: MotionExportDialogComponent component. Find more at https://angular.io/errors/NG0100
    at throwErrorIfNoChangesMode (core.mjs:9931:11)
```

MSoeb (changes requested):
To finalize the PR, content aspects must first be clarified. This will not take place until mid-August. Until then, work on the PR can be paused.

MSoeb:
Work on this PR can be stopped. The motion import will not be developed further and no bugs will be fixed.

Elblinator:
https://github.com/OpenSlides/openslides-client/pull/4377

Issue #3055, opened 2023-11-29
Refactor `http.service.ts` error handling


bastianjoel:
Currently the `http.service.ts` tries to handle certain backend errors by itself and returns `null` if it does so. 

This is bad because the caller will not be notified properly that a request failed and lacks information about the reasons for that. In case of a password reset which returns `null` on success it is completely impossible for the caller to determine if the request was successful. 

While I think that it is okay to display a snackbar on fail we should change the following things here:
1. Always throw if an error occurred with all information about the response
2. Option to disable opening a snackbar on error

Especially 1. needs testing and maybe some considerations.

jsangmeister:
From my perspective option 2) is sufficently implemented

bastianjoel:
I still think not throwing an error should not be the default. 

> From my perspective option 2) is sufficently implemented
I guess you mean 1)?

bastianjoel:
I will change this to cleanup because the main part is now at least doable.

PR #3056, opened 2023-11-29
Display password change errors

### LINKED ISSUE #3054: Show error message if password change failed
**Describe the bug**

If the user attempts to change her/his own password, a success message displayed, but the password is not changed.

**How to Reproduce**

1. Install OpenSlides 4.0.16.
2. Log in and password that was set up during installation
3. Open user menu > “My account” > “Change password”
4. Type password that was used to login into “Old password” field
5. Type a different password into “New password” and “Confirm new password” fields
6. Click “Save”
7. A message “Password changed successfully!” is displayed.
8. Log out and attempt to log in again.

**Observed behavior**

- Login with the password set up during installation is possible.
- Login with the changed password is rejected.

**Expected behavior**

- Login with the password set up during installation is rejected.
- Login with the changed password is possible.

**System information**

- OpenSlides version: 4.0.16
- OS: Debian Linux
- Browser: Firefox
- Browser version: 120.0

**Additional context**

- It seems that the message “Password changed successfully!” is always displayed, even if the original password is wrong.
- The following messages are logged in the backendAction container when changing the password:
```
INFO:openslides_backend.shared.handlers.base_handler:Performing action user.set_password_self.
ERROR:openslides_backend.shared.handlers.base_handler:Error occured on index 0: Wrong password
```

bastianjoel:
resolves #3054 
This is a very bad workaround. 

To resolve this properly we have to do some deeper changes. I opened #3055 for that.

### DIFF
--- a/client/src/app/site/modules/global-headbar/components/account-dialog/account-dialog.component.ts
+++ b/client/src/app/site/modules/global-headbar/components/account-dialog/account-dialog.component.ts
@@ -150,9 +150,22 @@ export class AccountDialogComponent extends BaseUiComponent implements OnInit {
 
     public async changePassword(): Promise<void> {
         const { oldPassword, newPassword }: PasswordForm = this.userPasswordForm;
+
+        // TODO: Part of workaround
+        if (this.snackbar._openedSnackBarRef) {
+            this.snackbar._openedSnackBarRef.dismiss();
+        }
+
         this.repo.setPasswordSelf(this.self!, oldPassword, newPassword).then(() => {
-            this.snackbar.open(this.translate.instant(`Password changed successfully!`), `Ok`);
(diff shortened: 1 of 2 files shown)


luisa-beerboom (changes requested):
If you go into `error-map-utils.ts` and, in the getActionErrorMap function, do for `UserAction.FORGET_PASSWORD_CONFIRM` and `UserAction.SET_PASSWORD_SELF` exactly what has already been done for `MotionAction.CREATE_FORWARDED`, the HttpService should throw an error instead of opening a snackbar.
You can then wrap the repository function call in a try/catch block and open whatever snackbar is appropriate

luisa-beerboom (approved):

Elblinator (approved):

PR #3061, opened 2023-11-30
Remove translation marker

### LINKED ISSUE #3052: XLSX exports should not translate content fields
The xlsx export currently translates content fields. It shouldn't.

Reproduction:
1. Create a motion with a unique title
2. In the meeting settings enter a custom translation for the same motions title
3. XLSX export on the motion list -> The title field contains the custom translation for the title string instead of the actual title

Elblinator:
closes #3052


MSoeb (approved):
Fix does what it should.

bastianjoel (approved):

PR #3072, opened 2023-12-06
Fix autoupdate handleError error type check

### LINKED ISSUE #3071: Re-Login problem after logout
**Description:** If a user is in a meeting and logs out of it, an error occurs when logging in again directly. The login page displays a permanent spinning wheel. Logging in is only possible after reloading the page. When reloading, the user is logged in directly. 

**Reproduction:**
1. open a meeting
2. log out
3. log in
4. -> spinning wheel appears.

**What should happen:**
Logging should be possible without a page reload and without a permanent spinning wheel.

bastianjoel:
Client errors are currently not correctly handled by the shared worker. This could result in unexpected behavior when the autoupdate reports an client error. 

resolves #3071

### DIFF
--- a/client/src/app/gateways/http-stream/http-stream.ts
+++ b/client/src/app/gateways/http-stream/http-stream.ts
@@ -328,7 +328,6 @@ export class HttpStream<T> {
     }
 
     private async handleStreamError(error: unknown): Promise<void> {
-        console.log(error);
         if (error instanceof HttpErrorResponse) {
             try {
                 error = this.parseCommunicationError(error.error as any, error);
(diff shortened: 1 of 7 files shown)


luisa-beerboom (approved):

bastianjoel:
Test are failing now because of https://github.com/OpenSlides/openslides-autoupdate-service/issues/789

client/src/app/gateways/http-stream/http-stream.ts
```diff
@@ -327,10 +328,10 @@ export class HttpStream<T> {
     }
 
     private async handleStreamError(error: unknown): Promise<void> {
+        console.log(error);
```
luisa-beerboom, line 331: delete

Elblinator (approved):

PR #3074, opened 2023-12-07
Work on the speaking times component in autopilot

reiterl:
Remove the speakers tab and add some logic to the speaking times part.


bastianjoel:
Will be continued in #3045

PR #3094, opened 2023-12-12
Fix import error handling and cleanup participant import icons

luisa-beerboom:
In this PR I did the following:

1. Previously when the import failed with a non-preview-related error (i.e. a string error that doesn't update the preview with an explanation), a snackbar with the error would be shown, but the view would continue showing the preview in a context that implied the import to still be ongoing. -> Fixed that by making the view revert to a clean pre-import state
2. The participant import has a system where trying to change fields that the user has no right to edit will get the field payload removed. This is signified in the preview through a new field state `remove`. Until now that state had no associated icon, causing such fields to show the default icon `block`, which suggests that there is an error on that field, when really there isn't -> Made OpenSlides use the `remove` icon for fields with the `remove` state.

### DIFF
--- a/client/src/app/gateways/error-mapping/error-map-utils.ts
+++ b/client/src/app/gateways/error-mapping/error-map-utils.ts
@@ -54,6 +54,9 @@ const MeetingCreateErrorMap: ErrorMap = new ErrorMap([
  */
 const getActionErrorMap: (data: any) => ErrorMap | null = data => {
     const actionName = Array.isArray(data) && typeof data[0] === `object` ? data[0][`action`] : null;
+    if (typeof actionName === `string` && actionName.endsWith(`.import`)) {
+        return MatchAllErrorMap;
+    }
     switch (actionName) {
         case MeetingAction.CREATE:
             return MeetingCreateErrorMap;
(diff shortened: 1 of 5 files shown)


client/src/app/gateways/error-mapping/error-map-utils.ts
```diff
@@ -54,6 +54,9 @@ const MeetingCreateErrorMap: ErrorMap = new ErrorMap([
  */
 const getActionErrorMap: (data: any) => ErrorMap | null = data => {
     const actionName = Array.isArray(data) && typeof data[0] === `object` ? data[0][`action`] : null;
+    if (typeof actionName === `string` && actionName.endsWith(`.import`)) {
+        return MatchAllErrorMap;
+    }
```
bastianjoel, line 59: Move this into the default case.

bastianjoel (changes requested):

client/src/app/gateways/error-mapping/error-map-utils.ts
luisa-beerboom: Done

Elblinator (changes requested):
after running into the fix for 1 once you need to reload the page to import the same file again

Elblinator (approved):
function
Change request from before will become it's own issue

PR #3102, opened 2023-12-13
Imports: Make sure file can be re-added after import fails with error

### LINKED ISSUE #3100: Fix import-file handling after one error in account import
The change request from #3094 is moved into it's own issue.
> Previously when the import failed with a non-preview-related error (i.e. a string error that doesn't update the preview with an  explanation), a snackbar with the error would be shown, but the view would continue showing the preview in a context that implied the import to still be ongoing. -> Fixed that by making the view revert to a clean pre-import state

Change request:
> After running into the fix for 1 once, you need to reload the page to import the same file again in chromium

**Reproduction:**
chromium specific:
1. open account import
2. upload a faulty file (e.g. gender null)
3. try to import
4. you get correct behaviour
5. now try to select the same file again
6. nothing happens

luisa-beerboom:
Closes #3100 

The original method of reproduction should be fixed by now, I don't know how to reproduce it on the other imports.

I would suggest using a purposefully broken backend to test this.
Simply temporarily add the line `raise ActionException("random text")` at the top of the the `participant_import.py`'s `validate_entry` method, restart your dev environment, then try uploading a participant csv file in the participant import.

### DIFF
--- a/client/src/app/ui/modules/import-list/components/via-backend-import-list/backend-import-list.component.ts
+++ b/client/src/app/ui/modules/import-list/components/via-backend-import-list/backend-import-list.component.ts
@@ -252,6 +252,9 @@ export class BackendImportListComponent implements OnInit, OnDestroy {
         this._importer.clearAll();
         this._requiredFields = this.createRequiredFields();
         this._importer.currentImportPhaseObservable.subscribe(phase => {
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
function

bastianjoel (approved):

PR #3103, opened 2023-12-14
Prevent some linter warnings and make OS4 more accessiblle

### LINKED ISSUE #3079: Fix linter warnings for accessibility
When executing the "npx ng lint --fix" command several accessibility related linting warnings are reported. We should fix them. 
If they are fixed the behaviour of the client should change (at least slightly) to be more accessible.

What was changed or what different behaviour is to be expected should be documented here or in the PR which fixes the warnings

Elblinator:
closes part of #3079 

Notes for testing:
which objects changed (are now reachable with tab and usable with enter)
- meeting > setting-list 
- copyright sign (and more)
- chat groups
- in file-list visual for focused files is different
- in file-liste folder are focusable
- "Hide more text"/"Show full text" (e.g. see motions -> comment)


jsangmeister:
ping @rrenkert @bastianjoel

bastianjoel (approved):

rrenkert (changes requested):
Metting->Settings is available via `enter` the backspace navigation is missing but should be availbale.
File-list is lacking the navigation to edit dialog via row item 'groups'

client/src/app/ui/modules/file-list/components/file-list/file-list.component.ts
```diff
@@ -89,6 +89,9 @@ export class FileListComponent extends BaseUiComponent implements OnInit, OnDest
         this._hiddenInMobile = [`indicator`, ...cols];
     }
 
+    @Input()
+    public isInaMeeting = true;
```
bastianjoel, line 93:
```suggestion
    public isInMeeting = true;
```

Elblinator:
> Metting->Settings is available via enter the backspace navigation is missing but should be availbale.
You can navigate back with alt + arrow left and you do not need tab to navigate back

rrenkert (approved):

PR #3114, opened 2023-12-21
Fix CI on feature branches

jsangmeister:


jsangmeister:
The github actions configures the submodules such that only the main branch is fetched, even if another one is directly requested. Now it's a very cumbersome command, but it works :D

bastianjoel (approved):

PR #3119, opened 2023-12-22
Clean up several constructors

Elblinator:


client/src/app/gateways/export/csv-export.service/csv-export.service.ts
bastianjoel, line 1: `inject` should only be used when needed. Using the constructor is still the default way to add services.

bastianjoel (changes requested):

Elblinator:
this PR is closed due to the false base and too many change recommendations 
see #3129

jsangmeister:
Just FYI, you can simply change the base by clicking the "edit" button next to the title ;)

PR #3120, opened 2024-01-03
Fix access group setting in mediafiles

### LINKED ISSUE #3088: Wrong access groups in mediafile list view
* Create new folder (without any access groups)
* open edit dialog for created folder
* select "Admin" group
=> in background: "Admin" group is already set in mediafile list view
* close dialog via "Cancel"
=> "Admin" is still visible in list view
* Reload
=> "Admin" is gone
mediafile-list

luisa-beerboom:
Closes #3088


Elblinator (approved):
code and function

PR #3123, opened 2024-01-04
Stop motion list from showing forwarded motions

### LINKED ISSUE #3104: Motions: Forwarding issue
**Description:** If a meetings contains forwarded motions, it is possible to see 'empty shells' (or zombie motions ^^) of forwarded motions in the motion list. If the user opens another menu or reload the page, the wrong motion isn't visible anymore.

Screenshot_20231218_085459

**Reproduction:**
preconditions: two committees; each with a meeting; a forwarding option from meeting A to B

1. open meeting A
2. open motions > three dot menu > workflow > simple workflow > allow forwarding of motions > set checkbox in status accepted 
3. head back to motions and create a motion. change status to accepted. forward it to meeting B.
4. now click on motions -> The created motion and the forwarded motion should appear in the motion list
5. same happens in meeting B: open it, open the forwareded motion and go back to motion list. 

**What should happen:**
If a users opens a forwareded motion and goes back to the motion menu, zombie compies of the forwarded motions should not appear. The list should only display the motions of the meeting.

luisa-beerboom:
Closes #3104


Elblinator (approved):
function

bastianjoel (approved):

PR #3125, opened 2024-01-04
Use user short name in chat messages

### LINKED ISSUE #3090: Chat should always show a name next to a message
We were surprised during our first meeeting with OpenSlides v4 that the chat did not show any names next to the messages. This was quite confusing, as every message could have been by any of the participants. Only now we found out, that this wasn't a bug or an access rights problem, but due to the fact that we are a nickname-based organization. We did not fill out the first name or last name fields.

As far as we can tell, OpenSlides v3 showed the username, if no other name was set. This fallback mechanism could be introduced again, possibly behind an option in case this behaviour was chosen deliberately to keep the usernames secret. If there is no name to show, it would help to have some generic Jon/Jane Doe name displayed or some other means to tell apart unnamed participants.

**Describe alternatives you've considered**
We have manually added the nickname as last name to each participant, but this is cumbersome and redundant.

reiterl:
Resolves #3090


Elblinator (approved):
function

bastianjoel (approved):

PR #3126, opened 2024-01-04
Fix preview on submitter edit

### LINKED ISSUE #3051: Motion > submitter list: missing preview, if list is empty
**Desription:** If a user edits the submitter list of a motion and clears it, the preview of new added submitters is empty until the new list is saved.

**Cleared submitter list:**
Screenshot_20231127_095907

**Adding user "amep"**
Screenshot_20231127_095955

**list is still empty:**
Screenshot_20231127_100020

**After saving, the user is visible:**
Screenshot_20231127_100046

**Reproduction:**
1. open a meeting
2. open motions > create a new motion
3. If motion create is done and the motion is opened: edit the submitter list > delete all submitter and save it > now open the submitter edit again and add another user. > users will not be displayed until saving. > preview of added submitters didn't work in this case.

**What should happen:**
Adding new submitter in an empty list must display the added user before saving it.

luisa-beerboom:
Closes #3051


Elblinator (approved):
function and code

PR #3139, opened 2024-01-12
Fix motion 'default change reco mode' setting

### LINKED ISSUE #3105: Settings > Motions > setting 'default version for...' didn't work
**Description:** The setting 'Default version for change recommendations and projection of motions' didn't work as it should. For example, if a user set this to final version, the final version of a motion should be displayed by opening a motion. This doesn't happen. The original version instead is opened. 

affected setting:
Screenshot_20231218_092114

wrong motion version opened with this setting:
Screenshot_20231218_092230

**Reproduction:**
1. open a meeting
2. open motions > create a motion > create an amendment to it > change the status of the amendment and motion to accepted. > create in the motion a final version
3. open settings > motions > change the setting 'Default version for change recommendations and projection of motions' to Final Version.
4. open motions > open the motion -> final version should now open, but original version is opened.

**What should happen:**
By changing the setting 'Default version for change recommendations and projection of motions' to a prefered a version, the correct motion version should open, if it is available. 

**Hint:** By fixing this, please check the other versions if they are displayed correctly.

luisa-beerboom:
Closes #3105


Elblinator (changes requested):
If "final version" or "changed version" is selected in the settings, then the "original version" is initially displayed in the motion-detail view

luisa-beerboom:
That's because, until a change recommendation is actually made, the "changed version" is the same as the original version. Also changing the code so that changed mode, if it is set as default, is triggered under all circumstances, even those that don't include a recommendation, would, with the current way that mode is implemented, cause the view to break a bit (i.e. no possibility to create change recommendations and no way to switch modes).

Similar reasoning for the final version. That one needs at least one of "change recommendations" or "accepted amendments" to be present for the motion. Else it doesn't work.

If the requirements for the views are there, they will be displayed in the correct format however.
Does this really need fixing (and if yes: to what extend?), or is this alright as is?

jsangmeister:
ping @MSoeb

bastianjoel (approved):

rrenkert (commented):
With the latest changes the "final" and "changed" version show the original version of the motion also if amendment state and motion state is accepted.
@MSoeb  can explain that in more detail.

MSoeb (approved):

bastianjoel (approved):

PR #3141, opened 2024-01-12
Make it possible to set reset-password input field to visible

### LINKED ISSUE #3029: Password reset form, possible to make visible
When a user uses the reset form (via reset-mail), the password-fild only shows dots (of cause).

It would be nice to add a eye symbol on the right side, to show the password.

luisa-beerboom:
Closes #3029


Elblinator (changes requested):
please 
give the icon the same theme colour as the other login icon 
a padding of 12px 
and remove the shadow (the other login icon also doesn't have a shadow -> meantion as reference)

bastianjoel (approved):

Elblinator (approved):
function

PR #3163, opened 2024-01-19
List of speakers export

### LINKED ISSUE #3156: Add a meeting-level speaker list view
A list view for all speakers in all lists of speakers in the meeting should be implemented. The possible sorting options are "by name" and "by date" (equals `begin_time`). Each speaker should have a link to its respective list of speakers. Additionaly, the list should be exportable as a CSV file.

bastianjoel:
resolves #3156

We might need to discuss if the exported data is sufficient. 

Also I would stress that it might make sense to move the page to the agenda as I think that LOS is more related to the agenda then to participants.

### DIFF
--- a/client/src/app/gateways/export/csv-export.service/csv-export-for-backend.service.ts
+++ b/client/src/app/gateways/export/csv-export.service/csv-export-for-backend.service.ts
@@ -54,7 +54,7 @@ export class CsvExportForBackendService {
         const header = columns.map(column => {
             let label = ``;
             if (isPropertyDefinition(column)) {
-                label = column.property as string;
+                label = (column.label || column.property) as string;
             } else if (isMapDefinition(column)) {
                 label = column.label;
             }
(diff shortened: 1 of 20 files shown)


luisa-beerboom (commented):
I haven't found anything to complain about concerning code style, however I can't go further in my review, since neither the PR description nor the title tell me what exactly this PR is supposed to achieve. There is also no issue linked. Please rectify that.

Elblinator (changes requested):
The title-columns in the CSV-file should be changed
- The Name column should be splitted into three columns (Title, first name ans surnam)
- an additional row for the participant number should be added
- The naming for the Name column should change (in order from left to right) = Title,  First name, Surname, Structure level, Participant number, Speech-Type, Time, Element

The file-name should be changed to lists_of_speakers.csv

Updated 22.01.24 15:35

luisa-beerboom (approved):

luisa-beerboom (approved):

Elblinator (approved):
function

PR #3164, opened 2024-01-19
Add new motion fields

### LINKED ISSUE #3153: Add new motion fields
see https://github.com/OpenSlides/openslides-backend/issues/2119, which must be implemented first

- [x] implement relations
- [x] add new settings
- [x] add dropdowns to set the new relations in which participants can be selected, analogously to, e.g., the supporter dropdown, with the exception that only a single participant can be selected. Show the dropdowns only if the respective setting is enabled and the user has `motion.can_manage`
- [x] add filter in the motion list to filter by the relations. The filter is also only visible if the respective settings are enabled

jsangmeister:
resolves #3153

needs https://github.com/OpenSlides/openslides-autoupdate-service/pull/844 and https://github.com/OpenSlides/openslides-backend/pull/2201


rrenkert (changes requested):
1. The chip content does not follow the dark mode colors
2. using the small "x" near the label does not clear the chip content. A selected user is still visible and it can be assumed that the value is saved.

Everything else looks good.

jsangmeister:
Should be fixed now.

jsangmeister:
As discussed with @emanuelschuetze and @rrenkert: The new fields are changed to list fields to new models, analogously to the submitters (see https://github.com/OpenSlides/openslides-backend/issues/2196). The client components can be merged with the submitter one, as they all should work exactly the same.

bastianjoel (approved):

jsangmeister:
I pinged @ostcar in https://github.com/OpenSlides/openslides-autoupdate-service/pull/844 to impleemnt the autoupdate side of things.

rrenkert (approved):

jsangmeister:
@bastianjoel create a user with only `motion.can_manage_metadata`, not `motion.can_manage`. The new fields will be shown, but after creating an editor, it is not delivered via autoupdate.

jsangmeister:
Fixed.

PR #3165, opened 2024-01-19
Change styling in structure Level projection slides

Elblinator:


bastianjoel (approved):

rrenkert (changes requested):
The structure_level text and time font is to big. the name should be visible in all cases and sizes.

rrenkert (approved):

PR #3179, opened 2024-01-22
Allow stopping interposed questions without speaker

### LINKED ISSUE #3168: Change interposed Questions warn dialog
**Current behaviour:**
If an interposed question is stopped while no speaker is selected the following dialog opens:
sel

**Reproduction:**
1. navigate to settings -> LoS -> activate interposed question
2. project something onto the autopilot
3. navigate to the autopilot
4. put someone on the LoS and start their contribution
5. activate the interposed question
6. start and stop it
7. dialog opens

**Wanted behaviour:**
There should be a text which says "Are you sure you want to stop the interposed question without selecting a speaker, the "Last speakers"-List will display this interposed question with a "Deleted user"?"
And the dialog needs an additional "continue" button

bastianjoel:
resolves #3168

Also includes a fix for last speakers speaking times ignoring pause times.


luisa-beerboom (approved):

MSoeb (changes requested):
This PR has a problem: The "select a speaker"-menu opens the same dialog as the stopp-button. This should not be. 

What should happen: By opening the "select a speaker"-menu another dialog should appear which only offers the possible action to add a user. 

screenshot-client3179 staging openslides com-2024 01 23-10_09_27

MSoeb:
Addtionally I suggest to improve the dialog text "Are you sure you want to stop the interposed question without selecting a speaker, the "Last speakers"-List will display this interposed question with a "Deleted user"?".

New text idea: "Are you sure you want to stop the interposed question without selecting a speaker? Confirmation will add the interposed question to the 'Last speakers'-list as an 'Anonymous (Or Unknown) user'.

Also I think it is better to add a unknown or anonymous user to the last speakers list instead of a deleted user. We should only use deleted user, if a user is deleted and not if the user is unknown. -> Solution: Change Deleted User to Unknown or Anonymous user on the last speakers list.

@rrenkert Please check, if it's okay from your point of view.

bastianjoel:
I also think that the text is quite long for a title. Maybe we should use a title and a description in that dialog.

MSoeb:
@bastianjoel Yes, good advice. We should split the title from the following sentence. 

Please take care: If adjusted some points my last replies. 1. The wording of the text a little bit. (Unknown instead of anonymous user). 2. Dialog menu for adding users: I've added the dialog question, which should appear.

rrenkert:
> Addtionally I suggest to improve the dialog text "Are you sure you want to stop the interposed question without selecting a speaker, the "Last speakers"-List will display this interposed question with a "Deleted user"?".
> 
> New text idea: "Are you sure you want to stop the interposed question without selecting a speaker? Confirmation will add the interposed question to the 'Last speakers'-list as an 'Unknown user'.

Yes, we should use this text to warn the user.

> 
> Also I think it is better to add a unknown or anonymous user to the last speakers list instead of a deleted user. We should only use deleted user, if a user is deleted and not if the user is unknown. -> Solution: Change Deleted User to Unknown user on the last speakers list.

I am not sure if we can decide if a user was deleted or if no user was set in the client. @bastianjoel can we?

bastianjoel:
No we can't

bastianjoel:
- Remove dialog when stopping.
- Add icon to last speakers where one can assign a speaker to interposed questions without speakers. 
- Rename deleted user for interposed questions in last speakers to "Unknown".

Elblinator (changes requested):
if an interposed question is stopped then it is saved with time 0:00

Elblinator (approved):
function

MSoeb:
Note: If a speaker has an interposed question and the stop-button of the main speaker is clicked, an error message appears at the bottom. This seems like a backend issue. Therefor the PW will be approved.

MSoeb (approved):

luisa-beerboom (approved):

jsangmeister:
> Note: If a speaker has an interposed question and the stop-button of the main speaker is clicked, an error message appears at the bottom. This seems like a backend issue. Therefor the PW will be approved.

Cannot reproduce. Please open a detailled backend issue if the problem persists.

PR #3180, opened 2024-01-22
Fix los detail moderator note not loaded

### LINKED ISSUE #3174: Moderation note is not loaded correctly
**Current behaviour:**
If you reload a los/detail view then the moderation note appears empty and if you edit it it also appears empty

**Reproduction:**
1. navigate to a topic -> LoS detail view 
2. write something in the Moderation Note (it"s saved and displayed correctly)
3. reload the page 
4. the content of the moderation note is gone
5. navigate back to the agenda
6. navigate to the LoS, Moderation note content is loaded

**Wanted behaviour:**
The moderation Note content should always be loaded if the moderation note is visible

bastianjoel:
resolves #3174


Elblinator (approved):
code-review

MSoeb:
Checked it also: Found no problems with this fix. Works as it should.

PR #3182, opened 2024-01-22
Stop speaker with interposed questions warning

### LINKED ISSUE #3169: Change behaviour of interposed question
**Current behaviour:**
If the parent from an interposed question is stopped the interposed question is left behind without it's parent

**Wanted behaviour:** 

-  if the child does not have a start time yet, then it should be dismissed
- if the child has a start time then the child should be saved and the end time  should be the same as the parents (if no speaker is selected then it should be saved as Deleted User)
- A dialog should open which warns the user of each of this 
  - Title "Are you sure you want to end a contribution which still has interposed question(s)?"
  - content = "_(number of interposed questions without start time)_ interposed questions will be cleared, _(number of interposed questions which have a start time)_ will be saved with an inaccurate time,  _(number of interposed questions which have a start time and NO selected User)_ of them will be saved with 'Deleted User' as their speaker"
  - the content should only show the categories which have a number bigger than zero

bastianjoel:
resolves #3169


luisa-beerboom (approved):

MSoeb (approved):
Works as it should. 

Note: Maybe we need to improve the descriptions in the dialog. But this will be another issue.

PR #3176, opened 2024-01-22
Speaking times added to los detail view

reiterl:
Draft version, replaces the speakers time management component.

PR #3183, opened 2024-01-23
LoS extension UI changes

### LINKED ISSUE #3166: [Meta] Two UI- change requests/bugs in los-extension
- [x]  in a LoS the participant is currently displayed as "name ( number) structure level"
  - [x]  1.  The wanted display should be "name (structure level middot number)"
  - [x]  2. the text color should be the like the name and the font-style should be the same
 (see main as reference) 

- [x] the interposed question:
  - [x]  1. should not be bold 
  - [ ]  2. "Select or search new speaker" should be in box brackets

- [ ] The speaking time card: the color line and the structure level name should be aligned right
- [ ] in the autopilot:
  - [ ] 1. there should be a max-distance 75px between the left side of mat-cards and the right side (speaking times)
  - [ ] 2. if mobile view is activated then the speaking time card should jump directly under the LoS

bastianjoel:
resolves the checked points in #3166 

~~currently waiting for #3176~~


Elblinator (commented):
Everything is fine, besides the fix for: 
1. there should be a max-distance 75px between the left side of mat-cards and the right side (speaking times)
That fix looks suboptimal. @MSoeb Can you please take a look at it?

MSoeb (approved):

PR #3190, opened 2024-01-24
Speaking times permissions update

### LINKED ISSUE #3187: Widget Speaking time: permission problem
**Description:** Users who don't have the permission to administrate the projector (Can manage the projector) see the 3-dot-menu in the new widget speaking time. The menu can be used, but content can't projected, because this is currently not working. Nevertheless it's wrong, that user can see it. 

**Example: 3-dot-menu on the right side**
image

**Repro:** 
1. open a meeting
2. open settings > list of speakers > enable the new widget
3. open agenda > create an item > project it
4. open participants: create a new user without the group delegates
5. log in with the new created user and open the autopilot you should see the new widget with the wrong permission

**What should happen:** 
The three-dot-menu in the new widget should only be available for persons with the permission 'Can manage the projector'. Other user should not see it.

reiterl:
Resolves #3187

### DIFF
--- a/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.html
+++ b/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.html
@@ -105,7 +105,7 @@
             <mat-card-content>
                 <div class="action-title">
                     <p class="subtitle-text" translate>Speaking times</p>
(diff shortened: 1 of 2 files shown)


bastianjoel (approved):

bastianjoel:
I will add this to #3176 directly.

PR #3193, opened 2024-01-24
Init moderator notes form part on toggle edit

### LINKED ISSUE #3186: Moderation notes: Editing issue
**Description:** Editing existing moderation notes in a different tab or as another user is currently bugged. By opening the edit tool, the text box is empty. The original text is not visible. Editing is not possible.

**Existing moderation notes:**
image

**Editing mode without existing content:**
image

**Repro:** 
1. open a meeting
2. create an agenda item and a moderation note
3. try to edit it in another tab or as another user

**What should happen:**
Editing should display the original text, so that it can be edited.

reiterl:
Closes #3186


rrenkert (approved):
Works as expected now.

PR #3197, opened 2024-01-25
Clean extra moderator notes from autopilot

reiterl:
Clean extra moderator notes parts from the autopilot, this code was lava-code.


Elblinator (approved):
function and code

PR #3217, opened 2024-01-26
Add participant mandate switch dialog

### LINKED ISSUE #3157: Add tool to switch participant data
Two participants of a meeting should be selectable in the tool. Both are then presented side-by-side with the following data: full name, structure level, groups, number, presence. Upon confirmation, the groups and the number of both participants are switched (via `user.update`).

jsangmeister:
closes #3157 

renamed two folders along the way as they mistakenly had the suffix `.service`

The wording as well as the layout is just a suggestion yet, I'm sure you'll have better ideas for that ;)

@bastianjoel Currently, if the backend action throws an error for some reason, the success message will be displayed anyway. IIRC you had a similar problem recently, what was the fix for that, i.e. where can I access the "success" status of an action?


client/src/app/gateways/error-mapping/error-map-utils.ts
bastianjoel, line 1: I am not entirely sure if this could have side effects as the user update is also used in places where the promise is not awaited. If I understand this right this would result in users not getting notified if an error occurred there.

bastianjoel (commented):
I am not entirely sure if this could have side effects.

client/src/app/gateways/error-mapping/error-map-utils.ts
jsangmeister: Okay, I see. What's the solution for this then?

bastianjoel (approved):

jsangmeister:
As per suggestion by @luisa-beerboom, I changed the arguments of the http service to use a parameter interface to make the code less redundant.

bastianjoel (approved):

PR #3219, opened 2024-01-29
Create user from string with username generated in backend

### LINKED ISSUE #2921: Fix "+ create User" action in participant search fields
**Current behaviour:**
If you search for a user in a participant search field and the participant is not in the participant-list then it is possible to run into one Error: ```Error: A user with the username <name> already exists```

**Reproduction:**
1. create the account Firstname: xx, Surname xx; automatically will create username: xxxx
2. navigate into a meeting > motion > LoS
3. Search for a participant xxxx
4. you should get the option to "+ create User"
5. if you activate the option the Error is thrown: ```Error: A user with the username xxxx already exists```

**Wanted behaviour:**
1. the input field: `Select or search new speaker ...` should send the whole string as the surname 
    If you have the input `Kim Meier` and create a new user, then the created account should have no first name and the surname `Kim Meier`
2. the User which is created should be inactive
3. after a user is created a snackbar should open which says: "A user with the Surname '<name>' and the username '<username>' was created"

reiterl:
Resolve #2921

In this pullrequest I have slightly changed the createFromString function, to allow the backend to generate the username.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
+++ b/client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
@@ -161,9 +161,9 @@ export class ParticipantSearchSelectorComponent extends BaseUiComponent implemen
         );
         this.snackBar.open(
             this.translate
-                .instant(`A user with the username 'USERNAME' and the surname 'LAST_NAME' was created.`)
+                .instant(`A user with the username 'USERNAME' and the first name 'FIRST_NAME' was created.`)
                 .replace(`USERNAME`, user.username)
-                .replace(`LAST_NAME`, user.last_name),
+                .replace(`FIRST_NAME`, user.first_name),
             this.translate.instant(`Ok`)
         );
     }
--- a/client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
+++ b/client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
@@ -161,9 +161,9 @@ export class ParticipantSearchSelectorComponent extends BaseUiComponent implemen
         );
         this.snackBar.open(
             this.translate
-                .instant(`A user with the username 'USERNAME' and the first name 'FIRST_NAME' was created.`)
-                .replace(`USERNAME`, user.username)
-                .replace(`FIRST_NAME`, user.first_name),
+                .instant(`A user with the username '%username%' and the first name '%first_name%' was created.`)
+                .replace(`%username%`, user.username)
+                .replace(`%first_name%`, user.first_name),
             this.translate.instant(`Ok`)
         );
     }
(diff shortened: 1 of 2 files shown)


Elblinator (commented):
> the User which is created should be **inactive**

currently the created account/participant is active, please change this to inactive

> after a user is created a snackbar should open which says: "A user with the Surname 'bla bla' and the username 'blabla' was created"

this is currently missing, please add

reiterl:
I have added code for set inactive and opening the snackbar.

client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
```diff
@@ -149,9 +151,16 @@ export class ParticipantSearchSelectorComponent extends BaseUiComponent implemen
...
-        const newUserObj = await this.userRepo.createFromString(username);
+    public async createNewSelectedUser(name: string): Promise<void> {
+        const newUserObj = await this.userRepo.createFromString(name);
         this.emitSelectedUser({ userId: newUserObj.id, user: newUserObj });
+        const user = await firstValueFrom(
+            this.userRepo.getViewModelObservable(newUserObj.id).pipe(filter(user => !!user))
+        );
+        this.snackBar.open(
+            `A user with the username '${user.username}' and the surname '${user.last_name}' was created.`,
+            `Ok`
```
bastianjoel, line 162: This is not translatable.

bastianjoel (changes requested):

client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
reiterl: Use instant and replace to add a translation here.
The keywords USERNAME and LAST_NAME are used in the message.

Elblinator (changes requested):
given that the name is saved in first and last or only the first name:
Please change the snackbar message from the last name/ Surname to the first name

reiterl:
Changed the snackbar message to use first name instead of last name.

client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
```diff
@@ -161,9 +161,9 @@ export class ParticipantSearchSelectorComponent extends BaseUiComponent implemen
         );
         this.snackBar.open(
             this.translate
-                .instant(`A user with the username 'USERNAME' and the surname 'LAST_NAME' was created.`)
+                .instant(`A user with the username 'USERNAME' and the first name 'FIRST_NAME' was created.`)
```
jsangmeister, line 164: AFAIK we currently use two different replacement systems for translations:
- Simply using `{}` and replacing by order, for example here: https://github.com/OpenSlides/openslides-client/blob/main/client/src/app/site/modules/translations/marked-translations.ts#L51
- encasing the variable in percent signs, I think only used when sending emails: https://github.com/OpenSlides/openslides-client/blob/main/client/src/app/site/services/user-controller.service.ts#L209

Unless @luisa-beerboom or @bastianjoel know that you method is used anywhere, I would argue that we do not want a third way of translating variables and you should change it to one of the above.

jsangmeister (changes requested):

Elblinator (approved):
function

client/src/app/site/pages/meetings/modules/participant-search-selector/components/participant-search-selector/participant-search-selector.component.ts
reiterl: I changed the var names to %username% and %first_name%. It is just a renaming.

jsangmeister (approved):

bastianjoel (approved):

PR #3222, opened 2024-01-30
Limit distance between columns in autopilot

### LINKED ISSUE #3198: [Meta] Design from speaking-times.component.html
#3166 -> the not finished tasks are moved to this issue

- [x] The speaking time card:
   - [x] the color line, the countdown and the structure level name should be aligned right 
   - [x] the color line should be an element of it's own (maybe hline) and as long as the whole row minus the icon
   - [x] the more_vert icon should be right aligned
   - [x] the countdown should be in mono-font

- [x] in the autopilot:
  - [x] 1. there should be a max-distance 75px between the left side of mat-cards and the right side (speaking times)
  - [x] 2. #3196

bastianjoel:
resolves #3198

### DIFF
--- a/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.html
+++ b/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.html
@@ -4,45 +4,38 @@
     </div>
 </os-head-bar>
 <div class="content-container">
-    <div class="col-main">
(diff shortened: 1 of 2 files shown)


Elblinator (commented):
The distance is now as it is wanted, but the speaking times card is max. 547.99px wide

if the space is available, then the speaking times card should be as wide as the other cards

PR #3223, opened 2024-01-30
Export speaker agenda item number

### LINKED ISSUE #3220: Enhance topic column in LoS export
Prefixes like "TOP X" should be exported as well in the same column.

bastianjoel:
resolves #3220

Also fixes missing agenda item requests from autoupdate.


jsangmeister (approved):

MSoeb (changes requested):
Prefix is corrected correctly, but there is an Autoupdate error:

If the data is exported without opening the agenda menu first, the prefixes won't be exported. 

Repro:
0. An instance with existing data in the agenda is needed.
1. open participants without opening the agenda first > three dot menu > contributiions > export the data. prefixes are not exported.
2. open agenda and do step 1. again. -> prefixes are exported.

rrenkert (approved):

emanuelschuetze (approved):

PR #3224, opened 2024-01-30
Structure level selection for interposed questions

### LINKED ISSUE #3221: Add structure level to interposed questions
An interposed question with a selected speaker should display their structure level in the same way as a normal speaker and should allow to select a specific level if the user has multiple structure levels. This does not impact the timer of the structure level in any way and is purely for visual purposes.

bastianjoel:
resolves #3221

### DIFF
--- a/client/src/app/gateways/repositories/speakers/speaker-repository.service.ts
+++ b/client/src/app/gateways/repositories/speakers/speaker-repository.service.ts
@@ -52,6 +52,7 @@ export class SpeakerRepositoryService extends BaseMeetingRelatedRepository<ViewS
         info: {
             speech_state?: SpeechState;
             meeting_user_id?: Id;
+            structure_level_id?: Id;
         },
         viewModel: ViewSpeaker
     ): Promise<void> {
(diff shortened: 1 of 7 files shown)


MSoeb (dismissed):
Found two problems:

- If a user with a structure level which isn't currently displayed on the speaking times widget is added, the new structure level appears on the speaking times widget. -> What should happen: Adding a user as an interposed question with a new structure level should not add his structure level to the speaking times widget. Interposed questions and the speaking times widget are not connected.

- I guess it's another issue, but for completion: Start - Pause - Start - Pause an interposed question will trigger the structure level timer. This is wrong. - This should not happen.

bastianjoel:
Both problems are not directly related to this PR.

jsangmeister:
> Found two problems:
> 
> * If a user with a structure level which isn't currently displayed on the speaking times widget is added, the new structure level appears on the speaking times widget. -> What should happen: Adding a user as an interposed question with a new structure level should not add his structure level to the speaking times widget. Interposed questions and the speaking times widget are not connected.

This will be fixed by https://github.com/OpenSlides/openslides-client/issues/3227.

> * I guess it's another issue, but for completion: Start - Pause - Start - Pause an interposed question will trigger the structure level timer. This is wrong. - This should not happen.

Did you test with the current backend feature branch? This should no longer happen, as it was fixed by https://github.com/OpenSlides/openslides-backend/pull/2164.

Elblinator (approved):
code and function

PR #3225, opened 2024-01-30
Fix problem with search selector and keydown

### LINKED ISSUE #2721: Fix search field for LoS
**Current behaviour:**
If you add someone via the search field and only one result is left then you have to select this person with your mouse. 
And if you want to cancel searching in the search field and press escape to do so, nothing happens

**Recreation:**
1. navigate to a LoS 
2. search for a participant
3. type in a participant name until only one result is left
4. try arrow down to select the last participant or press enter to select the participant 
5. nothing happens
6. now press escape to exit the search 
7. nothing happens again

**Wanted behaviour:**
- [ ] even if only one participant is left in a search the down arrow should be able to select the participant 
- [ ] and if only one participant is left then pressing enter should immediately add the participant
- [ ] pressing escape while in the search field should cancel the search

reiterl:
Closes #2721

### DIFF
--- a/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.html
+++ b/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.html
@@ -22,7 +22,7 @@
             class="background-card"
             [formControl]="searchValueForm"
             placeholder="{{ 'Search' | translate }}"
-            (keydown)="onSearchKeydown($event)"
+            (keydown.space)="$event.stopPropagation()"
         />
     </div>
     <ng-container *ngIf="multiple && showChips">


Elblinator (commented):
Tested with chromium

Please clarify which of the wanted behaviours you wanted to fix: 
**Wanted behaviour:**
- [ ] even if only one participant is left in a search the down arrow should be able to select the participant 
- [ ] and if only one participant is left then pressing enter should immediately add the participant
- [ ] pressing escape while in the search field should cancel the search 

I cannot confirm the fix of any of them

Elblinator (approved):
re tested, there was a problem with the test instance

bastianjoel (changes requested):
Seems like `onSearchKeydown` is now unused and could be deleted. 

Also I am still a bit skeptical that this works as you are switching from whitelisting the propagation of three keys to blacklisting only one key.

Elblinator (commented):
The first option should always be focused/highlighted

if shift is pressed the focused or first participant is put on the list -> shift should not do anything. only enter should put someone on the list

Alt + "arrow-down" or "arrow-up" closes the panel, the alt key + different key should not  close the options

under unknown circumstances the participants are in a mat-pseudo-checkbox/ selected state, this should be prevented (it is possible that this is its own issue/bug)

If you search a participant via the participant number or structure level and only one participant is left, then enter is not adding the participant to the list

reiterl:
Okay, my pr cannot handle that stuff. It is just a small one. So I close it here.

PR #3228, opened 2024-01-31
Fix navigation of reset password confirm in no data case

### LINKED ISSUE #3070: Password reset page doesn't navigate back to login if there is no data
Reproduction:
1. Without being logged in enter the url <your instance url>/login/forget-password-confirm
2. You will be sent to the password reset page and there will be a snackbar message saying that the link is broken

What's wrong:
The snackbar message is correct and should show up, but the user _should_ also be re-routed to the login page, which doesn't happen.

reiterl:
Resolve #3070 

Direct send to login.


Elblinator (approved):
function and code

PR #3245, opened 2024-02-02
Fix reset default settings

### LINKED ISSUE #2589: Fix "reset to factory defaults" in settings
**Description:**
In a meeting: it is not possible to restore the factory settings for some of the setting pages (main settings page, general settings page and motions settings page)

**Wanted behaviour:**
it should be possible to restore the factory settings

Elblinator:
closes #2589 

I decided to prevent the workflows to be reset to a predetermined id (like I did with the meeting name/title), because it is possible that the workflow id does not exist anymore, and will therefore thrown errors


bastianjoel (changes requested):
It would be better if you use the setting keys instead of the labels.

bastianjoel (approved):

rrenkert (approved):

PR #3254, opened 2024-02-03
Fix user edit for illegal legacy username

### LINKED ISSUE #3250: User edit: Filter out unchanged data from payload and change validation for username
The payload sent with the user.update action should not include fields that the operator hasn't changed, please filter them out somehow. According to @Elblinator this may already be implemented, so maybe this point is mute. If this _has_ to be implemented, the organization settings already has this and may serve as an example.

In either case there is a definite requirement to change the username validation so that it allows old values regardless of current validity. I.e. allow usernames with spaces as long as it is the username in the database.

By order of @rrenkert it should be done based on latest stable!

luisa-beerboom:
Closes #3250 

Should filter out most unchanged keys if it is an update, please check if this breaks _any_ field in account or participant edit.
Also check for create nothing should be changed.
Also check what happens if you edit a user, save, and then edit again.

This PR _should_ also make it possible to save even if a legacy username with empty spaces is left unchanged in the preview. Idk, I couldn't test it

### DIFF
--- a/client/src/app/site/modules/user-components/components/user-detail-view/user-detail-view.component.ts
+++ b/client/src/app/site/modules/user-components/components/user-detail-view/user-detail-view.component.ts
@@ -314,7 +314,8 @@ export class UserDetailViewComponent extends BaseUiComponent implements OnInit, 
     private propagateValues(): void {
         setTimeout(() => {
             // setTimeout prevents 'ExpressionChangedAfterItHasBeenChecked'-error
-            this.changeEvent.emit(this.getChangedValues(this.personalInfoForm.value));
+            const changes = this.getChangedValues(this.personalInfoForm.value);
+            this.changeEvent.emit(changes);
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

Elblinator (changes requested):
if you create a new participant and fill any of the four fields from step 1 (Given name, last name, email or username) then these values are not send in the payload from step 3 (creation) 
The values have to be re entered/change/ not pristine to be send in the payload

Elblinator (approved):
function

bastianjoel (approved):

luisa-beerboom:
Merging ist @peb-adr s responsibility

peb-adr:
Merging into `stable/4.1.x`.
New workflow for this is described in wiki:
https://github.com/OpenSlides/OpenSlides/wiki/Update-Workflow#hotfixes

Issue #2619, opened 2023-08-11
Generate-defaults script is broken


luisa-beerboom:
The script currently fails if executed and trying to fix it will inevitably lead one to the likes of `ERR_REQUIRE_ESM` errors.
Someone should probably take a look at this

bastianjoel:
As mentioned in #3006 we should maybe think about replacing `axios` with `fetch` to get rid of that dependency when doing this.

PR #3258, opened 2024-02-05
Use meta repository

jsangmeister:
I added it to the source folder since this makes most of the tooling work out of the box, e.g. in all docker setups, the src folder is mounted anyway. If we want to move it somewhere else, we would have to adjust all out setups.


jsangmeister:
I updated all `generate-*` scripts and removed axios as it's no longer needed. I could not fix the `generate-defaults` script, however (https://github.com/OpenSlides/openslides-client/issues/2619). @bastianjoel maybe you could have a look, I would like to have everything working again if we already restructure everything.

bastianjoel (approved):

Issue #2946, opened 2023-10-25
Amendment-detail fix change recommendation


Elblinator:
**Current behaviour:**
In Amendment-detail view (you need an amendment which changes more than one line)
- [x] A. If a change recommendation is added then the original version of the motion is displayed, not the Diff Version anymore. 
- [x] B. If the change recommendation is deleted then you need to reload the page to see the change recommendation anymore
- [x] B.1. If a change recommendation is added, the whole motion text can't be activated with the checkbox "Show entire motion text". By clicking it an error appears in the console. Error message "Cannot read properties of null (reading 'offsetTop')" 
- [x] C. If an amendment changes several lines and has one change request then the line numbers are not correctly displayed
- [x] D. If an amendment has a change recommendation then the option "show entire motion text" vanishes

Descriptions changed because Case C and D are resolved and A/B needed a rework.

**Reproduction:**
In a meeting:
1. create a motion with several lines/paragraphs (I recommend 7 lines)
2. create an amendment which changes several (but not all) lines in the motion
3. add a change recommendation to the amendment
    3.1. hover around the beginning of the sentence (a bit more to the left) until you find a blue cross-icon
    3.2. press the icon
    3.3. change something in the sentence and save (hint: doesn't matter, if it is a part of the change recommendation or the regular text.)
4. ( **Case A)** ) now the original version of the amendment is active, not the Diff Version (compare: in a motion the Diff Version is active after adding a change recommendation.)
5. **(Case B)** Original version must be active.-  Open the console. Now activate the checkbox "Show entire motion text". An Error message appears in the console. The entire motion is not displayed. -> Reload the page to switch back to a working version.
6. **(Case B.1)** Delete the change recommendation. -> CR stays active and visible. By deleting it twice, a correct error message appears.
7. reload the page -> the change recommedantion is gone

**Wanted behaviour:**
- [x] A. - After adding a change recommendation to an amendment, the Diff Version should be visible. It should follow the same logic like adding an amendment to a motion. Please check, if this is connected to the function "Default text version for change recommendations and projections of motions".
- [ ] B. - the change recommendation should vanish after deletion without reload
- [x] B.1 - activated checkbox "Show entire motion text" should show whole motion text
- [x] the displayed lines should be correct even if there is a change request
- [x] the optiojjn to display the whole text should always be available

Elblinator:
Case D) only happens if the Diff Version is selected. @emanuelschuetze @MSoeb is this a feature?

[...]

PR #3266, opened 2024-02-08
Some fixes amendment detail change recommendation

reiterl:
The motion detail with the line numbering is difficult.

This pull request should solve case A and B1 from #2946 
But there are still some strange behaviour in there.


Elblinator (changes requested):
B1 is fixed 

If the amendment has a change recommendation only the line with change recommendation is now displayed in the diff version:

reiterl:
I can see that, but it is the amendment diff displayed. I just set diff mode. These components are cans of worms.

reiterl:
I won't continue here.

bastianjoel:
Please update your PR to just fix part of the issue and maybe remove regressions, close it or provide further detail on how we should proceed with this.

reiterl:
As I won't continue here, I close this pr.

Issue #3067, opened 2023-12-04
Investigate `get_forwarding_meetings` and `motion.create_forwarded`


luisa-beerboom:
See OpenSlides/openslides-backend#2003 and OpenSlides/openslides-backend#2009

According to tracebacks in the logs it seems these two processes were called with insufficient payloads.

The `get_forwarding_meetings` presenter was seemingly called with a payload that didn't contain a `meeting_id`
The `motion.create_forwarded` action was seemingly called without a `meeting_id` and, another time, without an `origin_id`.
The latter action also has other required fields, but they would not have caused the same error in the backend.

Please take a good look at these actions in particular and check if the client can actually be made to send these types of payloads and, if yes, fix that

reiterl:
The get_forwarding_meetings presenter is called in the `motion-forward-dialog`. The `activeMeeting.meetingId` could be null in some cases, so add a check here.

reiterl:
I looked at the `MotionRepositoryService.createForwarded()` function. I don't find the reason, why meeting_id or origin_id could be null or undefined. Mmh, no idea. This seems to be the only function in which the CREATE_FORWARDED action is used.

rrenkert:
#3527 and #3529 are features in same scope. Can we have this one in mind when working on them?

rrenkert:
@luisa-beerboom is this still relevant?

luisa-beerboom:
If it hasn't been worked on yet, generally yes

PR #3270, opened 2024-02-09
Small improvement to the get-forwarding-meetings presenter call

reiterl:
Update the get-forwarding-meetings call to be more safe.

Part of #3067


bastianjoel:
Why has this the `needs investigation` tag?

reiterl:
The issue has "needs investigation", so I used it here too.

Elblinator (approved):
I tested the forwarding a bit 
It is possible that this solution only fixes part of the problem due to the unknnown circumstances of this bug

But everything worked right now

bastianjoel (changes requested):
I am not sure if this is the solution we want here. Your fix is just suppressing the error. 

I guess the problem here is `forwardingMeetingsAvailable` which is called within constructors of other components and could therefore be called before a meeting id is available. If this is the case this could result in the forwarding button not being displayed. 

To make this more safe I would suggest waiting for a meeting id to be available. The `ActiveMeetingIdService` contains `undefined` in its meeting id observable as long as it is not initialized which you could use to differentiate between "not in a meeting" and "meeting not loaded".

reiterl:
Mmh, I need to think about this issue, when I have more time.

bastianjoel (approved):

Elblinator (approved):

PR #3272, opened 2024-02-12
Fix to base search selector with null instead 0

### LINKED ISSUE #2528: fix new motion block in motion blocks
if a new block is created and added to the agenda, and then it is explicitely stated that it has no parent-element the following error occurs:
_Fehler: data.agenda_parent_id must be bigger than or equal to 1_

**How to recreate:**

1. navigate into motion > motion blocks
2. start adding a new motion block
3. while in the adding dialog add it to agenda 
4. on the field "parent agenda item" select the empty line to indicate nothing as a parent element
5. save the new block
6. the new block is not saved and the Error occurs: data.agenda_parent_id must be bigger than or equal to 1

**wanted behaviour:**
the motion block should be addable to the agenda with explicitely sying that is has not parent element

reiterl:
Resolve #2528

### DIFF
--- a/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.ts
+++ b/client/src/app/ui/modules/search-selector/components/base-search-selector/base-search-selector.component.ts
@@ -225,7 +225,7 @@ export abstract class BaseSearchSelectorComponent extends BaseFormFieldControlCo
     protected noneItem: Selectable = {
         getListTitle: () => this.noneTitle,
         getTitle: () => this.noneTitle,
-        id: 0
+        id: null
     };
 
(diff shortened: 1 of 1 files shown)


jsangmeister:
This needs intensive testing! AFAIK, I tried to change this once in the same way, and it broke multiple components because they could not differentiate between "none-item selected" and "no item selected".

bastianjoel (changes requested):
With this change the none item will not be marked as selected in the search selector on at least some pages. 

Happens for example in the `Upload to` field at the file upload.

reiterl:
Then I discard my solution. Closed.

PR #3275, opened 2024-02-13
Fix motion multiselect forwarding error

### LINKED ISSUE #2764: Thrown correct Error if motion  in wrong status is forwarded
Edited: 13.02.24

**Current behaviour:**
Even if a motion is in a status which cannot be forwarded, you can "try" to forward it via multi-select.
This action will result in an error:
```Error talking to autoupdate service```

**Reproduction:**
1. create two meetings , one meeting needs to be allowed to forward motions to the other
2. go into the meeting which can forward
3. create a motion
4. make sure that this motion is not in a status which can be forwarded
5. activate multiselect > select the related motion > forward it
6. error is thrown

**Wanted behaviour:**
If a motion which is not allowed to be forwarded is selected via multi-select, then the forwarding option should throw a **correct** error, aka the wrong autoupdate error message should be prevented

jsangmeister:
resolves #2764


Elblinator (approved):
function and code

PR #3276, opened 2024-02-13
Extend shared worker

### LINKED ISSUE #3248: Connection to ICC service gets closed after 30min
This is a very critical issue as this prevents meetings from functioning correctly and browsers are used by about 100% of all users (https://en.wikipedia.org/wiki/Usage_share_of_web_browsers). As this does not happen for the autoupdate connection which is handled by the shared worker, we suspect that moving the ICC connection to the shared worker might sove this issue. Therefore, we should move the ICC handling to the shared worker ASAP.

### LINKED ISSUE #3259: Autoupdate long polling
https://github.com/OpenSlides/openslides-autoupdate-service/pull/833 added a fallback for SSE. This should be integrated into the shared worker. 

It might be a good idea to implement this together with #3248.

### LINKED ISSUE #3401: Move auth handling into shared worker
Currently logout events are mainly detected by the autoupdate module within the shared worker and passed to the `autoupdate-communication.service` which passes it to the `auth.service`. Also the `who-am-i` calls are currently separately done by client and shared worker. 

To make this process more simple, clear and reduce calls to the auth service we should handle `who-am-i` requests and auth change events completely in the shared worker. 
With #3276 the auth process is already extracted from the `autoupdate-strem-pool` into its own singleton (https://github.com/OpenSlides/openslides-client/pull/3276/files#diff-c249106595858b0e21e271410086dc5d077601eb375124f4ea2cbbdd2f05a94d) which should make this relatively simple.

bastianjoel:
resolves #3248 
resolves #3259
resolves #3401

This PR involves a lot of cleanup in the shared worker. Also tests were added. 

TODO:
- [ ] Autoupdate
    - [ ] Add tests to autoupdate part
    - [x] Rework unexpected stream resolve handling (session termination, restart)
    - [x] Fix issues with stream pausing
    - [x] Longpolling
- [x] ICC
    - [x] Implement icc protocol in shared worker
    - [x] Update icc adapter
- [x] Auth via shared worker


client/src/app/worker/http/auth.ts
```diff
@@ -0,0 +1,162 @@
...
+    private updateAuthPromise: Promise<boolean> | undefined;
+
+    private _authTokenRefreshTimeout: any | null = null;
+    private _waitingForUpdateAuthPromise = false;
+
+    constructor() {
+        this.updateAuthentication();
+    }
+
+    private destroy() {
```
luisa-beerboom, line 86:
```suggestion
    private destroy(): void {
```

client/src/app/worker/http/http-stream.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    private _receivedData: Promise<void>;
+    private _receivedDataResolver: CallableFunction;
+    public get receivedData(): Promise<void> {
+        return this._receivedData;
+    }
+
+    public get failedConnects(): number {
+        return this.failedCounter;
+    }
```
luisa-beerboom, line 33: Sort declarations by visibility

client/src/app/worker/http/http-subscription-polling.ts
```diff
@@ -0,0 +1,130 @@
...
+                this.reopen = false;
+            }
+        }
+
+        this.abortCtrl = undefined;
+
+        if (this.abortResolver) {
+            this.abortResolver();
+        }
+    }
```
luisa-beerboom, line 113: This method is too long, it should be possible to split it up to improve readability

client/src/app/worker/http/http-subscription-sse.ts
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+
+        this.abortCtrl = undefined;
+        if (this.abortResolver) {
+            this.abortResolver();
+        }
+
+        this._active = false;
+    }
```
luisa-beerboom, line 96: Too long, please make this shorter

luisa-beerboom (changes requested):
Beside the other things, I saw multiple constructors without visibility modifiers, please change that

bastianjoel:
Using accessibility modifiers for constructors is not consistently done. I don't like introducing this rule right now with this PR.

In the future we could think about adding an eslint rule https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md

luisa-beerboom:
> Using accessibility modifiers for constructors is not consistently done in the client. I don't like introducing this rule right now with this PR.

It is not an introduction, it is overwhelmingly done in the client for the simple reason that it _is_ an established convention to put visibility modifiers on everything, the exceptions where it is inconsistent are merely instances that were not caught during the code review.

luisa-beerboom (approved):
Or fix this in another PR

bastianjoel:
@emanuelschuetze @MSoeb @Elblinator @rrenkert Maybe it is not completely obvious by the title and description but this PR includes the longpolling implementation and is finished since last week. 
If anybody is feeling qualified to test this feel free to add a review. 

The [performance tool](https://github.com/OpenSlides/openslides-performance) has a new tool `broken-proxy` to test this.

rrenkert (changes requested):
Works as expected.

But the warning sign is not centered:

image

bastianjoel:
>:(

bastianjoel:
This is a problem with the banner service. Offline mode and out of sync have the same problem. For that reason I think this is totally unrelated to this PR. Please open another issue.

Elblinator:
I created a new Issue 
see #3506

PR #3277, opened 2024-02-13
Allow meeting import only for superadmins

### LINKED ISSUE #3274: Allow `meeting.import` only for superadmins
see https://github.com/OpenSlides/openslides-backend/issues/2207

reiterl:
Resolve #3274


Elblinator (changes requested):
in a committee-list if a user is a only committee admin then the three dot menu (upper right corner) is still visible and opens an empty menu
please remove the three dot menu

reiterl:
Ahh, missed that moth... Okay, You need to be at least Orga Manager to see the menu now.

bastianjoel (approved):

PR #3278, opened 2024-02-13
Add min distance between speaking card and left autopilot side

### LINKED ISSUE #3241: Autopilot prevent speaking times card and left side from autopilot to be too close
**Current appearance:**
Screenshot_20240202_135717

**Wanted appearance:**
there should be a margin/padding/distance of min 20 px in between the two cards

Elblinator:
closes #3241


MSoeb (approved):
Tried it with different browsers sizes and scalings. Can't reproduce the bug anymore. Looks good.

PR #3283, opened 2024-02-13
Fix speaking times change detection

### LINKED ISSUE #3152: Fix too frequent updates in mat-card
**Current behaviour:** 
With the dev-tools, if you investigate the speaking times mat-card p "speaking times", then you can observe that the p element with in it is being updated once every second ~

**Wanted behaviour:**
The title of the mat-card should not be updated every second

bastianjoel:
resolves #3152


Elblinator (changes requested):
if the "more_vert" menu from a single structure level is opened The following Error is thrown in the console:
```
TzpeError: Cannot read properties of undefined (reading 'countdown') at SpeakingtimesComponent.isInOvertimeAndNotSpeaking speaking-times.component.ts
```

Elblinator (approved):

PR #3298, opened 2024-02-16
Implement datetimepicker for motion workflow

### LINKED ISSUE #3148: Make motion submission date editable
needs https://github.com/OpenSlides/openslides-backend/issues/2112 

**Current behaviour:**
currently the submission date is automatically created by the program

**Wanted behaviour:**
- This submission date should be editable by participants with the permission motion.can_manage_meta_data.
- The submission date needs the edit icon next to the string 
- if the submission date is being edited a date and time picker needs to open.

### LINKED ISSUE #3248: Connection to ICC service gets closed after 30min

luisa-beerboom:
Closes #3148 
Based on OpenSlides/openslides-backend#2194

- [x] Implement clear option
- [x] Make picker recognize preset value
- [x] Fix bug where the material datepicker styles (that the datetimepicker uses) won't be loaded until a datepicker was clicked (can be solved with the workaround from the solution of [this stackoverflow question](https://stackoverflow.com/questions/77015161/why-my-angular-ngx-mat-datetime-pickers-and-mat-datepickers-css-clash-if-they) but it _is_ very ugly)


Elblinator (changes requested):
If the submission date is changed the the following snackbar error is shown: 
``` 
Error: data must not contain {'workflow_timestamp'} properties
```

rend request is: 
{
    "action": "motion.update",
    "data": [
        {
            "id": 5,
            "workflow_timestamp": 1708080137
        }
    ]
}

Is it possible to show not only the date but also the time while the edit submission date option is activated but the datepicker is not open: 
date-picker

client/package.json
```diff
@@ -37,6 +37,7 @@
     "get-available-languages": "npm run ts -- cli/get-available-languages.ts"
   },
   "dependencies": {
+    "@angular-material-components/datetime-picker": "^16.0.1",
```
bastianjoel, line 40: It seems like this library is not maintained and currently not compatible with angular 17.

bastianjoel (changes requested):

client/package.json
luisa-beerboom: Seems like I'll have to do a custom implementation with the `ngx-material-timepicker`

bastianjoel (approved):

Elblinator (approved):
function

PR #3300, opened 2024-02-16
Speaking times: rm projection options

### LINKED ISSUE #3295: Speaking times widget: remove projection options
**Description:** The speaking times widget includes a three-dot-menu with the options to project content of it. This breaks the general logic of the projection options. A Projection should not be possible from the autopilot. 

image

**What should happen:**
The three-dot-menu with the projection options should be deleted from the speaking times widget. The projection (activate/deactivate) of this content should only be possible from the Projector menu. With this change, the projection of the content will have the same logic like other speaking visualization options.

reiterl:
Resolve #3295


Elblinator (changes requested):
please re-add the option to project the structure levels in the list-of-speaker detail view
The project option should only be removed from the autopilot

speaking times

reiterl:
Closed, because of changed requirements and a new pr.

PR #3305, opened 2024-02-16
Add hint for counting up countdown

### LINKED ISSUE #3076: Add a count up timer option
Currently, we are using a countdown timer of 0 minutes to mimic a count up timer (stopwatch). Some of our members have expressed confusion about why the timer is negative and red, so we would like a upwards counting timer.

I would like an option that allows the timer to count upwards rather than down. This would act as a stopwatch instead of a timer for timekeeping purposes.

**Describe alternatives you've considered**
Mentioned above.

**Additional context**
None

### LINKED ISSUE #3247: Add hint for counting up timers
**Current behaviour:**
Thanks to the new features it is now possible to create timer which count up and not down

**Reproduction:**
1. navigate to a projector
2. set a countdown to zero seconds
3. start countdown
4. the countdown will count up

**Wanted behaviour:**
The countdowns should have a hint that they can indeed count up if they are set to zero

Elblinator:
closes #3247 
closes #3076

### DIFF
--- a/client/src/app/domain/models/projector/projection-default.ts
+++ b/client/src/app/domain/models/projector/projection-default.ts
@@ -32,7 +32,7 @@ export const PROJECTIONDEFAULT_VERBOSE: { [key in ProjectiondefaultKey]: string 
     assignment: _(`Elections`),
     mediafile: _(`Files`),
     projectorMessage: _(`Messages`),
-    projectorCountdown: _(`Countdowns`),
+    projectorCountdown: _(`Countdowns/Timers`),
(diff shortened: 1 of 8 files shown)


bastianjoel (approved):

MSoeb (changes requested):
The Countdown/Timers dialog needs an improvement for the box infos. The box field texts should be shrinked to:

- Title
- Description
- Time

Also the Hint 'If set to zero this will be a timer' should be changed to 'If the value is set to 0, the time counts up as a timer.'

MSoeb (approved):

PR #3306, opened 2024-02-19
Add no speech/no structure level filters

### LINKED ISSUE #3292: Add filter options for contribution-list
**Current behaviour:**
You can currently filter 'speech type' and 'structure level' and every active option of these two types. But the options to search for a not set structure level or speech type is missing, even though you can have contributions without a set structure level or speech type

**Wanted behaviour:**
The options to filter for 'no structure level' or 'no speech type should be possible

reiterl:
Resolve #3292


Elblinator (changes requested):
Please capitalise the 'n' from 'no [...]'

reiterl:
I capitalized labels of the two new filters.

Elblinator (approved):

PR #3307, opened 2024-02-19
Add new option to sort contribution list and update sort default

### LINKED ISSUE #3294: Contributions menu: Sort function 'speak end time' needs to be reworked
**Description:** Currently it is not possible to sort the speakers in the contributions list according to the last speakers list. The last speakers list sort the speeches by start time. The sort function doesn't include this. But it has the option the sort by speak end time. This is not helpful at all and needs to be reworked. The sort function 'speak end time' tracks the wrong desired data and is also incoherent to the export column speech start time in the csv export file. 

**What should happen:**
- [ ] The sort function 'speak end time' should be redesigned to track the speak start time. 
- [ ] Speak end time should than be renamend to Speak start time. 

In general the sort function should works in the same logic like the last speakers list. So please take care of for example interposed questions.

reiterl:
Resolve #3294

Add a sorting after begin_time option and set that to be the default sorting.


Elblinator (changes requested):
It was a requirement to not only make the 'start_time' an available filter option, 
But the end_time should be removed 

Please remove the end_time as a filter option

reiterl:
Ah, okay, I have removed the end_time sorting part.

bastianjoel (approved):

MSoeb (approved):

Elblinator (approved):

PR #3308, opened 2024-02-19
Update moderation notes visibility if no projection

### LINKED ISSUE #3299: Moderation note: visible without active projection
**Description:** The moderation note widget is visible in the Autopilot without an active projection if a user has the permission 'Can manage moderation notes'. 

**What should happen:**
The moderation notes widget should only be visible, if content is projected in the Autopilot.

reiterl:
Resolve #3299 

Test if active projection and only show moderation notes then.


Elblinator (approved):
function and code

PR #3320, opened 2024-02-20
Improve service worker app updates

### LINKED ISSUE #3291: Fix loading problems if browsers is closed/opened
**Current behaviour:**
And if I start my Pc and then open/ call the url of the instance (also in productive Setup) then there is a loading symbol and 'Loading data. Please wait ...'

**Wanted behaviour:**
Opening openslides for the first time of the day should not send you into a indefinitely spinning wheel

removed login-part of issue, because it was due not code-related things

bastianjoel:
resolves #3291

### DIFF
--- a/client/ngsw-config.json
+++ b/client/ngsw-config.json
@@ -5,7 +5,7 @@
         {
             "name": "app",
             "installMode": "prefetch",
-            "version": 1,
+            "version": 2,
             "resources": {
                 "files": [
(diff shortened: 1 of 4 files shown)


peb-adr (approved):
Tested together with @bastianjoel on prod build. Worked as expected.
Changes also look sensible to me.

rrenkert (approved):

PR #3323, opened 2024-02-20
Increase action worker timeout

jsangmeister:


bastianjoel:
In `action-worker-watch.service.spec.ts` one of the tests needs to be adjusted.

jsangmeister:
> In `action-worker-watch.service.spec.ts` one of the tests needs to be adjusted.

Yes, I figured, but I'm against merging this anyway. It is a quick fix for instances where the imports take extraordinarily long in the datastore, since during this time the backend cannot write action worker updates. This goes hand in hand with increasing the datastore worker timeout to even allow such long-running queries, which should also not be done too often.

jsangmeister:
obsolete with https://github.com/OpenSlides/openslides-datastore-service/pull/316

PR #3335, opened 2024-02-22
Update committee export (username, labels, timestamp)

### LINKED ISSUE #3273: Committee export
**Current behaviour:**
If a committee is exported then some field are missing, in the wrong format or otherwise not correct

**Wrong format in the CSV export file:**
if committees are exported then: 
- [x]  the managers are exported with their first and last name, but the import needs their username
- [x]  the meeting_start/end_time is exported as YYYYMMDD, but the import needs the date to be YYYY-MM-DD
- [ ]  the meeting admins are not exported

reiterl:
Resolve #3273 

I worked on the committee export, updated the fields to match the backend fields, use usernames, update the date/timestamp stuff.
I have included an empty meeting_admins. I have in this context no idea how to get the meeting_admins. So excluded loading that.


Elblinator (approved):
approved function with the note that the issue should not be closed bc the last part with the meeting admin is missing

bastianjoel (approved):

Issue #3253, opened 2024-02-03
Improve meeting settings


jsangmeister:
- [x] Add option to add plain text segment with the following parameters:
  - text
  - button text and target (optional)
- [x] Add option to indent specific settings
  - either by directly defining the indentation level or by nesting the settings in the definition
- [x] Add option to disable settings depending on other setting(s)

PR #3357, opened 2024-02-26
Enhance meeting settings

jsangmeister:
resolves https://github.com/OpenSlides/openslides-client/issues/3253


bastianjoel (approved):

jsangmeister:
Implemenented the following changes:
- moved the help text in front of the button
- made help text optional to allow standalone buttons
- added link type enum: use `Meeting` for links inside the meeting (e.g. `motions`), `Organization` for links inside the organization (e.g. `designs`) or `External` for external links (e.g. `https://github.com`)
- fixed the save button in combination with the `disabled` attribute
- fixed the disablement of PoO categories

I did not implement the transitive disable feature as this would be too much work for something which can be achieved explicitly: If the order is A -> B -> C (meaning B should only be active if A is true and C only if B is true), just add the requirement that A must be true to C as well, so that the function looks like this:
```
disabled: settings => !settings.A || !settings.B
```
This way, both other settings must be true. It makes the definition file a little bit more verbose, but saves code and work.

bastianjoel (approved):

jsangmeister:
@emanuelschuetze ping

emanuelschuetze (changes requested):
I tried this one for my A->B->C example:
```
disable: settings => !settings.list_of_speakers_enable_point_of_order_categories || !settings.list_of_speakers_enable_point_of_order_speakers
```
If A was disabled and saved you get this error:
`You cannot enable point of order categories without enabling point of order speakers.`

jsangmeister:
@emanuelschuetze should be fixed now

jsangmeister:
@emanuelschuetze I decided to remove the check in the backend instead of implementing a half-hacky solution here. I checked all places in the client and backend and this should not affect the current behaviour. https://github.com/OpenSlides/openslides-backend/pull/2326 is now needed to test this.

emanuelschuetze (approved):
Tested

bastianjoel (approved):

PR #3371, opened 2024-02-28
Prevent second scrollbar by rm wrong style

### LINKED ISSUE #3356: Prevent two scrollbars -> impressum count speakers
**Current behaviour:**
In the Impressum if 11 or more contributions are counted then two scrollbars appear next to each other

**Reproduction:**
1. Start and stop at least 11 contributions 
2. Navigate to the impressum count 'completed requests to speak'
3. you should see the double scrollbars

**Wanted behaviour:**
Only one scrollbar should be displayed

reiterl:
Resolve #3356

### DIFF
--- a/client/src/app/site/pages/meetings/pages/home/pages/meeting-info/components/user-statistics/user-statistics.component.html
+++ b/client/src/app/site/pages/meetings/pages/home/pages/meeting-info/components/user-statistics/user-statistics.component.html
@@ -25,7 +25,6 @@
         </tr>
     </table>
     <os-list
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
function

bastianjoel (commented):
I am not sure if it is a good idea to remove the height there. Was this tested with sufficient entries @Elblinator?

Elblinator (commented):
I tested it with 11, 47, ~75 and 110 entries, 
I could only provoke onle scrollbar independent from height and wdith

bastianjoel (approved):

PR #3372, opened 2024-02-28
Change motion deletion warning

### LINKED ISSUE #3361: Remove amendments if motion is deleted
**Currently:**
If a motion with amendments is deleted then the amendments are left behind and practically unusable.

**How to reproduce:**
1. create a motion
2. create an amendments for this motion
3. delete this motion
4. a dialog opens which warns you that some amendments will lose their parent motion

**Wanted:**
When a motion which has amendments is deleted then the amendments should be deleted together with the motion
The warning should be adjusted accordingly

reiterl:
Resolve #3361 

Waiting for the backend issue


reiterl:
Is already fixed by @luisa-beerboom

PR #3373, opened 2024-02-28
Change motion delete warnings

### LINKED ISSUE #3361: Remove amendments if motion is deleted

luisa-beerboom:
Closes #3361


reiterl (approved):
LGTM
It should wait on the ba pr.

Elblinator (commented):
I'll test this again after the CR in the backend is implemented

https://github.com/OpenSlides/openslides-backend/pull/2253

PR #3376, opened 2024-02-28
Remove speakingTime errors when toggling los_can_see

### LINKED ISSUE #3304: Strange permissions interaction with error message
**Description:** There is a strange interaction between the permissions "can see moderation notes", "Can see internal items and time scheduling of agenda", "Can put oneself on the list of speakers" and "can see list of speakers". If 'can see list of speakers' is deactivated, an error appears in the Autopilot with the message "Error talking to autoupdate service".

**Reproduction:**
1. open a meeting
2. open agenda, create a topic and project it. 
3. open participants, create a user and put it in group delegates
4. open participants > groups and grant the group delegates the following Agenda permissions: Cann see agenda, Can see internal items and time scheduling of agenda, Can put oneself on the list of speakers and Can see moderation notes
5. open Autopilot as Delegate user. Now revoke the permission Can see list of speakers. The error will appear on the delegates autopilot

**What should happen:**
Revoking the permission Can see list of speakers should not lead to an error message.

luisa-beerboom:
Closes #3304

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/speaking-times/speaking-times.component.ts
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/speaking-times/speaking-times.component.ts
@@ -154,22 +154,24 @@ export class SpeakingTimesComponent implements OnDestroy {
     }
 
     private updateSpeakingTime(speakingTime: ViewStructureLevelListOfSpeakers) {
-        if (speakingTime.isInactive) {
-            this.structureLevels.delete(speakingTime.id);
-        } else {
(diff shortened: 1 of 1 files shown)


jsangmeister (commented):
What happens if you remove `can_see` for a user which has the component open, then remove a structure level by removing its only speaker, then give the permission back to the user? Does the removed structure level get correctly remvoed for the user?

luisa-beerboom:
Yes.

In the first place this would break nothing that wasn't broken already

jsangmeister (approved):

MSoeb (approved):

Issue #3362, opened 2024-02-26
Add free text motion submitter


Elblinator:
**Currently:**
It is currently only possible to add actual participants as submitters 

**Wanted:**
It should always be possible to add a submitter as plain text. 
There should be a free text field under 'Select or search new participant' 
![14](https://github.com/OpenSlides/openslides-client/assets/69210919/fcf72cc4-9b82-4650-9740-f946afa11874)

PR #3377, opened 2024-02-29
Add free text submitter field

jsangmeister:
resolves https://github.com/OpenSlides/openslides-client/issues/3362

needs https://github.com/OpenSlides/openslides-autoupdate-service/pull/864 and https://github.com/OpenSlides/openslides-backend/pull/2258

### DIFF
--- a/client/src/app/domain/models/motions/motion.ts
+++ b/client/src/app/domain/models/motions/motion.ts
@@ -44,6 +44,7 @@ export class Motion extends BaseModel<Motion> implements MotionFormattingReprese
     public state_extension!: string;
     public recommendation_extension!: string;
     public sort_weight!: number;
+    public additional_submitter!: string;
     /**
      * Client-calculated field: The tree_weight indicates the position of a motion in a list of
      * motions in regard to the call list.
@@ -117,6 +118,7 @@ export class Motion extends BaseModel<Motion> implements MotionFormattingReprese
         `workflow_timestamp`,
         `start_line_number`,
         `forwarded`,
+        `additional_submitter`,
         `lead_motion_id`,
         `amendment_ids`,
         `sort_parent_id`,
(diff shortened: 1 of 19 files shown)


bastianjoel (approved):

Elblinator (changes requested):
The additional Submitter is not exported in the pdf

motion-list: 
- additional submitter is not displayed in the motion-list view
- If you search in the  motion-list for an additional submitter, you need to find the motion with the submitter
- the additional submitter needs to be sorted in the same manner as the other submitters

The submitter also needs to be searchable via global search

If the submitters are projected there is one too many spaces

Change 'additional submitter(s)' string to 'Extension'

jsangmeister:
- [x] added to CSV and PDF export
- [x] added to motion list view
- [x] added to filterable properties in the list view
- [x] added to `search.yml` for global search, see https://github.com/OpenSlides/openslides-meta/pull/38
  - Note: I did not create a PR for this as the search service will get updated automatically after the meta PR is merged
- [x] changed input label to "Extension"

> * the additional submitter needs to be sorted in the same manner as the other submitters

What do you mean by that? I currently simply append it to the normal submitters.

> If the submitters are projected there is one too many spaces

Cannot reproduce.

jsangmeister:
Included the new submitter field when sorting the motion list by submitter.

bastianjoel (approved):

Elblinator (changes requested):
if a motion has only an extension as a submitter then the extension string is not displayed in the motion-list view

Elblinator:
>>  the additional submitter needs to be sorted in the same manner as the other submitters

> What do you mean by that? I currently simply append it to the normal submitters.

already fixed that part

>> If the submitters are projected there is one too many spaces

> Cannot reproduce.

Ignore that part, there might have been a loading bug, cannot reproduce anymore

Elblinator (approved):
function

Issue #3239, opened 2024-02-01
Change filter/sort/search bar 


Elblinator:
As discussed in the dev-meeting, the behaviour and design from the filter, sort and search bar should change:

- add icons for filter and sort 
- change the icon for search 
- in mobile only the icons should be visible
- in mobile if an option (filter/sort) is activated the the name should be displayed behind this icon
- in mobile if search is activated the icon for filter and sort should vanish
- filter and sort (if you actively select stuff) should fill the whole screen

Mock-ups will follow from @Eich-hoernchen

Eich-hoernchen:
![001-Desktop_add_Icons](https://github.com/OpenSlides/openslides-client/assets/60174105/dfe1c53c-e48d-4570-a719-d360ab0c6d49)
add icons for filter and sort

![002-Filter_Sort_Search-Icon](https://github.com/OpenSlides/openslides-client/assets/60174105/b3bc2853-c712-44f0-9e96-1a5c911c877c)
in mobile only the icons should be visible, change the icon for search

![003-Sort](https://github.com/OpenSlides/openslides-client/assets/60174105/c222164a-2fc2-44b7-b840-101e05ec8855)
sort-list with the new icon -refresh- to reset to default sort

![004-Filter](https://github.com/OpenSlides/openslides-client/assets/60174105/1df625bf-ba8e-4bb2-a05b-a0c5c159cc03)
in mobile if an option (filter/sort) is activated the the name should be displayed. The selected filters are shown in the bar. Selected Filters can be slide from right to left and deleted by touching the cross.

![005-Search](https://github.com/OpenSlides/openslides-client/assets/60174105/333909c6-c8a7-4663-ac3b-350b16f2a5fb)
by touching the new search icon a search field opens. The filter icon turns smaller, so the count of the filters will be visible.

Elblinator:
icon names: 
sort: sort
filter : filter_list
search: manage_search

Elblinator:
The description of the Issue is updated according to the answers

PR #3388, opened 2024-03-04
change filter sort search (part 1)

reiterl:
This is part 1 of #3239 

This should be evalutated. It includes:

- Add icons
- an update of the mobile checker
- only icons in mobile view.

### DIFF
--- a/client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
+++ b/client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
@@ -170,7 +170,7 @@ export class SortFilterBarComponent<V extends Identifiable> implements OnInit {
 
     private _searchField = ``;
 
-    public isMobile = false;
+    private _isMobile = false;
 
     public constructor(
         protected translate: TranslateService,
@@ -181,7 +181,7 @@ export class SortFilterBarComponent<V extends Identifiable> implements OnInit {
 
     public ngOnInit() {
         this.vp.isMobileSubject.subscribe(v => {
-            this.isMobile = v;
+            this._isMobile = v;
             this.cd.markForCheck();
         });
     }
@@ -254,6 +254,10 @@ export class SortFilterBarComponent<V extends Identifiable> implements OnInit {
         }
         const itemProperty = option.property as string;
         return itemProperty.charAt(0).toUpperCase() + itemProperty.slice(1);
+    }
+
+    public get isMobile(): boolean {
+        return this._isMobile;
     }
 
     @HostListener(`document:keydown`, [`$event`]) public onKeyDown(event: KeyboardEvent): void {
--- a/client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
+++ b/client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
@@ -1,9 +1,11 @@
 import {
     ChangeDetectionStrategy,
+    ChangeDetectorRef,
     Component,
     EventEmitter,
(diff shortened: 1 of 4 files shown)


client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
```diff
@@ -244,6 +261,19 @@ export class SortFilterBarComponent<V extends Identifiable> {
         return itemProperty.charAt(0).toUpperCase() + itemProperty.slice(1);
     }
 
+    public get isMobile(): boolean {
+        return this._isMobile;
```
bastianjoel, line 265: You could just pass the subject to the template and use the `async` pipe.

bastianjoel (changes requested):

client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
reiterl: Worked on this. I replaced the _isMobile code with the use of observers.

Elblinator (changes requested):
- the font-weight from 'SORT' and 'FILTER' should be 400
- the new search icon should always have the same colour as the other icons 
- in mobile view:
  - the new search icon should have the same padding as the other two icons
  - the search element (where you can write into) should change into only the icon if you press out of the search element
  - the search element does not appear if the screen width is > 501px

client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
```diff
@@ -174,6 +177,14 @@ export class SortFilterBarComponent<V extends Identifiable> {
...
     ) {}
 
+    public ngOnInit() {
+        this.vp.isMobileSubject.subscribe(v => {
+            if (v) {
+                this.searchEdit = false;
+            }
+        });
+    }
+
```
bastianjoel, line 187: The main intention of my last cr was that I think this subscription could be fully deleted with the proposed change. 

If you think that this is not the case you still need to close the subscription when the component is destroyed.

bastianjoel (changes requested):

client/src/app/ui/modules/list/components/sort-filter-bar/sort-filter-bar.component.ts
reiterl: Done, I have add an unsubscribe.

reiterl:
"the search element (where you can write into) should change into only the icon if you press out of the search element" I don't understand this point.

Elblinator:
> "the search element (where you can write into) should change into only the icon if you press out of the search element" I don't understand this point.

in mobile view: 
- First you only see the icon (correct) 
- If you press/activate the icon then the field to type in text appears (correct)
- If you now press on the filters/sort option, or just somewhere where you do not change pages but change the focus from the search field to something else, then the searc hfiled stays big (wrong)
- the search field shrink back to only the icon if it is out of focus

reiterl:
I have updated this component:
* font-weigth of sort and Filter.
* icon color of search icons
* search icon padding (now all three are the same)
* rm the display none, which causes the search input field to disappear
* rm the focus large mode, which causes the don't shrink issue, and problems in phone mode.

Elblinator (approved):
Approve function and visuals and retract the last change request 
> "the search element (where you can write into) should change into only the icon if you press out of the search element" I don't understand this point.

This is part of issue #3239 but can be considered as a different point from the list  (I'll update the Issue description)

bastianjoel (approved):

PR #3395, opened 2024-03-06
Prevent certain speech state changes

### LINKED ISSUE #3358: Prevent change of speech state for current_speaker
**Current behaviour:**
In a LoS if a speaker already started to speak then the speech state can be changed to any other speech state

**Reproduction:**
1. activate pro/contra speech state, enable star icon, activate intervention, activate countdowns for structure levels, enable point of orders for other participants
2. navigate to any LoS
3. put someone on the who has a structure level
4. select the structure level
5. start the participant
6. now change the normal contribution to any of the above
7. the countdown for the structure level will continue to count down even if the speech state is changed to a state without/with it's own countdown

**Wanted behaviour:**
- the speech states PoO, Pro, Contra, star should be exclusive to each other, only one of them should be active at a time 

If the currently speaking person speaks for/as....
- person speaks with no chosen speech  state/PoO/Pro/Contra/star, then the different speech states (no chosen speech  state/PoO/Pro/Contra/star) should still be selectable/deselectable
- If the person speaks as intervention, then nothing should be changeable 
- If the person speaks with a structure level, then only pro contra and star should be selectable/deselectable

jsangmeister:
fixes #3358

### DIFF
--- a/client/src/app/domain/models/speakers/speech-state.ts
+++ b/client/src/app/domain/models/speakers/speech-state.ts
@@ -5,3 +5,5 @@ export enum SpeechState {
     INTERPOSED_QUESTION = `interposed_question`,
     INTERVENTION = `intervention`
 }
+
+export const SPECIAL_SPEECH_STATES = [SpeechState.INTERPOSED_QUESTION, SpeechState.INTERVENTION];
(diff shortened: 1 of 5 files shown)


Elblinator (changes requested):
Please hide the mat-divider above clear if the intervention is startet (only clear is available as option but the mat-divider ist visibe)

While the current speaker is speaking with an PoO:
- Please make Forspeech, Contra speech, and contribution visible and directly activatable
- please change the string 'Point of order' to 'Remove Point of order' while the point of order is selected (also while the person is not yet speaking

Issue #2934, opened 2023-10-20
Angular material mdc migration


bastianjoel:
With angular 15 there was an update for the angular material components. 

Currently we are using `MatLegacy...` modules which should be replaced in the near future with the new MDC components. 

There is a migration tool which automatically replaces the used modules and updates some other things. However this needs testing and manual intervention in at least most cases. For that the migration tool has a feature to do this migration only for specific modules (which has the drawback that the main `styles.scss` will not be modified). 

A migration guide can be found here: https://material.angular.io/guide/mdc-migration

Components to migrate
- [x] Autocomplete
- [ ] Button #3703 
- [x] Card #3487
- [ ] Checkbox #3696 
- [ ] Chips
- [x] Dialog #3397 
- [x] Form Field #3002
- [x] Input #3002
- [ ] List
- [ ] Menu
- [x] Option / Optgroup #3002
- [x] Paginator
- [x] Progress Bar #3097
- [x] Progress Spinner #3097
- [x] Radio #3001
- [x] Select #3002
- [x] Slide Toggle
- [x] Slider #2998
- [x] Snack Bar #3097
- [ ] Table #3694 
- [x] Tabs #2998
- [x] Tooltip #2999

bastianjoel:
When working on this: Do not try to stick to the old appearance too much and just adapt the changes from the library. This will make our life easier especially with the upcoming material design 3 update.

PR #3397, opened 2024-03-06
Update mdc dialog

Elblinator:
part of #2934 

Changed the dialog
The appearance should have slightly changed


client/src/app/site/pages/meetings/pages/agenda/modules/topics/modules/topic-poll/components/topic-poll-dialog/topic-poll-dialog.component.html
```diff
@@ -1,79 +1,81 @@
...
+                    {{ optionsWarning | translate }}
+                </i>
+            </ng-container>
+        </div>
     </div>
 </div>
 
 <!-- Actions -->
-<div mat-dialog-actions>
+<div mat-dialog-actions class="padding-left-24 padding-top-bottom-16">
```
bastianjoel, line 78: I am not sure if we should do this. In my opinion we should just stick to the defaults from mdc as much as possible. 

In future versions the appearance could always change and adding imho unnecessary adaptions should be prevented.

bastianjoel (changes requested):

Elblinator:
The formatting does not wok for client/src/app/site/modules/global-headbar/components/account-dialog/account-dialog.component.scss if the padding:0 is removed 
I would really prefer this padding here

bastianjoel (approved):

MSoeb (approved):

PR #3400, opened 2024-03-06
Prepend pub key size in cipher text

### LINKED ISSUE #3398: Support other curves in vote-decrypt
As described in OpenSlides/vote-decrypt/pull/26 we need to add the pub key size at the beginning of the ciphertext.

rrenkert:
closes #3398


luisa-beerboom (approved):
Fix linting please

PR #3402, opened 2024-03-07
Change visibility of projection icon

### LINKED ISSUE #2981: Change visibility of projection-icon
**Current behaviour:**
If a participant is allowed to (ONLY) see the projectors then everything which is projected has a projector icon next to it:

**Reproduction:**
in a meeting: 
1. have at least two projectors
2. create two agenda topics
3. project the first topic onto the reference projector
4. project the other topic onto the other projector
5. login as a delegate (make sure the delegates cannot manage the projectors)
6. navigate onto the agenda 
7. you should see something similar to the screenshot

**Wanted behaviour:**
A participant should only be able to see every projection icon if they are allowed to **manage** the projectors. 
If a participant is only allowed to see the projections, then they should only see the item which is projected by the reference projector

luisa-beerboom:
Closes #2981


bastianjoel (approved):

Elblinator (approved):
function

PR #3403, opened 2024-03-07
Fix for meeting settings default groups

### LINKED ISSUE #3382: Fix resetting 'Default groups with voting rights'
**Current behaviour:**
If the 'Default voting type' or 'Add to agenda' is changed then every entry from 'Default groups with voting rights' is deselected

**Reproduction:**
1. navigate to Settings -> continue to Agenda/Motions/Elections
2. select anything in 'Default groups with voting rights'
3. save the change
4. change 'Default voting type' (in Motions and Elections) or 'Add to agenda' (in Agenda)
5. save the change
6. the old default groups are still displayed
7. reload page
8. the default groups are empty

**Example payload from election:**
```
[
    {
        "action": "meeting.update",
        "data": [
            {
                "assignment_poll_default_group_ids": null,
                "assignment_poll_default_type": "pseudoanonymous",
                "id": 2
            }
        ]
    }
]
```

**Wanted behaviour:**
The 'Default groups with voting rights' should not ne automatically changed. They should stay the same even if other settings are changed

reiterl:
Resolve #3382

### DIFF
--- a/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
+++ b/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
@@ -233,7 +233,10 @@ export class MeetingSettingsGroupDetailFieldComponent extends BaseComponent impl
                 })
             )
             .subscribe(form => {
-                if (this._comparedForm || JSON.stringify(form.value) !== JSON.stringify(this._firstValue)) {
+                if (
+                    this._comparedForm ||
+                    (JSON.stringify(form.value) !== JSON.stringify(this._firstValue) && this.setting.type !== `groups`)
+                ) {
                     this.onChange(form.value);
                 }
             });


Elblinator (approved):
function

(I tried to test several case,, and could not find any bug)

bastianjoel (commented):
I am not sure about this. The `_comparedForm` part looks very odd to me. Also the distingUntilChanged check stands in contradiction with the check within the subscription. 

@luisa-beerboom introduced the `_comparedForm` in #2103 where I was not involved. Maybe she can give better input on this.

luisa-beerboom (changes requested):
This is just going to cause `onChange` to not be called with `groups` type fields.

At this point lines 229 to 231 is pointless and the `groups` section of `onChange` will not even get called. Due to the latter this component will never send an update event, leading me to question, if changes made to group fields will even get passed on to the backend.

Please check/fix that.

reiterl:
As far as understand the code, the `_compareForm` flag is set, after the first compare. So the onChange code is used then, because `_compareFlag` is true. @luisa-beerboom  can you verify this?

luisa-beerboom:
Ah right, my bad.
I think this looks fine then.

luisa-beerboom (approved):

PR #3411, opened 2024-03-08
Add editors and working group speakers to pdf export

### LINKED ISSUE #3303: Motion - new meta data: PDF export misses new data
**Description:** The new meta data editor and working groups are currently not exportable in the pdf file. This is an information loss. 

**What should happen:**
The new meta data should be exported in the pdf file. This should affect at least the single export of a motion and amendment. Furthermore it should be included in the call lsit export and in the export via multi selection. The furthermore aspects needs to be discussed what is the best option.

reiterl:
Resolve #3303


Elblinator (approved):
function

only the single motion export was fixed (as described in the issue, I'll create anew issue for the furthermore part of the issue)

bastianjoel (approved):

PR #3439, opened 2024-03-14
Fix projection of LoS

### LINKED ISSUE #3410: Fix projection of LoS
**Current behaviour:**
If the LoS is projected onto a projector there is an interposed question on the LoS the LoS is wrongly displayed.
Depending on whether or not the interposed question is started/paused/stopped the displayed LoS is wrong in different ways

- [x] The 'Slide' projection should be fixed as described
- [x] The 'Overlay' projection should be fixed as described
- [x] the additional missing icons should be added

Examples only portrait the Slide bug, but the bug also entangles the Overlay, there is one screenshot showing the Slide and Overlay projection at the same time at the end

**A) LoS and projected LoS _before_ interrposed question is _started_:**

Problems: 
- the numbering is wrong (interposed questions should not be numbered)
- the interposed question is not indented (it should be indented)
- the interposed question should be directly under their parent and the gap should be between the last interposed question and the next parent speaker

**B) LoS and projected LoS _when_ interrposed question is _started_:**

Problems:
- Main speaker (here B) vanishes (B should still be visible, but without the mic)
- Interposed question should stay where it was before and have a mic

**C) LoS and projected LoS _when_ interrposed question is _paused_ :**
Same screenshot as  in the projection B)
Problems:
- if everyone is paused, then no one should have a mic

**D) LoS and projected LoS _when_ parent is _started again_:**

Problems:
- parent should be displayed 
- parent should have the mic again

**Overlay projection next to Slide projection:**

**Additional wanted behaviour:**

- the icons from interposed question intervention and the structure level are missing

luisa-beerboom:
Closes #3410 

Needs OpenSlides/openslides-autoupdate-service#890

### DIFF
--- a/client/src/app/site/pages/meetings/modules/projector/modules/slides/components/list-of-speakers/modules/common-list-of-speakers-slide/common-list-of-speakers-slide-data.ts
+++ b/client/src/app/site/pages/meetings/modules/projector/modules/slides/components/list-of-speakers/modules/common-list-of-speakers-slide/common-list-of-speakers-slide-data.ts
@@ -7,6 +7,8 @@ export interface SlideSpeaker {
     speech_state: SpeechState;
     note: string;
     point_of_order: boolean;
+    begin_time: number;
+    pause_time: number;
 }
 
 interface TitleInformation extends TitleInformationWithAgendaItem {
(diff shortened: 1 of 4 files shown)


Elblinator (changes requested):
Sorry, I overlooked another issue with the projected LoS:
Please let only a selected structure level be displayed. 
Example: If Kim has two structure levels then both structure levels are currently projected, even if one or none is selected. 
Note: The display of all structure levels is however correct when the structure levels are not connected with the countdowns


And please change the distance from class=nextSpeakers back to margin-top: 20px !important; 
Your change with the interposed-question class is enough :)

bastianjoel (approved):

Elblinator (approved):

Issue #3085, opened 2023-12-10
Deactivate statute feature


emanuelschuetze:
The feature statute amendments can be removed. It's never used any more.
Recommended way to create statute amendments: Copy full statute in a normal motion (with special category). Now, user can create (normal) amendments to this statute/motion. You get the full support for diff/amendent/pdf/projection. No special function for statute amendments. So please remove old code of statute amendments. BTW: the feature currrently not work (`NullInjectorError` after try to open list view).

jsangmeister:
I assume the same holds true for the backend: https://github.com/OpenSlides/openslides-backend/issues/2053. Does this only pertain to statute _amendments_ or statute _paragraphs_ as well? Meaning, is https://github.com/OpenSlides/openslides-backend/issues/1715 still needed or can that be closed and all statute paragraphs related features be removed as well?

MSoeb:
Discussed clean up process with @rrenkert. Clean up should be done in two steps:
1. Deactivate the check box in the settings and make it invisible. -> Feature can't be used anymore
2. After 2 weeks second step starts with clean up of everything that is associated with the feature.

emanuelschuetze:
As discussed with @MSoeb: You can hide the statue amendment checkbox in settings now.

jsangmeister:
Checkbox is hidden now

Elblinator:
The statute amendment can be removed now

PR #3440, opened 2024-03-14
Hide statute amendments checkbox

Elblinator:
first step of #3085

### DIFF
--- a/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail/meeting-settings-group-detail.component.html
+++ b/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail/meeting-settings-group-detail.component.html
@@ -28,6 +28,7 @@
             <h3 class="accent" *ngIf="settingsGroup.subgroups.length > 1">{{ subgroup.label | translate }}</h3>
             <ng-container *ngFor="let setting of subgroup.settings">
                 <os-meeting-settings-group-detail-field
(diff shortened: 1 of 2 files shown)


emanuelschuetze (commented):
It it really the right way to hide a settings option? In earlier versions we used a single line "hide" attribute for settings. Maybe we can support this?

Elblinator:
If I understand the Code correctly this option cannot simply be called but needs new implementing. 
Given that we do not plan to hide settings in general, but just for the moment, I would prefer this temporary workaround

But I could misunderstand the code. @bastianjoel what do you think about this?

jsangmeister:
Why not simply remove the setting from the definition list? Seems cleaner than implementing this workaround.  @emanuelschuetze the `hide` attribute was probably in OS3 - in OS4, this does not exist (yet).

Elblinator:
> Why not simply remove the setting from the definition list?

Removing it in the definition-list makes it completely unusable. Other files need the definition to determine whether or not statutes are allowed

jsangmeister:
I meant from the `meeting-settings-definition.ts`. That only defines what is displayed where in which group and not which settings are generally available.

jsangmeister:
The previous suggestion did not work since the `MeetingSettingsDefinitionService` parses the definition list to determine which fields to request, which I did not know. I implemented a more general solution instead which can hide arbitrary entries, like @emanuelschuetze suggested. @Elblinator please review.

Elblinator:
Looks very nice and clean :)

emanuelschuetze (approved):

bastianjoel (approved):

PR #3472, opened 2024-03-22
Prevent line break in motion list number column

### LINKED ISSUE #3471: Motion list prevent line breaks in motion number

bastianjoel:
resolves #3471 

Now uses a static width of 29px for up to three letters and a dynamic width of up to 8*7px for up to eight digits.


emanuelschuetze (changes requested):
I don't like to move the column width to right. We have to reduce the projector button column width instead. Please crop e.g. the padding-right from projector column. We have big white space between button and number ...

PR #3488, opened 2024-03-27
Add option to customize autopilot content

### LINKED ISSUE #3430: Autopilot: Add menu to make content selectable
**Description:** Currently the Autopilot content is static. The only possible selection is to choose, what content should be displayed and in relation to this the content in the Autopilot changes. After setting this up, the content is static and all activated parts are visible. This leads to the problem, that small displays like tablets have a long scrolling bar and the user need too scroll to the desired content. Especially for administrative usage this can be quite hard to handle.

**What should happen:**
The Autopilot needs an option to choose, which parts of it should be visible on a local base; not for whole Account. 

**Possible Solution:**
New menu in Autopilot which can be reached via a three-dot-menu on the top right side (like other three-dot-menus). In the menu the user can decide, which parts should be local visible and which not. The menu shows all available parts of the Autopilot and the user can choose with a checkbox, which should be visible. 
After the selection is done, the Autopilot should rearrange the parts automatically according to the selection.
The settings should be persisted in local storage to be available across browser tabs.
Independent of the users groups or permissions all widget options should be included in the menu.

bastianjoel:
resolves #3430 

This also fixes a race condition in `base-repository.service.ts` where if one registered an sortList and directly unregisters it an exception is thrown causing content to not being displayed.

### DIFF
--- a/client/src/app/gateways/repositories/base-repository.ts
+++ b/client/src/app/gateways/repositories/base-repository.ts
@@ -434,11 +434,10 @@ export abstract class BaseRepository<V extends BaseViewModel, M extends BaseMode
             this.sortListServices[key] = sortService;
             this.pushToPipeline({
                 funct: async () => {
+                    const sortListService = this.sortListServices[key];
                     this.updateForeignBaseKeys(key);
-                    await this.sortListServices[key].hasLoaded;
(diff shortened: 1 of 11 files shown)


luisa-beerboom (approved):

luisa-beerboom:
Looks good code-wise

rrenkert (approved):

PR #3360, opened 2024-02-26
Add new user permissions

jsangmeister:
resolves #3158

AU and backend are already merged.

Issue #3158, opened 2024-01-18
Add new user permissions


jsangmeister:
see https://github.com/OpenSlides/openslides-backend/issues/2122, which must be implemented first

The new permissions must be implemented in the permission matrix. The update of participants should be made possible with only `user.can_update`.

jsangmeister:
see https://github.com/OpenSlides/openslides-client/pull/3360#issuecomment-1999542022 for the remaining TODOs

PR #3489, opened 2024-03-27
Change user name fallback

### LINKED ISSUE #3464: Change user title fallbacks to use id instead of username
With https://github.com/OpenSlides/openslides-autoupdate-service/pull/892, the restrictions are changed in such a way that the username of a user is only visible with `user.can_see_sensitive_data`. Therefore, the currently implemented fallback of displaying the username when a user does not have a first or last name will no longer work. The fallback should be changed to simply display `User {id}` instead. In real instances, almost all users have a first or last name anyway, so this will not really cause any changes for existing users.

bastianjoel:
resolves #3464

### DIFF
--- a/client/src/app/gateways/repositories/users/user-repository.service.ts
+++ b/client/src/app/gateways/repositories/users/user-repository.service.ts
@@ -35,7 +35,7 @@ export interface AssignMeetingsPayload {
 }
 
 export interface NameInformation {
-    username: string;
+    id: Id;
     first_name?: string;
     last_name?: string;
 }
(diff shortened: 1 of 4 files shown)


Elblinator (approved):
is works and looks fine 

however two questions: 
- should this PR wait for https://github.com/OpenSlides/openslides-client/issues/3158 (https://github.com/OpenSlides/openslides-client/pull/3360)
- is it really wanted that the username is not displayed at all? OR am I missunderstanding the Issue/PR?

jsangmeister:
> * should this PR wait for [Add new user permissions #3158](https://github.com/OpenSlides/openslides-client/issues/3158) ([Add new user permissions #3360](https://github.com/OpenSlides/openslides-client/pull/3360))

No, since the restriction change is already merged in the AU - therefore, the current fallback no longer works.

> * is it really wanted that the username is not displayed at all? OR am I missunderstanding the Issue/PR?

Yes, the username is now considered "sensitive data" which can only be seen with the new `user.can_see_sensitive_data` permission, which will be implemented in the client with https://github.com/OpenSlides/openslides-client/pull/3360 (in the current main, `user.can_manage` is required to see the username, as the new permission is implemented in the backend, but cannot be set in the client).

jsangmeister (approved):

Issue #2066, opened 2023-02-28
Flicker when switching between pages on pdf projection


bastianjoel:
When changing the page on a pdf slide the screen is completely white for a moment. 

It should be waited until a slide is ready before displaying it.

rrenkert:
@MSoeb Can you check this issue again and close or confirm?

bastianjoel:
WIth big pdf files this should still happen as we still do not do a prerendering of slides.

MSoeb:
I've tested it with a small (3 pages) and larger (51 pages) pdf file. The page switches looked smooth and good. No white page between page switch. 

From my perspective it looks fine and therefor I will close this issue.

PR #3490, opened 2024-03-27
Fix blank screen on projection switch

### LINKED ISSUE #2922: Prevent white projector when motion switch
**Current behaviour:**
If one motion is projected and then another motion is projected there is a 1-2 sec delay where the projector has no text but white space 

**Reproduction:**
in a meeting:
1. create two motions with different titles and texts
2. project one of the motions 
3. open the projector-list
4. now project the other motion
5. there should be a short moment where the projector is completely white

**Wanted behaviour:**
The time to switch the motions should be reduced

bastianjoel:
resolves #2922 

With this PR before a new projection is displayed it will be waited until all current projections loaded their content.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/projector/components/projector/projector.component.ts
+++ b/client/src/app/site/pages/meetings/modules/projector/components/projector/projector.component.ts
@@ -4,6 +4,7 @@ import {
     BehaviorSubject,
     combineLatest,
     distinctUntilChanged,
+    filter,
     map,
     merge,
     mergeMap,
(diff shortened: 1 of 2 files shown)


bastianjoel:
Should now also improve the situation with #2066

emanuelschuetze:
@rrenkert needs performance tests with 1000+ users

bastianjoel:
I don't think this necessarily needs performance tests. The issue was reproducible even without any load for motions.

rrenkert (approved):
Did the performance tests. Works great.

PR #3492, opened 2024-03-27
Internal state indicator for change recommendations

### LINKED ISSUE #3428: Internal recommendation: Add info about it in summary of changes
Description: Currently the visibility status of an change recommandation - internal or visible for all - could not be determined quite easily. Only indicator for this is the metadata and the internal label beside the Recommendation. (see green quare)

Example: The green square represent the current information, if a recommendation is in an internal state. The summary of changes (red square) instead doesn't provide this information. The change recommendation misses the information, that it is currently in an internal state and can't be seen by everyone. 

image

Reproduction:
1. open a meeting
2. open motions > open workflows > open complex workflow > set permission 'Show recommendations not public' for first state
3. open motions and create a motion with complex workflow
4. open settings > motions > add 'ABK' in Name for recommender
5. head back to the motion > field ABK in Metadate should have an internal information
6. Add now an change recommendation to the motion text > summary of changes should appear without the needed information

What should happen:
If a motion/ an amendment has a workflow state which sets recommendations to internal, than change recommendations in the summary of changes should also have an information about there current visibility state. 

The summary of changes for change recommendation in an internal state should like the following example:
Line 4 (Change recommendation) ] – _Replacement_ (internal)

Important: When the workflow changes to a state, which doesn't have internal change recommendations, than INTERNAL should not be visible, like the in the metadata.

bastianjoel:
resolves #3428

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-diff-summary/motion-detail-diff-summary.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-diff-summary/motion-detail-diff-summary.component.html
@@ -20,7 +20,9 @@
                     </a>
                     <i class="user-subtitle" *ngIf="isChangeRecommendation(change)">
                         – {{ getRecommendationTypeName(change) | translate }}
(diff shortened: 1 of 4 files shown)


Elblinator (approved):
code and function

tested together with @MSoeb

Issue #3442, opened 2024-03-15
Fix new linter warnings


Elblinator:
PR #3427 added new warnings for 
'@typescript-eslint/explicit-member-accessibility' and '@typescript-eslint/explicit-function-return-type'

we should continuously try to fix these warnings

PR #3510, opened 2024-04-08
Cleanup of linter warnings

Elblinator:
part of #3442

### DIFF
--- a/client/src/app/site/services/global-search/global-search.service.ts
+++ b/client/src/app/site/services/global-search/global-search.service.ts
@@ -56,7 +56,7 @@ export class GlobalSearchService {
         };
     }
 
-    public getTitle(collection: string, content: any): any {
+    public getTitle(collection: string, content: any): string {
         if (collection === `user`) {
             const firstName = content.first_name?.trim() || ``;
             const lastName = content.last_name?.trim() || ``;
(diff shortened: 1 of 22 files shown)


client/src/app/site/services/global-search/global-search.service.ts
```diff
@@ -190,7 +190,7 @@ export class GlobalSearchService {
         };
     }
 
-    private getText(collection: string, result: GlobalSearchResponseEntry, results: GlobalSearchResponse) {
+    private getText(collection: string, result: GlobalSearchResponseEntry, results: GlobalSearchResponse): any {
```
luisa-beerboom, line 193: I don't like that the `getText` function returns any, would it be possible to make it more precise to find the possible values of `GlobalSearchResponseEntry.content` and maybe cast the data/throw an error if it isn't as expected and maybe edit the `GlobalSearchResponseEntry` interface to be more precise?

luisa-beerboom (approved):
Generally good, just one remark.
If that part can not be fixed right now, just merge this and maybe look into creating a cleanup issue for that section :)

Elblinator:
I updated the return value to only string and tested it with different return values, I got no error of any kind

luisa-beerboom (approved):

PR #3532, opened 2024-04-10
Fix mat-hint overlapping with following components

### LINKED ISSUE #3514: Meeting > Settings > Participants: Help text not fully readable
**Description:** In the participants e-mail settings the 'sender name' help text is not fully readable. Some text segments are under the next setting field. How much text can't be read depends on the used language. 

**Reproduction:**
1. open a meeting
2. open settings > participants > E-Mail > sender name
3. Switch Languages to check how much isn't readable.

image

**What should happen:** 
The help text should be readable in every language and the space between help text and the next field should be improved.

reicda:
close #3514 

Originally it was anticipated that it's a single case in meetings > settings > participants. That was not the case as the bug was just not detected in other parts of the application.


bastianjoel (approved):

bastianjoel:
The subscript sizing is actually a new feature from angular material. In the past a dynamic subscript sizing was not possible and we had a workaround for this. Therefore related to #2934.

Elblinator (approved):

PR #3551, opened 2024-04-12
In pdf export of attachment trim trailing '/' of url setting

### LINKED ISSUE #3539: PDF Export: file attachment has wrong URL
**Description:** If a motion with a file attachment is exported, the URL of it is wrong in the PDF file. The URL contains one / too much.

problem - double slash behind .com:
image

**Reproduction:**
1. open a meeting
2. open files and add a file
3. open motions > create a motion and add the file as an attachment
4. open motions > three dot menu > Export > PDF Options > activate Attachments > click 'Export'
5. open PDF File and check to attachment link.

**What should happen:**
The file attachment URL in the pdf should be correctly included - without the double slash - in the pdf file.

reiterl:
Resolve #3539

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
@@ -740,7 +740,10 @@ export class MotionPdfService {
     private createAttachments(motion: ViewMotion): object {
         let width = this.pdfDocumentService.pageSize === `A5` ? PDF_A5_POINTS_WIDTH : PDF_A4_POINTS_WIDTH;
         width = width - this.pdfDocumentService.pageMarginPointsLeft - this.pdfDocumentService.pageMarginPointsRight;
-        const instancUrl = this.organizationSettingsService.instant(`url`);
+        let instancUrl = this.organizationSettingsService.instant(`url`);
+        if (instancUrl.endsWith(`/`)) {
+            instancUrl = instancUrl.slice(0, -1);
+        }
 
         const attachments = [];
         attachments.push({
--- a/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
@@ -741,7 +741,7 @@ export class MotionPdfService {
     private createAttachments(motion: ViewMotion): object {
         let width = this.pdfDocumentService.pageSize === `A5` ? PDF_A5_POINTS_WIDTH : PDF_A4_POINTS_WIDTH;
         width = width - this.pdfDocumentService.pageMarginPointsLeft - this.pdfDocumentService.pageMarginPointsRight;
-        const instancUrl = this.organizationSettingsService.instant(`url`);
+        const instanceUrl = this.organizationSettingsService.instant(`url`);
 
         const attachments = [];
         attachments.push({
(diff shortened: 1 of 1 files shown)


Elblinator (approved):

jsangmeister:
This is not a clean solution, URLs should never be simply concatenated, but joined by a dedicated functions. This removes the possibility of these kinds of errors. Therefore, the string concatentation in lines 769f.:
https://github.com/OpenSlides/openslides-client/blob/204a66dcf2cc7819a0783e6d897cc5433faaec91/client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts#L769-L770

should be replaced by a suitable URL join method instead of trimming the instance URL manually.

reiterl:
@jsangmeister Please give an example solution for this. I discussed this issue 0.75 h with another frontend developer. We didn't come to a solution. (URL doesn't work here).

jsangmeister:
Why does `URL` not work? From a quick search, https://developer.mozilla.org/en-US/docs/Web/API/URL/URL seems to do what we want.

reiterl:
The base with subpathes doesn't work like expected
```
e.g. url = "https://example.com/test/"
path = "/system/get/4"
new URL(url,  path).toString() -> "https://example.com/system/get/4"
```

bastianjoel:
> The base with subpathes doesn't work like expected
> 
> ```
> e.g. url = "https://example.com/test/"
> path = "/system/get/4"
> new URL(path, url).toString() -> "https://example.com/system/get/4"
> ```

It should work if you remove the first slash from the path. But I second that using `URL` is not really necessary here. As said in my previous comment you might just use another way to get the instance url so you do not need to rely on user input.

jsangmeister:
`Location.joinWithSlash` seems to be a solution (https://angular.io/api/common/Location#static-properties)

reiterl:
I will check the Location.joinWithSlash Solution.

reiterl:
The Location joinWithSlash solution seems to work fine. I also checked url with subpaths. 

Please review again.

client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf.service/motion-pdf.service.ts
```diff
@@ -763,8 +764,8 @@ export class MotionPdfService {
                 attachments.push({
                     ul: [
                         {
-                            text: attachment.getTitle() + `: ` + instancUrl + fileUrl,
-                            link: instancUrl + fileUrl,
+                            text: attachment.getTitle() + `: ` + Location.joinWithSlash(instancUrl, fileUrl),
+                            link: Location.joinWithSlash(instancUrl, fileUrl),
```
jsangmeister, line 768:
```suggestion
                const link = Location.joinWithSlash(instanceUrl, fileUrl);
                attachments.push({
                    ul: [
                        {
                            text: attachment.getTitle() + `: ` + link,
                            link: link,
```
While you're at it, please fix the variable name typo (`instancUrl` -> `instanceUrl`)

jsangmeister (changes requested):

bastianjoel (changes requested):
I still don't like this usage of `this.organizationSettingsService.instant('url')`.

reiterl:
@bastianjoel  I didn't change the way how to get the organization setting here. What don't you like about it this way?

jsangmeister (approved):

bastianjoel (approved):
Because you could just use the current url and with that prevent dealing with issues coming from a url that can be set by user input. 

Anyway I see that this might be helpful to for example create documents from staging instances.

reiterl:
@Elblinator Please check, if this is still okay, after the improvements to the code.

Elblinator (approved):
seems to work

PR #3553, opened 2024-04-12
Set http subscriptions inactive after error

bastianjoel:


emanuelschuetze (approved):
works with tablet test

luisa-beerboom (approved):

PR #3573, opened 2024-04-17
Improve meeting user relation handling

### LINKED ISSUE #2789: Client relation handling doesn't do back-relations anymore
It used to be that if one followed a relation field in a subscription, the contra-relation would automatically be filled for the other viewModel as well.

I.e. if one subscribed to a meeting and followed the motion_ids, the `ViewMeeting` would have a field `motions` with all of the corresponding `ViewMotion`s and the corresponding `ViewMotion`s would all have a field `meeting` that was filled with the `ViewMeeting` in question.

It seems like that has somehow stopped happening. 
There have recently been multiple bugs that were caused because a subscription followed a field, but the counter-relation wasn't named anywhere in the fieldset, causing the connection to either not be built, or only be built one-way (I'm not quite sure which).
An example of this would be #2710 which could not be fixed without adding the back-relation for the meeting-users `user_id` field into the fieldset of the follow.

bastianjoel:
resolves #2789 

Also cleans up all places were `meeting_user_ids` were subscribed.

### DIFF
--- a/client/src/app/gateways/repositories/meeting_user/meeting-user-repository.service.ts
+++ b/client/src/app/gateways/repositories/meeting_user/meeting-user-repository.service.ts
@@ -1,4 +1,5 @@
 import { Injectable } from '@angular/core';
+import { Id } from 'src/app/domain/definitions/key-types';
 import { MeetingUser } from 'src/app/domain/models/meeting-users/meeting-user';
 import { toDecimal } from 'src/app/infrastructure/utils';
 import { ViewMeetingUser } from 'src/app/site/pages/meetings/view-models/view-meeting-user';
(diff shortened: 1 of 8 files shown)


bastianjoel:
It might be that this comes with performance penalties in some scenarios. We should do some testing before merging this.

luisa-beerboom (approved):

bastianjoel:
I tested this now a bit and could only observe a minor performance impact. 
To reduce this impact I added two minor tweaks that should especially improve autopilot performance. In my opinion this is ok to merge.

bastianjoel:
I have a better Idea to make this more performant. Please do not merge yet.

luisa-beerboom (approved):

Elblinator (approved):

PR #3598, opened 2024-04-23
Show username if canSeeSensibleData in participant-list

### LINKED ISSUE #3548: Make Username visible if ```can_see_sensible_data``` ist true
**Current behaviour:**
Even if a person is allowed to see sensible data the Username in the participant-list will never be displayed

**Reproduction:**
1. login as an admin
2. navigate to a participant list
3. create a participant without surname and firstname and only username; save
4. navigate back into the participant-list 
5. User ## will be displayed in the participant-list 

**Wanted behaviour:**
Given that a participant has the permission to see the username:
In the participant-list:
The display should be like the display in the account list
account-list:
accound-list
participant-ist:
participant-list
(description: the username should be displayed under the other name, if vote weight is activated then the username should be on the most left hand side and vote weight and name should be separated with a middot)

Several snackbars, views or pages: 
with diverse actions you can see the string "User ##" so far I cannot identify an actual loss of functionality **IF** the username is also additionally displayed in the participant-list. this may need some revisiting after the participant-list is fixed. But this should be done is a separate Issue then.

reiterl:
Resolve #3548

### DIFF
--- a/client/src/app/site/pages/meetings/pages/participants/pages/participant-list/components/participant-list/participant-list.component.html
+++ b/client/src/app/site/pages/meetings/pages/participants/pages/participant-list/components/participant-list/participant-list.component.html
@@ -50,7 +50,13 @@
                         <div *ngIf="user.pronoun">({{ user.pronoun }})</div>
                     </ng-container>
                 </div>
-                <div class="user-subtitle" *ngIf="showVoteWeight">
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
function

bastianjoel (approved):

PR #3616, opened 2024-04-29
Fix 'user undefined' in election projections

### LINKED ISSUE #3612: Candidate names undefined in assignment poll projection
To reproduce:
* create new election
* add 3 candidates
* create new analog ballot and enter some vote values
* publish ballot
* project ballot to default projector
=> The 3 candidates on projector have NO names ("User undefined"), see:
assignment-poll-slide-without-names

Error is caused by PR #3489 (Change user name fallback). 

Autoupdate-service still sends content-object with username:
```
{
    "collection": "user",
    "content_object_id": "user/1",
    "username": "Administrator (Test structure level)"
}
```
But client has removed "username" from user repository:
https://github.com/OpenSlides/openslides-client/pull/3489/files#diff-a6f945f0c9ba474be45544503717105f5b2bd1e72301c0e0be40467aa94cae16L274

This issue is critical because you cannot project an assignment poll result!

Elblinator:
closes #3612 

I searched for the places the new function is called, the only place I found which calls it is where 'User undefined' is returned.

### DIFF
--- a/client/src/app/gateways/repositories/users/user-repository.service.ts
+++ b/client/src/app/gateways/repositories/users/user-repository.service.ts
@@ -36,6 +36,7 @@ export interface AssignMeetingsPayload {
 
 export interface NameInformation {
     id: Id;
+    username: string;
     first_name?: string;
     last_name?: string;
 }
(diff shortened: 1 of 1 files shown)


bastianjoel (changes requested):
This should be fixed inside the slide as the data returned from the projector does not represent the usual user structure and therefore should be handled differently.

PR #3619, opened 2024-04-29
Change appearance of several detail views

### LINKED ISSUE #2728: Change Appearance of several detail-views
**Which detail views should be changed:**
- Motion (both the text and the reason)
- Agenda
- Election
- Participant

**What should be changed:**
- every title should stay how they are (relevant for Motion, Agenda and Election)
- H1 should not have primary or accent colours, but should be black/greyish like the rest of the text
- H3 should have font-weight: 600
- H4 should have font-weight: 600 and font size: 1 em
- H5 should be kursive and have font-size: 1 em
- H6 should not be kursive or bold, and font-size: 1 em

**Changes which will follow _after_ this one:**
- The pdf-export of these four views should be adjusted (see https://github.com/OpenSlides/openslides-client/issues/2118)
- The projector-view of these four views should be adjusted

reicda:
closes #2728


Elblinator (approved):

bastianjoel (changes requested):
It would be better if you use a scss mixin for this or add the styles to the global stylesheet.

bastianjoel (approved):

PR #3621, opened 2024-04-29
Fix editorial final version in a motion

### LINKED ISSUE #3613: Editorial final version switch back to diff version after each autoupdate
To reproduce:
* create new motion
* set motion state to accepted
* create editorial final version
=> error 1: text should switch to created "editorial final version" view - diff version is shown instead
* select "editorial final version"
* click edit button -> edit form opens
* make some text changes (without saving)
* Now: open 2nd tab and create an autoupdate (e.g. set a user to present)
=> error 2: motion detail view in tab 1 is switched from "editorial final version" to "diff version"! Edit form is closed without saving! Expected: should keep the selected form open

Workaround for error2: 
go to motion settings and change "default text version" from "Diff version" to "Final version".

This issue is critial because you cannot work in final version editor as long as default is "diff version" (which is default during meeting). Each autoupdate breaks your work...

Elblinator:
closes #3613 

New behaviour for the motion detail view:
If a editorial final version exists that version will now be displayed per default (even if the settings say sth else)


emanuelschuetze (changes requested):
After saving editorial final version I switched back to (default) diff version. Can we change this and keep the user in the (saved) editorial final version?

emanuelschuetze (approved):

PR #3624, opened 2024-04-30
FIX Search Dialouge Layout

### LINKED ISSUE #3568: Global search: layout of dialog
**Description:** The layout of the dialog has changed. The top edge is to far away from the search field. The lower edge is too near to the filter marks. I guess, this has happened with an angular update. 

example:
image

**Reproduction:**
1. open an instance
2. open the global search

**What should happen:**
The layout should be optimized to fit the content of the dialog. May be a relocation of the whole content a bit higher would be helpful. The distance between content and upper/lower edge should be balanced.

reicda:
closes #3568


Elblinator (approved):

bastianjoel:
Please update the commit messages before merging.

PR #3626, opened 2024-04-30
Fix list views not switching language

### LINKED ISSUE #3580: Filter in list-view: language switch needs reload to be active
**Description**: If a user is in the contributions menu and sets the language from German to English, the English terms in the filter will only be visible after reloading. 

**affected menu in contributions:**
image

**Reproduction:**
1. open an instance and set your language to german
2. open a meeting
3. open participants > contributions
4. change language to german
5. check filter menu
6. If this don't work, do a reload after language switch to german. 

**What should happen:**
Switching the languages from german to english in the contributions menu should instantly change the terms in english.

reicda:
closes #3580


bastianjoel (approved):

Elblinator (changes requested):
Still not immediately translated: 
Agenda > Tags: 'not specified'

Elblinator (approved):

PR #3632, opened 2024-04-30
Fix too many updates for meetingSettingsService.get() 

### LINKED ISSUE #3631: Fix too many updates for meetingSettingsService.get()
the MeetingSettingsService.get() gets triggered/updated with every Autoupdate, even if nothing changed about it.

For example that the recommendation mode jumped back to the default (see https://github.com/OpenSlides/openslides-client/issues/3613) is a side effect of this

The MeetingSettingsService.get() should only be updated if the settings actually get updated

Elblinator:
closes #3631
reverts part of #3621


bastianjoel (approved):

bastianjoel (approved):

Elblinator:
waiting for @rrenkert

rrenkert (approved):

PR #3649, opened 2024-05-07
Meeting list csv export

bastianjoel:


Elblinator (commented):
You can currently export all meetings you see in the list (you can only export a subset if you use the filter)
Is it also wanted to export meetings which are selected via multi-select
@emanuelschuetze

emanuelschuetze (approved):
works good. multiselect not required.

luisa-beerboom (approved):

PR #3670, opened 2024-05-15
pick-to-staging: Trigger on pr_target event type

peb-adr:


peb-adr:
I'll try and see if the secret can be accessed when running on this event type instead.\
This was the only difference I could find to the `.github/workflows/close-issues.yml` action, which seemingly has no problem reading secrets.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target \
In the second paragraph in this doc it says that this Action will run in the context of the base (i.e. upstream repo) instead of the merging branch. This might actually make the difference. Let's see ...

Issue #3605, opened 2024-04-25
[Meta] Motion Forward: motion identifier not shown and Error thrown in console


MSoeb:
**Description:** 
When opening a forwarded motion from the original motion, the motion identifier is displayed incorrectly in the target event. Instead of "Motion 03", for example, "Motion not found" is displayed. 

At the same time, an error message appears in the console. I cannot say whether this is the cause or whether another error is occurring at the same time. 

**target meeting with wrong motion identifier:**
![image](https://github.com/OpenSlides/openslides-client/assets/105491030/25f7a530-1072-47ac-834b-6b8e0e49a29b)

**error message after forwarding:**
![image](https://github.com/OpenSlides/openslides-client/assets/105491030/3a12eddd-296d-4aeb-a003-36f1263df2b8)

**Reproduction:** 
0. Activate motion forwarding in two committees and create a meeting in each one
1. open console to see Error when it is thrown
2. open meeting A 
3. open motions > workflow > > complex workflow > allow forwarding in state approved
4. open motions > create a motion with workflow complex and as submitter "Test". 
5. change status of the motion to approved and forward the motion to meeting B
6. now open the target meeting by clicking on "Motion forwarded to" link
7. Error is thrown and title is wrong

**What should happen:**

- [ ] 1. The motion identifier should be readable after opening a forwarded motion from the original meeting (or reverse).
- [x] 2. The error message should not appear in the console. #3688

Elblinator:
If I click on the link to move to the origin/forwarded meeting then I can reproduce the Error

I can not reproduce the first bug

I'll make a Meta Issue out of this and edit the description accordingly

reiterl:
After the steps clicking on forward/origin shows the "not found" in Motion header. I don't see an error here.

reiterl:
Stop here for now.

Elblinator:
I cannot reproduce the 'Motion not found' part
Please clarify the reproduction

MSoeb:
I've tried to reproduce it, but it works fine in the current OS version 4.1.9. 

Because of this, I've changed the issue text. The issue can now be closed after the second part of it is fixed and merged.

Elblinator:
The first bug from the meta issue is still relevant I could reproduce it once 
I forwarded the motion and navigated vis the link to the forwarded motion white fast (maybe something was still loading)

MSoeb:
[...]

PR #3688, opened 2024-05-21
Prevent the error (in console) from forwarded motions

Elblinator:
part of #3605
I still cannot reproduce the other part of the bug but I can reliably reproduce the error in the console


MSoeb (approved):
Works as intented. No error in console.

client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-meta-data/motion-meta-data.component.ts
```diff
@@ -353,8 +353,12 @@ export class MotionMetaDataComponent extends BaseMotionDetailChildComponent impl
...
                 .pipe(
                     map(motion => [
-                        motion.referenced_in_motion_recommendation_extensions,
-                        motion.recommendation_extension_references as ViewMotion[]
+                        motion.referenced_in_motion_recommendation_extensions
+                            ? motion.referenced_in_motion_recommendation_extensions
+                            : null,
+                        motion.recommendation_extension_references
+                            ? (motion.recommendation_extension_references as ViewMotion[])
+                            : null
```
bastianjoel, line 361: I don't see how this should fix the issue. 

Trying to read a property of `null` will still result in throwing a type error. The error message in the issue looks to me like `motion` in line 355 is null. Which also does not really make sense. 

If it is an empty `motion` attribute the solution should look like this:
```suggestion
                        motion?.referenced_in_motion_recommendation_extensions,
                        motion?.recommendation_extension_references as ViewMotion[]
```

bastianjoel (approved):
If you disagree with my last comment I am okay with merging this. I did not look to much into this myself.

Elblinator:
Both solutions seem to prevent the Error and yours is shorter, so I would prefer that one.

MSoeb (approved):

PR #3694, opened 2024-05-22
Update material table

Elblinator:
part of #2934 

the following view pages were changed: 
- history
- groups
- category-detail
- workflow-detail


MSoeb (changes requested):
Here are various minor, recurring problems and one major one.

Minor problems: Various titles that were previously slightly gray and not bold are now bold. This occurs with various content in the instance. A good example can be found in the chronicle. 

Major problem: The width of the boxes has changed in the Autopilot. They are now narrower and the speaking times widget is wider.

Both should be fixed.

Elblinator:
The titles in e.g. motion-meta-data and the size of the boxes in the autopilot were not changed by this PR and this should therefore not be fixed here 

The titles/appearances which should be slightly different are in the four named views (history, groups, category-detail, workflow-detail). 
The difference in appearance is the new default, I strongly recommend not diverting from the new default

bastianjoel (approved):

bastianjoel (changes requested):
Please do not merge into main. 
This PR is included in #3706. I will take care of the future changes made here and add them there. 

The changes made here can also be tested here.

bastianjoel:
closed in favor of  #3706

PR #3695, opened 2024-05-22
Wait until delegation loaded before setting poll ready

### LINKED ISSUE #3687: Fix setting/permission calls on motion-polls
**Description:**
If a User who cannot vote -due to them delegating their vote- opens motion-details views, then they can still see the poll buttons and send their vote. 

**Reproduction:**
(I called stop-dev and run-dev before the reproduction)
1. login as admin
2. activate vote delegations
3. delegate a's vote to b
4. navigate to the first motion in the motion-list (1 -1)
5. put to state submitted, create and start a poll
6. login as a 
7. navigate to the motion
8. you will see the three possible vote options even though the sentence 'Your voting right was delegated to another person.'
9. If you navigate to motion 2 'ohne' the permission will be correctly loaded 
10. the other motions will don't load the permission correctly
11. After the permission was correctly loaded once every motion will show the correct permission

**Wanted behaviour:** 
The permission should be correctly loaded

bastianjoel:
resolves #3687

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/base/base-poll-vote.component.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/base/base-poll-vote.component.ts
@@ -77,7 +77,10 @@ export abstract class BasePollVoteComponent<C extends PollContentObject = any> e
         super();
         this.subscriptions.push(
             this.operator.userObservable.pipe(debounceTime(50)).subscribe(user => {
-                if (user) {
+                if (
+                    user &&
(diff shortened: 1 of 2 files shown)


Elblinator (changes requested):
I opened the motion-detail view, waited for more than 2
10 minutes and the Message why I cannot vote did not appear
Navigating in between motion3 and A1 does not change that.
Navigating to motion 2 loads the message, afterwards the message is displayed everywhere
The poll looked like that:
Screenshot_20240523_091528

Elblinator (approved):

PR #3747, opened 2024-06-06
Show sort-option for vote weight if vote weight is active

### LINKED ISSUE #3746: Fix sort option vote weight
**Current behaviour:**
The sort-option for vote weight is shown even if it should/shouldn't be shown

**Reproduction:**
1. login as an admin
2. Deactivate the vote weight
3. navigate to participant-list
4. the sort option for vote weight is visible

**Wanted behaviour:**
If the vote weight is activated then the sort option for the vote weight should be displayed

Elblinator:
closes #3746


luisa-beerboom (approved):

MSoeb (approved):
works as it should

PR #3760, opened 2024-06-10
Fix polls when delegation settings are changed

### LINKED ISSUE #3685: Fix polls when delegation settings are changed
**Current behaviour:**
If delegations are activated and a poll is not loaded due to this setting then the poll stays not loaded even if the delegations are deactivated (vise verse with deactivated -> activated)

**Reproduction:**
1. in a meeting activate vote delegation 
2. add User A, and delegate their vote 
3. create a poll (choose from topic, motion, election)
4. login as User A and navigate to the poll
5. The poll should say 'Your voting right was delegated to another person.'
6. As the admin deactivate the delegation settings
7. The message disappears but the poll is not loaded
8. As User A reload the page -> the poll is visible 
9. activate the delegation setting
10. As User A the poll stays visible and the user can still send their vote

**Wanted behaviour:**
If the User is allowed to vote then the poll should be visible 
and vise versa the poll should be hidden if the User is not allowed to vote

Elblinator:
closes #3685

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/base/base-poll-vote.component.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/base/base-poll-vote.component.ts
@@ -114,7 +114,11 @@ export abstract class BasePollVoteComponent<C extends PollContentObject = any> e
 
     public canSeePoll(user: ViewUser = this.user): boolean {
         if (user === this.user) {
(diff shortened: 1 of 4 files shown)


bastianjoel:
Closed in favor of #3790

PR #3766, opened 2024-06-11
Merge users together 

### LINKED ISSUE #3725: User merge
The User merge.together functionality in the backend needs appropriate client implementations as well
See the backend PR here: https://github.com/OpenSlides/openslides-backend/pull/2427

**Mock-ups:**
Board

The merge should only be possible from the account-list

1. You select via multi-seelct the accounts you want to merge 
2. then in the three dot menu there should be the option to merge the selected accounts
3. dialog opens
4. you select the reference account, the data from this account will be used, every other data will be **NOT** used
5. the not selected account will be strike through (like the mock-up)
6. The name of the user should be shown like in a user-detail view
7. hovering over the number of meetings should open the toolbox and show up to 10 meetings, if the user is in more than 10 meetings then after 10 there should be a plus
8. hovering over the letter should show when the last invitation-mail was send (same behaviour as the account-list)
9. pressing merge will move all references from the selected accounts onto the same account and delete the now unused accounts
:mage:

reiterl:
Resolve #3725 

It needs the backend action user.merge_together to be implemented.
PR https://github.com/OpenSlides/openslides-backend/pull/2427

At the moment, the user selection dialog is a basic one. We **need** a fancy version of the dialog AFAIK.


reiterl:
Needs the backend pr to be complete.

client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-merge-dialog/account-merge-dialog.component.ts
```diff
@@ -0,0 +1,53 @@
...
+export class AccountMergeDialogComponent {
+    public constructor(
+        public dialogRef: MatDialogRef<AccountMergeDialogComponent, AccountMergeDialogAnswer>,
+        @Inject(MAT_DIALOG_DATA) public data: AccountMergeDialogData
+    ) {}
+
+    public selectedUserId: number;
+
+    public get sortedUsers(): ViewUser[] {
+        return this.data.choices.sort((a, b) => a.name.localeCompare(b.name));
```
bastianjoel, line 22:
```suggestion
    ) {
        return this.data.choices.sort((a, b) => a.name.localeCompare(b.name));
    }

    public selectedUserId: number;

    public get sortedUsers(): ViewUser[] {
        return this.data.choices;
```

Maybe also rename. No need to resort the choices on every change detection cycle.

bastianjoel (changes requested):

client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-merge-dialog/account-merge-dialog.component.ts
reiterl: done

Elblinator (changes requested):
- Make the dialog bigger (see width from the design-dialog)
- Show the last login in a new line and not the icon 
- separate the different accounts with a line 
- If you hover over the icon "event_available"
     - give each meeting it's own line
     - remove the comma between the meeting names
     - if an account is in more than 9 meetings then the number is cropped, display the number fully 
- this "event_available" icon should always be displayed even if the person is in no meeting (then there should be a zero)
- if a primary account is selected strike through every not selected name (title given name and surname)
- the icon "block" should not be displayed for the accounts which are not selected (the icon which indicates that an account is inactive should still be displayed)
- in the first line make the gender middot and pronouns smaller and greyer (like in the second line, for reference see the mock-up, this was in the issue not explained thoroughly)
- Change the string "Attention: Not selected users will be merged and then removed." to "Attention: Not selected accounts will be merged and then deleted."

Elblinator (changes requested):
- Disable the merge button if no primary account is selected

reiterl:
Another version of the merge dialog is ready. Please review.

Elblinator (changes requested):
- The listed meetings from the tooltip from "events_available": please make them align on the left side
- Remove the line under the last account
- A whole account row (every line but the first) is not aligned centrally, please align a row central (the distance between the two lines should be equal
- It is possible to change the appearance from the name (first line): 
currently: it is "title first name surname (pronoun) icons" all in the same colour and size
wanted: 
      - title, first name, surname and icon stay the same
      - only the title, first name and surname should be strike through if not selected
      - the pronoun should lose the brackets and size and color vise should be the same as the second line
      - before the pronouns the gender and a middot should be displayed


It looks already look very pretty! :)

reiterl:
I have changed the component from mat-list to mat-table, because it is more near than what we want. (alignment, colored lines, cells)

reiterl:
I don't know how to fix the meeting left alignment of the tooltip. The rest seems to be done.

reiterl:
To debug the Tooltip, you need to set matTooltipHideDelay high, so that the tooltip is still there, if you want to look at the styles etc in `os-icon-container`.

bastianjoel:
> * The listed meetings from the tooltip from "events_available": please make them align on the left side

This is not possible as the text needs a certain width for angular to detect it as a multiline tooltip. 
See: https://github.com/angular/components/blob/main/src/material/tooltip/tooltip.ts#L1066

reiterl:
Include a **data loss** warning, if two users are in the same meeting.

Elblinator (approved):
For my side appearance is fine as it is now. 
@MSoeb or @emanuelschuetze have to check it in the last step and review whether or not they want something changed

luisa-beerboom (changes requested):
Please look at the user.merge_together wiki entry in the backend:
- The user.merge action replaces data in the poll, aside from the fact that we should possibly include a warning, there is also the fact that the `entitled_users_at_stop` list couldn't simply be re-wired. There are now two new properties that may appear in the list entries called `user_merged_into_id` and `delegation_user_merged_into_id` that mark which users the user, to whom the vote belongs, and the user, to whom the vote was delegated, were merged into. This is important because the original user ids won't exist anymore. Please change it so that in those instances the entitled users list of the poll shows something like `Old account of <current user name>` instead of `deleted user`
- At the end there is a paragraph `Client` that describes things that the operator should probably be made aware of, possibly through warnings. I don't see these things mentioned anywhere. If that was a conscious decision it is fine, otherwise please discuss it and build it in.

MSoeb (changes requested):
Some points still need to be adjusted:

- [ ] Entitled user list (in polls and elections): For merged users, 'Anonymous' is displayed in the entitled users list and not the new user name to which the old account was merged. The merged account name must be displayed here.

- [ ] History: In the history, all old account names in the 'Edited by' column are hidden after a merge. Information is lost in the merge process. Instead of an empty field, the FQ user ID should be displayed so that it is still possible to track which user has edited.

- [x] Merge dialog: If there are several users in a meeting that are to be merged, an additional warning is displayed. This is correct. In addition, a note should appear below this warning indicating how many meetings are affected. 

About Luisa's notes:
- The second point is obsolete. We will make this information available via the FAQ.
- A warning is mentioned in the first bullet point, but this is not required.

reiterl:
Add the count of the effected meetings to the warning.

reiterl:
I add code to the `EntitledUsersTable` and the `BasePollPdfService` about the new entitled user entry info.
I also show the fqid in the history in the case non user can be loaded case.

reiterl:
This feature needs a large documentation in the FAQ as mentioned by @MSoeb .

client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.html
```diff
@@ -18,14 +18,36 @@
...
                         <!-- Delegation -->
-                        <div *ngIf="entry.vote_delegated_to_user_id">
+                        <div *ngIf="entry.vote_delegated_to_user_id && !entry.delegation_user_merged_into_id">
                             <span>
                                 ({{ 'represented by' | translate }} {{ entry.vote_delegated_to.getShortName().trim() }})
                             </span>
                         </div>
+                        <div *ngIf="entry.vote_delegated_to_user_id && entry.delegation_user_merged_into_id">
+                            <span>
+                                ({{ 'represented by old account' | translate }}) {{ delegatedMergedIntoName(entry) }}
```
luisa-beerboom, line 28: should probably be `represented by old account of ...`

luisa-beerboom (approved):
Better, just one small remark

client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.html
reiterl: I have changed this.

rrenkert (changes requested):
The search in the "Entitled users" of polls (tested with "non-nominal") does not find user that were merged into an other account. An account of that type is named with "Old account of ...".

MSoeb (changes requested):
Errors occur in the list of eligible voters during roll-call votes or elections.

The individual vote record no longer shows how a participant voted. Only the name is visible, without the voting behavior with icon and text. An error message about the problem is displayed in the console.

Reproduction:
1. open a Meeting
2. open motions > create a motion > create nominal vote and select 'delegates' as entitled to vote group > start the voting
3. open participants > create a user A as delegate 
4. log in as user A in another browser > open user profile and set yourself as present > open the autopilot > vote
5. head back to superadmin browser > open motions > stop the poll and publish it > open the detail view of the poll with single votes and the entitled user list. Single votes list is broken. Additionally check your dev-console. several error will be shown.

reiterl:
Fixed the entitled user table search problem.

reiterl:
I cannot reproduce the failure @MSoeb reported. It has nothing to do with merged users I suppose, because no user is merged in the step-by-step description.

MSoeb (changes requested):
Tried it again with an updated instance of this pull request. My described problem still occurs for nominal polls and elections. Tested it also again on version 4.1.11. In this version it doesn't occur. 

This is a soft indicator, that the problem occurs because of the PR changes or it occurs because of some other developments. I can't say, what is the reason for it. Nevertheless it should work. 

Image of the error message:
image

If you need help for reproduction, please contact me.

MSoeb:
In addition, in the electoral roll, only the name can be searched for when changing names, but not all text elements. The name works great. It would be good if the remaining words could also be searched for. This would make the search easier. 

Example: "Old account of Max Mustermann." 
You can search for the terms Max and Mustermann. The terms 'Old account of' do not work. 

Please implement this so that these terms can also be found via the search.

reiterl:
The terms "Old account of" and "represented by old account" can also searched for.

Elblinator:
The bug in the console now fixed

client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.html
```diff
@@ -18,14 +18,36 @@
...
+                    </div>
+                </div>
+                <div *ngIf="!entry.user && entry.user_merged_into_id">
+                    {{ entry.user_merged_into }}
+                    <div class="user-subtitle">
+                        <!-- Delegation -->
+                        <div *ngIf="entry.vote_delegated_to_user_id && !entry.delegation_user_merged_into_id">
+                            <span>
+                                ({{ 'represented by' | translate }}
+                                {{ entry.vote_delegated_to?.getShortName().trim() }})
```
bastianjoel, line 40: `getShortName` already trims its output. Therefore change to:
```suggestion
                                {{ entry.vote_delegated_to?.getShortName() }})
```

client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-merge-dialog/account-merge-dialog.component.html
```diff
@@ -0,0 +1,104 @@
...
+                    </div>
+                </td>
+            </ng-container>
+
+            <ng-container matColumnDef="icon">
+                <td mat-cell *matCellDef="let user">
+                    <div class="iconCell">
+                        <os-icon-container
+                            icon="event_available"
+                            [iconTooltip]="userMeetings(user)"
```
bastianjoel, line 70: This might be problematic in terms of performance as the method does processing with arrays. To fix this you could use change detection `OnPush` (preferred) or store and access the meeting names differently.

bastianjoel (changes requested):

client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.html
reiterl: done, changed at two places.

client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-merge-dialog/account-merge-dialog.component.html
reiterl: Now use OnPush.

bastianjoel (approved):

MSoeb (approved):
Important: The PR should not yet be merged. The merging is done by @rrenkert

Elblinator (approved):

rrenkert (approved):

PR #3773, opened 2024-06-12
Fix style of the home page

### LINKED ISSUE #3752: UI issue: wrong heading style on Home page
**Description:** The heading styling on the Home page has changed. It has the wrong color. Maybe other styling parts affected, but i'm not sure. Please check it. - I guess this happend by changing the headings in #2728 

**Old Style:**
image

**New wrong style:**
image

**What should happen:**
Heading should have the old Styling.

reiterl:
Resolve #3752

### DIFF
--- a/client/src/app/site/pages/meetings/pages/agenda/modules/topics/pages/topic-detail/components/topic-detail/topic-detail.component.html
+++ b/client/src/app/site/pages/meetings/pages/agenda/modules/topics/pages/topic-detail/components/topic-detail/topic-detail.component.html
@@ -56,7 +56,7 @@
                 <div>
                     <span *ngIf="topic.text">
                         <!-- Render topic text as HTML -->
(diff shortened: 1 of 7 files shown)


Elblinator (approved):
function

bastianjoel (changes requested):
I would prefer a solution without ng-deep. 

Maybe create a new component for user html content and set the components `encapsulation` to `None`.

reiterl:
If we would use an encapsulation of none in the user component, it styles would be used app wide. That is a problem.
encapsulation doesn't have an option for use styles for comp and sub comps.

bastianjoel:
You could use an unique class name then? 
Maybe just change the name `underlined-links` to something else for now.

reiterl:
Renamed 'underlined-links' into 'detail-view' as the class provides the 'detail-view-appearence'.

PR #3781, opened 2024-06-18
Fix countdown controls layout

bastianjoel:
Currently the alignment of the start/pause buttons within the projector countdown control cards is broken.


Elblinator:
The broken behaviour is that the buttons fall out of the mat-card if the screen is too small

Old appearance:
Screenshot_20240625_130818


New appearance:
Screenshot_20240625_130913

Elblinator (approved):

PR #3789, opened 2024-06-24
Update disabled setting fields at SaveAll, too

### LINKED ISSUE #3779: Fix wrong settings updates
**Current behaviour:**
If a sub setting is changed but the main setting is disabled then the changed values from the subsetting is not saved (or send in the send request)

**Reproduction:**
1. In a meeting navigate to settings > participants
2. open console
3. Activate vote delegation 
4. Save change
5. Activate the other delegation settings AND deactivate vote delegation
6. Save changes 
7. look in the console: only users_enable_vote_delegations is updated
8. Reload page, the subsettings are not updated

**Wanted behaviour:**
The subsettings should be updated/saved even if the mainsetting is disabled

reiterl:
Resolve #3779 

This resolves a problem with saving disabled fields in the settings dialog. I don't think this should be done this way.


bastianjoel (commented):
I doubt that this is a good idea. This was introduced by https://github.com/OpenSlides/openslides-client/pull/3357#issuecomment-1985399300

I believe this has something to do with
> * fixed the save button in combination with the `disabled` attribute

Maybe there are backend checks which throw errors when changing a disabled transitive setting?

reiterl:
Mmh, I rethought this pr. It is not useful.

PR #3796, opened 2024-06-27
Fix vote components using wrong change detection strategy

### LINKED ISSUE #3795: Cumulative votes get resetted while entering them
**Describe the bug**
While entering cumulative votes, the counter "available votes" firstly counts down correspondingly, but after some time (about 15-30 seconds) it counts upwards again - although nothing is changed - preventing the user from voting correctly.

**How to Reproduce, i.e., the situation in which we experienced this behaviour**
1. We had an election (min. votes = 1, max. votes = 15, max. votes per option = 3, 25 options) 
2. Started voting
3. Started distributing the 15 votes.
4. This was done in a very high load situation in which about 100 people were individually walking through about 120+ motions distributed over 7 meetings in all of which such voting was performed (the system responses were quite slow and regularly the connection was lost)
Initially, the "available votes" counter corresponded to the given votes in the front-end. However, always after some time, the counter increased again while no such change was performed in the entries. Since the election was quite big, going fast enough through all the entries to bring the counter to a consistent state was hard/not possible all the time.

**Expected behaviour**
The counter "available votes" equals the sum in the entries and does not change without the corresponding user action.

**System information**
- OpenSlides version: 4.1.7 (but one year ago we encountered the same problem)
- Additional version information (if any): 2024-04-18
- Device: Notebook
- OS: Windows 11, Windows 10
- Browser: Chrome, Firefox

**Additional context**
We experienced the same behaviour one year ago with an older version of OpenSlides 4 as well.

bastianjoel:
resolves #3795

Please test if motion and assignment polls are still working as intended. As this already works fine for topic polls like this I don't think changing the strategy should make any problems. 

Calls to `getVotesCount` after an autoupdate were reduced from +100 to ~4.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.html
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.html
@@ -63,7 +63,7 @@
                         }"
                     >
                         <div *ngIf="showAvailableVotes" class="grid-name-area"></div>
-                        <b *ngIf="!showAvailableVotes" class="grid-name-area">
+                        <b *ngIf="!showAvailableVotes && poll.option_ids" class="grid-name-area">
                             {{ poll.option_ids.length }} {{ optionPluralLabel | translate }}
                         </b>
                         <b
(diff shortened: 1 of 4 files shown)


luisa-beerboom (approved):

bastianjoel:
@Elblinator ping

Elblinator (approved):

PR #3801, opened 2024-06-27
Fix list of speakers names not displayed

### LINKED ISSUE #3798: Autopilot > List of speakers: Missing names after reload
**Description:** There is a strange error in the autopilot that hides the names of the users. When you reload the autopilot, it can happen that user names are hidden in the list of speakers and in the 'Last speakers' list. If you then hover over the X in the list of recent speakers, for example, the names will reappear. Unfortunately, I cannot say whether there is a system for hiding user names. It seems arbitrary. 

**Example of missing username in last speaker list**
image

**Reproduction:**
1. open the OS4 Demo > Meeting Bundeskonferenz 2024
2. open the Autopilot
3. Reload the page -> Usernames disappear
4. Open 'Last speakers' in Autopilot and hover over the X. Usernames appears again. Hover over the Play-Button -> Names will be visible again. 

Hint: Unfortunately, the problem can only be reproduced cleanly once after calling up the instance. To provoke it again, it sometimes works for me to reload the page with Ctrl+Shift+R. 

**What should happen:**
The list of speakers and the last speakers list should always display the usersnames after reloading the Autopilot.

bastianjoel:
resolves #3798

I was not able to reproduce the problem but I found something that could have caused it.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-content/list-of-speakers-content.component.ts
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-content/list-of-speakers-content.component.ts
@@ -116,6 +116,7 @@ export class ListOfSpeakersContentComponent extends BaseMeetingComponent impleme
         if (los) {
             this._listOfSpeakers = los;
             this.updateSpeakers();
+            this.cd.markForCheck();
         }
     }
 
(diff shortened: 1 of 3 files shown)


MSoeb (changes requested):
This fix doesn't work. It made it even worse. Now I got the problem directly by opening the Autopilot. - @bastianjoel let's talk about it.

bastianjoel:
Closing for now.

PR #3808, opened 2024-07-02
Fix motion polls

### LINKED ISSUE #3805: Fix motion poll layout
**Current appearance from a motion poll:**
Screenshot_20240628_104923

**Wanted appearance:**
The buttons from a motion poll should not overlap and be distinct from each other and you should see which button you actually activated

Elblinator:
closes #3805

### DIFF
--- a/client/angular.json
+++ b/client/angular.json
@@ -60,7 +60,7 @@
                                 {
                                     "type": "anyComponentStyle",
                                     "maximumWarning": "4kb",
-                                    "maximumError": "8kb"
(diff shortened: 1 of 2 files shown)


rrenkert (approved):

rrenkert (changes requested):
In motion detail view the buttons keep in one line when shrinking the screen size. They should switch to a vertical layout.

rrenkert (approved):
Layout is ok now.

luisa-beerboom (approved):

PR #3813, opened 2024-07-03
Fix several Layout issues

### LINKED ISSUE #3811: Several Layout issues
Description: Several layout issues occur. 

**Problems:**

Image: left side is Staging / right side is live
image
- [x] The bar in the navigation menu is in the wrong place. Apparently the spacing is no longer correct.  (This also occurs in meetings)
- [x] The distance between the active meetings is incorrect.

Image: Left side is staging / right side is live
image
- [ ] The matcard size of the meeting has changed. 
- [ ] The blue bar is also resized, which leads to a wrong positioning of the text. 

History:
- [ ] Position of the history matcard has changed. 

**What should happen:**
The original layout should be reestablished again.

Elblinator:
closes #3811

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.scss
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.scss
@@ -23,10 +23,6 @@
     }
 }
 
-.mat-divider {
-    margin-bottom: 16px;
-}
-
 .percent-base-panel {
     max-height: 300px !important;
 }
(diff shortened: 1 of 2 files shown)


Elblinator:
Only fixed the first part because that's the only part I know how to reproduce

The second commit fixes an unrelated Error in the console. For an easier staging process we decided to fix this in the same PR
Reference here: 
https://github.com/OpenSlides/openslides-client/pull/3766#pullrequestreview-2153544866

rrenkert (approved):

luisa-beerboom (approved):

PR #3833, opened 2024-07-09
Allow committee admin personal info edit

luisa-beerboom:

PR #3722, opened 2024-05-31
Reorder the user imports fields

reiterl:
Resolve #3719

PR #3852, opened 2024-07-15
Reset meta to 2e80fc0

peb-adr:


peb-adr:
It seems #3722 and #3833 mistakenly committed a new meta reference.

This will reset it to `2e80fc0` in order to have consistency again for staging 4.1.13

PR #3863, opened 2024-07-16
Support `{background-}color: inherit` in htmltopdf

### LINKED ISSUE #3856: Fix pdf export if background colour is set
**Current behaviour:**
If the background colour from a text part is set to e.g. blue and this text is exported as pdf then only the background is visible and the text is gone

**Reproduction:**
1. Create a motion where part of the text has a different background colour
2. export the motion as a pdf
3. Look at the pdf => only the background colour and not the text is visible

**Wanted behaviour:**
If a background colour is set the text should still be visible

bastianjoel:
Might resolve #3856 but I was not able to reproduce the issue.

### DIFF
--- a/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
+++ b/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
@@ -192,6 +192,7 @@ export class HtmlToPdfService {
                 newParagraph = this.createFormattedParagraph(createPayload);
                 break;
             }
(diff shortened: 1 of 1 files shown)


rrenkert (commented):
Does not fix the issue for formatting complete lines and the beginnig of a line. If the coloring starts in the middle of a line the text is correct formatted in the PDF.

bastianjoel:
This also fixes text colors not being applied in marked text.

rrenkert (approved):

PR #3877, opened 2024-07-17
Fix projector buttons not using contrast color for icon

bastianjoel:
Accent color buttons currently have black icons instead of white ones.


Elblinator:
Repo: 
1. create a motion
2. project the motion
3. the projection button should be white (wanted behaviour)

Elblinator (approved):

PR #3879, opened 2024-07-17
Cleanup controlflow

bastianjoel:
Simplifies some control structures in templates as we now can use `else`.


Elblinator:
npm run cleanup still throws @angular-eslint/template/prefer-control-flow warning
Is it okay, if you fix them in here as well?
The file which throws this is: src/app/site/pages/organization/pages/accounts/pages/account-detail/components/account-detail/account-detail.component.html

bastianjoel:
This is not a linter warning cleanup PR

bastianjoel:
> I think this also improves the performance a bit.

Yes, this is the intention of this PR.

PR #3898, opened 2024-07-18
Add missing 's' tag to html to pdf

### LINKED ISSUE #3889: Motion pdf export: Multiple formatting is ignored
**Describe the bug:**
The pdf export of motions does not work correctly if a text passage has more than one formatting options activated at the same time (e.g. cursive and bold). Only one formatting is exported. 
A cursive and bold word would only be cursive or would only be bold.

**How to Reproduce:**
1. open meeting
2. open motion
3. create motion > mark text with bold and striketrough* > save
4. export motion as pdf

*This is an example. Several other combinations didn't work well. Please check them. The images below show as an example, what does work and what not. But these are not all possible options. 

**Expected behaviour**
Every text formatting combination should be exported in the pdf export.

**Additional context**
Example Motion Text:
image

Example of this motions text as pdf export:
image

reiterl:
Resolve #3889 

Please check, if this improves the situation in #3889 . I just added the missing 's' for strike thru tag.
If there are other problems, I will need a more deep dive into the html-to-pdf.service.

### DIFF
--- a/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
+++ b/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
@@ -65,6 +65,7 @@ export class HtmlToPdfService {
         h6: [`font-size:10`],
         a: [`color:blue`, `text-decoration:underline`],
         strike: [`text-decoration:line-through`],
+        s: [`text-decoration:line-through`],
         // Pretty specific stuff that might be excluded for other projects than OpenSlides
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
The code seems fine
I tested a lot of cases and I didn't see any problems anymore

I tested every combination from bold (with italic, underline, strikethrough, font colour and background colour) and italic (with bold, underline, strikethrough, font colour and background colour) and everything works as expected

rrenkert (approved):

PR #3902, opened 2024-07-19
Add availableAdmins and use all account users for it

### LINKED ISSUE #3893: Committee admin: Meeting creation wizard - issue with admin selection
**Describe the bug**
When the committee admin creates a meeting, the 'Administrators' selection menu is empty. He does not even see himself for selection when creating a meeting. If the meeting is edited in the second step, only the committee admin is listed as an account in this menu. This behavior does not make sense, as the committee admin has to open a meeting in order to set an account as a meeting admin. This is an unnecessary step and makes no sense with the new filtered account list.

**How to Reproduce**
1. open an instance
2. open committees and create a committee
3. open the created committee and create a meeting 
4. open accounts > create two accounts. Account A = set as committee admin of the created committee. Account B = put it in the created meeting
5. Log in as Account A
6. open committes > created committee > create a new meeting > try to select an admin (no Account should be visible) > save meeting
7.  open the edit page of the meeting > try to select an admin. you should only see the committee admin as preselection. Account B can't be chosen, even it is available for the committee admin in the Account list.

**Expected behavior**
When creating meetings, all accounts to which the committee admin has access via the account list should be listed in the 'Administrators' selection.

reiterl:
Resolve #3893

### DIFF
--- a/client/src/app/site/pages/meetings/view-models/view-user.ts
+++ b/client/src/app/site/pages/meetings/view-models/view-user.ts
@@ -1,3 +1,4 @@
+import { Observable } from 'rxjs';
 import { User } from 'src/app/domain/models/users/user';
 import { BaseViewModel } from 'src/app/site/base/base-view-model';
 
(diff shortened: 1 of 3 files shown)


Elblinator (changes requested):
too many people are in the list now 
The committee manager should only be able to select users from their own committees

Another bug (fixed here https://github.com/OpenSlides/openslides-client/pull/3897) made the list too bis as well
Please let only people from their own committee show in the list

reiterl:
Use filter function from @luisa-beerboom to filter the accounts available for Admins in committee admin case.

Elblinator (approved):

luisa-beerboom (approved):

PR #3936, opened 2024-07-25
Change general approval option order

### LINKED ISSUE #3895: Change position of "General approval"
In a ballot with "No per candidate" voting method "general approval" can be activated.

**Current behaviour:**
the candidates are listed 
then there is a or-divider displayed
then the general approval then general abstain is displayed:
Screenshot_20240718_112216

**Wanted behaviour:**
This order should change
The general approval should be displayed first
then a or-divider
then the candidates
then another or-divider
and then the general abstain

Note:
- only the layout from the "No per candidate" should change
- the or-dividers should only be displayed if the connected general option is activated

bastianjoel:
resolves #3895

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.html
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.html
@@ -99,6 +99,55 @@
                                 }
                             </div>
                         }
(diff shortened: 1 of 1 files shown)


Elblinator (changes requested):
Hide the second "Or line" if only one general option is activated in a poll.isMethodN

Repo: 
- start creating a "No per candidate" poll
- activate only one general option
- both "or lines" will be displayed

PR #3937, opened 2024-07-25
Improve projector timers

### LINKED ISSUE #3678: Fix projected timers
**Current behaviour:**
- Timers in fullscreen mode overlap with each other
- "In the projector selection?"-dialog the display type and if the timer is projected as fullscreen or not is not remembered when opening the dialog again 
- A timer can overlap with the clock

**Reproduction:**
1. in the  projector-list edit your projector
2. activate "show clock", DEACTIVATE "show header and footer" and change the background colour to a colour where you can see the clock
3. navigate to projector-detail view 
4. create three timers with 1:30 minutes, 2 min and zero minutes
5. project two as fullscreen and display type "in Time and traffic light"
6. the timers are overlapping (see screenshot)
7. open the projection dialog of either of the timers 
8. the dialog saved on which projector the timer is projected but it does not save if it's fullscreen or the display type
9. project the third timer as not fullscreen
10. the timer overlaps with the clock (see screenshot)

**Screenshot:**
Screenshot_20240516_083700

**Wanted behaviour:**
- [ ] <s>Only one timer should be displayed in full screen mode. If one timer is already displayed in full screen then the option for full screen for other timers should be displayed and there should be a hint which explains that per projector only one timer is allowed in full screen</s> (Given that the discussed desired behaviour is that only one timer at a time should be displayed and this above described behaviour will hinder someone from projecting a timer, I will implement a slightly different behaviour): 
Only one timer should be displayed in full screen mode. If one timer is already displayed in full screen then and another timer is then projected in full screen then the new timer will overwrite the previous timer and there should be a hint which explains that per projector only one timer is allowed in full screen
 edited (15.07.2024) 
- [x] if a timer is currently projected and the edit dialog is opened then dialog should save if it's projected in full screen and the display type
- [x] If the clock is displayed (even without header and footer) the timer should not overlap with the clock

bastianjoel:
resolves #3678

Implementation differs from what was asked for in issue. For details see https://github.com/OpenSlides/openslides-client/issues/3678#issuecomment-2250568951


luisa-beerboom (approved):

Elblinator (approved):
approved by me and @emanuelschuetze 

Note for the future:
another button should also change it's behaviour, Issue will be created and implemented in the future

Issue #3955, opened 2024-07-30
Anonymous user


Elblinator:
see for context: https://github.com/OpenSlides/OpenSlides/issues/6798

**Wanted behavior**
Login mask: 
- [x] below the internal login should be a button with "Login as guest" as text
- [x] If this button is pressed the anonymous user should be:
   - If only one meeting is public -> should be navigated to the public meeting
   - If more than one meeting is public -> should be navigated to the dashboard, where every public meeting is display 
  
General: 
- [x] If you are using OpenSlides as a anonymous... In the top right corner 
  - The icon next to your name should be "no_accounts"
  - The displayed name should be "Anonymous"
  - the "My profile"-Option should be hidden
  - The "Logout" button should have the icon "Login", and if pressed you should be navigated to the login page "..../login"

Orga-level as an anonymous user: 
- [x] Only "Dashboard" should be visible
- [x] In the dashboard every meeting which is public should be displayed

Meeting:
- [ ] Settings:
   - There will be a new setting after Locked from inside (https://github.com/OpenSlides/openslides-client/pull/3869) is merged
   - This (this issue) new setting to allow anonymous users should be below locked_from_inside
      - it should be a checkbox 
      - the label (for now) should be "Allow anonymous users"
      - There should be a hint below this setting, the hint should be as follows:
         - If the anonymous group has no permissions at all: "The anonymous group will become visible at the url:  '##/participants/groups', please configure what your anonymous users are allowed to see"
         - If the anonymous group has one or more permissions: "The anonymous group is at url:  '##/participants/groups'. Your anonymous group is already allowed to see some of your contents, please configure the permissions according to your liking"
   - Locked_from_inside and make a meeting public should **NOT** be allowed to be true at the same time 
     - a locked meeting from the inside is not allowed to be public 
     - a public meeting is not allowed to be locked from the inside
     - If one checkbox is activated and the other is empty the empty one should be disabled, if both are (somehow) activated at the same time neither should be disabled
- [ ] Group-Matrix with the new anonymous group:
  - The group should only be visible if the setting "Allow anonymous users" is activated
  - Guest-group is only allowed to have reading-permissions, every writing permission should be off and disabled

**Additional context**
- we also have the new setting to lock a single participant out (https://github.com/OpenSlides/openslides-client/pull/3815). That setting can be active while a meeting is public --> these two settings are independent from each other

Elblinator:
For now the client will not display archived meetings if they are public and you are a guest

[...]

PR #3960, opened 2024-07-31
Read servertime from headers

### LINKED ISSUE #3503: Speaking Time Widget - wrong time visualization
**Description:** The time values are sometimes incorrectly displayed. This can have two different outcomes. A) Times are displayed in a 'NN:NN' style. B) Times are displayed with a high number value.

**Reproduction:** The reproduction is unclear. It seems to me, that the issue has a connection to a missing autoupdate, because this problem occurs sometimes in cases, were other menus (like for example participants) hadn't been opened. The crucial point is, that there is no clear reproduction.  

Error for 'NN:NN' happened again: Repro still unclear, but failure message appears  in console.
Cannot read properties of undefined (reading 'map')

**What should happen:**
The speaking times should always be visible in the correct time style Minutes:Seconds.

bastianjoel:
Instead of calling a servertime endpoint with this PR a request is sent to the server and the time gets obtained from the `Date` header. 

This is needed for #3955 

Very likely to resolve #3503


client/src/app/gateways/presenter/server-time-presenter.service.ts
```diff
@@ -66,13 +56,20 @@ export class ServerTimePresenterService {
      */
     private async refreshServertime(): Promise<void> {
         // servertime is the time in seconds.
-        const servertimeResponse = await this.presenter.call<ServerTimeResponse>(Presenter.SERVERTIME);
-        if (typeof servertimeResponse?.server_time !== `number`) {
```
bastianjoel, line 70: This could be the cause of https://github.com/OpenSlides/openslides-client/issues/3503 because the server time is used for the countdowns and NaN is actually a number and would therefore bypass this check.

Elblinator (approved):

bastianjoel:
@peb-adr Do you still want to have a look on this?

client/src/app/gateways/presenter/server-time-presenter.service.ts
```diff
@@ -66,13 +56,22 @@ export class ServerTimePresenterService {
...
+        const servertimeResponse = await fetch(`/assets/time.txt?${Date.now()}`, {
+            credentials: `omit`,
+            headers: {
+                'ngsw-bypass': `true`,
+                cache: `no-store`
+            }
+        });
+        const date = new Date(servertimeResponse.headers.get(`Date`));
+        if (servertimeResponse.headers.get(`Date`) && !isNaN(date.valueOf())) {
+            const serverDate = new Date(servertimeResponse.headers.get(`Date`));
```
peb-adr, line 68: Can this line not be

     const serverDate = date;

?

peb-adr (approved):
Minor nitpick (see comment).
But I approve.

With the changes in the fetching of `time.txt`, specifically the addition of `?${Date.now()}` I am now confident as well, that responses will not come from disk cache and therefore will indeed represent the servers current time.\
Until now that was my problem with this change.

bastianjoel:
Done. Also moved and renamed the service as it is not using a presenter anymore.

PR #3969, opened 2024-08-05
Fix amendment list performance

### LINKED ISSUE #3014: Motion list / amendment list: Performance Issue
**Description:** If a meeting contains a lot of long motions and complex amendments, the overall performace by opening the motion or amendment menu is slow. The menus (lists) load quite long for the first time. During the loading time, the user gets no respond what is happening. In the worst case, the user thinks the page won'T load at all. 

**Reproduction:** 
Not clear how much content needed to slow the meeting down. Just create for example 50 motions and several amendmend to them (at least 10 per Motion). Additionally, they have to be complex. Not only a small word or line change.

**What should happen:**
In general it would be quite good, if the overall performance of this lists could be improved. If this is not easy to handle, a loading bar should help. So the user could get a response that something is still loading and the page isn't down.

bastianjoel:
Fixes amendment list performance by just calculating diffs of amendments in viewport.

resolves #3014


luisa-beerboom (approved):

MSoeb (commented):
I've spoken with @emanuelschuetze about it the current state. And we want to talk you you about it. We'Ve found some good and not so good points, which could be improved. Also we've spoken about a heavy test with lots of data (amendment to one motion) and how well you PR can handle it and where are the limits. 

TLDR: Let's speak about it in the next weeks.

bastianjoel:
Is it possible that the response time for the amendment texts is the problem here then?

rrenkert (approved):

PR #3978, opened 2024-08-05
Add filters to poll single votes

### LINKED ISSUE #3899: Poll detail view: new filters and bar chart
**Current behavior**
In addition to a yes, no or abstention nominal vote, there may be various quorums that must be met additionally for some votes. For example, it may be the case that a) a majority from a certain present group or structure level must have voted with yes and b) a weighted majority from this pool must have voted with yes. Neither scenario can currently be visualized or easily read out of the data. Adjustments are required to make this possible.

**Wanted behavior**
The following features are required in the poll detail view to make it easier to recognize the necessary majorities in nominal votes on motions:

1. A filter with two filter options (by structure level and groups) should be added to single votes (See image: lower red mark)

2. Furthermore, a new horizontal bar chart should to be inserted above the table. 
The bar chart should be visible if one (or more) filters are activated
We want to use  form the that visualizes the (weighted) number of votes and percentages for yes/no/abstain vote based on the filtered list. Below the bar chart, the corresponding number of ballots cast (in other words the unweighted number of persons entitled to vote) for yes/no/abstention is shown. (see image: upper red mark)

With these features both desired quorums can be read off.

image

bastianjoel:
resolves #3899


Elblinator:
The filters have to be hidden in a non-nominal vote

The filters are currently displayed in ballots and topic polls as well but they are not working
They either have to work there as well or they have to hidden

please change my crude "cat.value" = "Y", "N", "A", to the actual words

bastianjoel:
9c2085066af5b5b8ddecb6eb8e2afc3399664d27 I don't get why you changed that. With that you calculate the value twice while also leaving two places in the template where you still check if the vote category is empty. 

Reverted for that reason.

Elblinator:
- if a  poll was resetted then the poll data can still wrongly be displayed
Repo: 
1. create, start, finish poll (you need data in the poll)
2. open the poll 
3. look at the new bar chart
4. via the arrow navigate back to the motion
5. reset the poll
6. navigate into the poll 
7. the bar is still displayed

- if a YN Method is activated then the Ballots cast string shows the Abstain option
@bastianjoel what do you think is the best way to prevent that? Change the Array or just hiding the last option?

emanuelschuetze (changes requested):
* Rename "Weighted votes" to "Valid votes"
* Hide line "Ballot cast: ..." if vote weight is disabled / not used in poll.
* Line "Weighted votes": add abstain to total, and add abstain value (_without_ percent!)
* Line "Ballot cast": add abstain value

Elblinator (approved):

PR #3979, opened 2024-08-06
Allow `YN` votes for motions

### LINKED ISSUE #3901: New voting style for motion polls
**Current behavior**
Currently, motion polls only offer the option to vote yes/no/abstain. This is insufficient, as there are scenarios in which a vote must only be done with Yes/No.

**Wanted behavior**
It should also be possible to vote on motion polls with a YES/NO option. Some adjustments are necessary here.
- The 'new vote' dialog should get a new select option called 'Poll options', where the user could chose between Yes/No/Abstain and Yes/No.
- If the new option Yes/No is chosen, than the selection "Yes/No/Abstain" in the '100% base' selection should be grayed out, because it can't be used. The rest of the selections should be possible to choose.
- All other parts of the dialog should normaly work and interact with the new type of poll options.

Example image of affected vote dialog
image

bastianjoel:
resolves #3901

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/services/poll.service/poll.service.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/services/poll.service/poll.service.ts
@@ -272,7 +272,7 @@ export abstract class PollService {
     }
 
     public getVoteTableKeys(poll: PollData): VotingResult[] {
-        return [
+        const keys: VotingResult[] = [
             {
                 vote: YES_KEY,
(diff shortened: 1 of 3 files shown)


Elblinator (changes requested):
The default voting method should be selectable in the settings
The new setting should be below "Default groups with voting rights

bastianjoel:
This would require a new meeting field `motion_poll_default_method`

Elblinator (approved):

PR #3991, opened 2024-08-09
Fix new file dialog

### LINKED ISSUE #3858: Fix new file dialog
**Current behaviour:**
If the "new file" dialog for a new motion is opened then the dialog is wrongly formatted:
Screenshot_20240716_143603

**Reproduction:**
1. start creating a new motion
2. add a new file
3. the dialog opens and looks like that

**Wanted behaviour:**
The dialog should be implemented in the same way our other dialogs are implemented

vkrasnovyd:
Resolve #3858 

1. Updated MDC in 'media-upload-content.component.html' similar to 'file-list.component.html' where file uploading card is displayed correctly.
2. Added paddings and set font size to 16px for '.mat-mdc-dialog-surface' container of the dialog (similar to the container on '/mediafiles/upload' page).

The changes above together made file upload dialog on the '/motions/new' page look the same as on the '/mediafiles/upload' page.

===
3. Additional ui improvement: made text of dropZoneLabel wrappable.
On the smaller screens file upload dialog on the '/motions/new' page is narrower than on the '/mediafiles/upload' page because of the styles of the rest of the page.
That made part of the sentense 'Drop files into this area OR click here to select files' invisible on small screens.
Moved upload area to the flex container and reduced line height of the sentense to 24px to let the text wrap correctly.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-view/motion-detail-view.component.scss
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/components/motion-detail-view/motion-detail-view.component.scss
@@ -172,3 +172,12 @@ button.motion-detail-nav-button {
     font-size: 16px;
 }
 
+.file-drop-content-style {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    
+    .ngx-file-drop__drop-zone-label {
+        line-height: 24px;
+    }
+}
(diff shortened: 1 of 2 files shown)


Elblinator (approved):

Elblinator:
I found two other places where this issue occurs
I'll create a new Issue for that though, because the dialog from the motion create was specifically mentioned int the issue

PR #3993, opened 2024-08-12
Fix list of speakers names not displayed

### LINKED ISSUE #3798: Autopilot > List of speakers: Missing names after reload

bastianjoel:
resolves #3798

This PR changes the `_as_observable` suffix to just `$` for view models to comply with angulars [naming conventions for observables](https://v17.angular.io/guide/observables#naming-conventions-for-observables). Besides from that the observable relations do not need to be specified manually anymore. Instead a new type `ViewModelRelations` is used. 

Another change included here is the change of the Shared worker type to `module` to support activating the shared worker while development. This ability was broken with switching to the new angular compiler.


MSoeb (changes requested):
Good: The names always loads correctly after several reloads.

Bad: The performance issue still occurs.

bastianjoel:
The performance issues are not cause by the client

client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/list-of-speakers-entry/list-of-speakers-entry.component.html
```diff
@@ -234,7 +235,7 @@
                 </div>
             </div>
             <ng-container *osPerms="permission.listOfSpeakersCanManage">
-                @if (speaker.speech_state === SpeechState.INTERPOSED_QUESTION && !speaker.meeting_user_id) {
```
luisa-beerboom, line 237: Bad. Revert this line.

luisa-beerboom (approved):
Nice

MSoeb (approved):
looks fine

PR #4015, opened 2024-08-13
Update duplicate from handling, templates, required

### LINKED ISSUE #3994: Allow committee admin to also duplicate meetings from their own committee
needs to work https://github.com/OpenSlides/openslides-backend/issues/2564
**Wanted behaviour:**
 if the setting require_duplicate_from is true and if the user is a committee admin
- then while creating a new meeting the duplicate from field should be indicated as a required field (there should be a star and it should be in warn colour if empty)
- committee-detail-view: the duplicate option should be visible and selectable in the three dot menu from a single meeting

reiterl:
Resolve #3994

Needs:

- https://github.com/OpenSlides/openslides-backend/pull/2568


reiterl:
I tried to add the color effect. But somehow it didn't work. It is not that important.
@bastianjoel can you review this ?

Elblinator (approved):

bastianjoel (approved):

PR #4025, opened 2024-08-15
Fix poll progress number

### LINKED ISSUE #3890: Voting: Progressbar issue with voting delegation
**Describe the bug**
The reference value of the maximum number of voters in the progress bar for votes or elections is incorrect if a vote delegation was previously switched on but was switched off in the settings before the vote was started. The transfer of voting rights is still considered active in the background, which means that the reference value is incorrect.

**How to Reproduce**
1. open meeting
2. open Settings > Participants > Activate vote delegation (main checkbox)
3. open Participants > Create 2 users with group Delegates (A and B)
4. transfer voting rights from user B to user
5. open Settings > Participants > Deactivate vote delegation
5. open participants > Add administrator as user to the event and set user A to present
6. applications > create application > create vote; define group delegates and admin as authorized voters. > Start voting.
7. vote with admin
8. vote with user A in another browser
9. check the progress bar as admin in autopilot. value should be 2/3 should. This value is incorrect, as the transfer of vote delegation is still included in the reference value.

**Expected behavior**
The reference value in the progress bar should be based on the number of voters currently present. Voting delegations should only be included if this function is activated.

Elblinator:
closes #3890

### DIFF
--- a/client/src/app/gateways/repositories/users/user-repository.service.ts
+++ b/client/src/app/gateways/repositories/users/user-repository.service.ts
@@ -3,8 +3,10 @@ import { Fqid } from 'src/app/domain/definitions/key-types';
 import { MeetingUser } from 'src/app/domain/models/meeting-users/meeting-user';
 import { BaseRepository } from 'src/app/gateways/repositories/base-repository';
 import { UserAction } from 'src/app/gateways/repositories/users/user-action';
+import { SubscriptionMap } from 'src/app/infrastructure/utils/subscription-map';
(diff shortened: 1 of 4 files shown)


bastianjoel (changes requested):
I would prefer a view model only solution. I am not exactly sure if you can access settings there directly. If not just set the method in the repository as done with `getFullName` for example.

MSoeb (approved):
Looks good

MSoeb (approved):
Looks good

bastianjoel (approved):

PR #4044, opened 2024-08-21
Fix linter missing content

Elblinator:
Adds aria Labels to the category detail, motion-block-detail, participant-list, account-list, committee-list, meeting-list, file-list

To verify that the aria-labels work properly navigate to the said pages and in the dev-tools > elements search for the strings (you can find the strings in the changed files

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -103,11 +103,7 @@ module.exports = {
                     alphabetical: true
                 }],
                 '@angular-eslint/template/prefer-control-flow': ['error'],
-                '@angular-eslint/template/elements-content': ['error', {
-                    "allowList": [
-                      "ariaLabel"
-                    ]
-                  }]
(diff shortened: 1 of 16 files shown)


client/.eslintrc.js
```diff
@@ -103,11 +103,7 @@ module.exports = {
                     alphabetical: true
                 }],
                 '@angular-eslint/template/prefer-control-flow': ['error'],
-                '@angular-eslint/template/elements-content': ['error', {
-                    "allowList": [
-                      "ariaLabel"
-                    ]
-                  }]
+                '@angular-eslint/template/elements-content': ['error']
```
bastianjoel, line 106: You can delete this line as the rule is already included in the accessibility rules template.

bastianjoel (changes requested):

bastianjoel (approved):

PR #4027, opened 2024-08-15
Customizable gender field

reiterl:
Resolve #3803 

Work in Progress, added internal logic and the list and dialog, the genders/gender fields still need to be replaced in the different code positions. 

Needs an *updated versions* of:[
* https://github.com/OpenSlides/openslides-autoupdate-service/pull/1025
* https://github.com/OpenSlides/openslides-backend/pull/2485

PR #4048, opened 2024-08-22
Import Pages - csv example file improvement

### LINKED ISSUE #3822: Import Pages - csv example file improvement
**Description:** Some improvements of the account and participant import csv example files are required. 

**What should be done:**

- [x] Participant import: Only one example line is needed. The csv content should look like the following image:
image

- [x] Account import: Only one example line is needed. The csv content should look like the following image:
image

- <s>If possible: Values in groups and gender should be translated in the csv example file when switching languages in the OpenSlides client. This does NOT affect the column name, only the content entered. (Example: male -> männlich, Delegates -> Delegierte). This would make it much easier for users to recognize, that they could use terms of their chosen OpenSlides client language.</s>

Info: This is part of the META Issue #3809

vkrasnovyd:
Resolves #3822 

1. Shortened the example csv for importing participants and accounts to 1 line, as was shown in the images.
2. Configured translating values in the selected columns to the current client language:
- in 'gender' column for account-example
- in 'gender' and 'groups' columns for participants-example

Notes in case at some stage example files are updated again:
- If example files are extended with extra lines, the translation will work for all the example lines. No further adjustments will be needed.
- Value in 'groups' column can be translated only if there's only one group name in a cell. Doesn't work for sample values like 'Delegates, Staff'.


bastianjoel (changes requested):
With custom translations it is possible to get a non injective set of genders. Therefore it might not be a good idea to allow users to use this. Especially with #4027 this will get much more complicated. 

For groups this is a bit different. Those are translated in the backend and therefore guaranteed to be unique. Still the meetings language instead of the users language needs to be used there but I think it might be even better to use an actually existing group name from the groups repo. 

@MSoeb please provide feedback on what you want here.

rrenkert:
> With custom translations it is possible to get a non injective set of genders. Therefore it might not be a good idea to allow users to use this. Especially with #4027 this will get much more complicated.
> 
> For groups this is a bit different. Those are translated in the backend and therefore guaranteed to be unique. Still the meetings language instead of the users language needs to be used there but I think it might be even better to use an actually existing group name from the groups repo.
> 
> @MSoeb please provide feedback on what you want here.

Talked to @MSoeb: We do not want client side translations for genders in the example file. For groups the group name exisiting in the groups repo is the correct value to use.

client/src/app/site/pages/meetings/pages/participants/export/participant-csv-export.service/participant-csv-export.service.ts
```diff
@@ -57,8 +58,41 @@ export class ParticipantCsvExportService {
...
+                groupsToExport = meeting.default_group.name;
+                break;
+            case 1:
+                groupsToExport = customGroupNames[0].name;
+                break;
+            default:
+                groupsToExport = customGroupNames
+                    .slice(0, 2)
+                    .map(group => group.name)
+                    .join(`, `);
```
bastianjoel, line 85: This can be simplified. 

`slice` will just use the length of the array when the end parameter exceeds its limits. For that reason `case 1` is not needed.

client/src/app/site/pages/meetings/pages/participants/export/participant-csv-export.service/participant-csv-export.service.ts
```diff
@@ -57,8 +58,41 @@ export class ParticipantCsvExportService {
...
 
-    public exportCsvExample(): void {
+    /**
+     * @param meeting
+     * @returns participants csv-example with added 'groups' value:
+     * - 2 custom group (not default or admin) names separated by comma by default
+     * - 1 custom group name if meeting has only 1 custom group
+     * - default group name if meeting has no custom groups
+     */
+    public addParticipantGroups(meeting: ViewMeeting): UserExport[] {
```
bastianjoel, line 68: Method should be `private`

client/src/app/site/pages/organization/pages/accounts/services/account-export.service/account-export.service.ts
bastianjoel, line 1: Please revert your changes here.

client/src/app/site/pages/meetings/pages/participants/pages/participant-import/services/participant-import.service/participant-import.service.ts
```diff
@@ -44,7 +44,7 @@ export class ParticipantImportService extends BaseBackendImportService {
     public constructor(
         importServiceCollector: ImportServiceCollectorService,
         private repo: ParticipantControllerService,
-        private activeMeetingIdService: ActiveMeetingIdService,
+        private activeMeeting: ActiveMeetingService,
```
bastianjoel, line 47: I would prefer importing `ActiveMeetingService` directly in `ParticipantCsvExportService` and keeping this file as it is.

bastianjoel (changes requested):

vkrasnovyd:
As we discussed with @rrenkert, when the custom gender feature is finalized, I'm going to extend the export files with the gender option from the organization repo. Until then the issue should remain open.

reiterl:
Ah @vkrasnovyd you need to assign @bastianjoel , if you want his review.

client/src/app/site/pages/meetings/pages/participants/export/participant-csv-export.service/participant-csv-export.service.ts
```diff
@@ -57,8 +60,37 @@ export class ParticipantCsvExportService {
         );
     }
 
-    public exportCsvExample(): void {
+    /**
+     * @param meeting
```
reiterl, line 64: I don't see the meeting param in the method header.

client/src/app/site/pages/meetings/pages/participants/export/participant-csv-export.service/participant-csv-export.service.ts
```diff
@@ -57,8 +60,37 @@ export class ParticipantCsvExportService {
...
+     * @returns participants csv-example with added 'groups' value:
+     * - 2 custom group (not default or admin) names separated by comma by default
+     * - 1 custom group name if meeting has only 1 custom group
+     * - default group name if meeting has no custom groups
+     */
+    private addParticipantGroups(): UserExport[] {
```
reiterl, line 70: Mmh, the name 'addParticipantGroups" is misleading, it should be like eg 'provideRows'...

client/src/app/site/pages/meetings/pages/participants/export/participant-csv-export.service/participant-csv-export.service.ts
```diff
@@ -57,8 +60,37 @@ export class ParticipantCsvExportService {
...
+        const meeting: ViewMeeting = this.activeMeeting.meeting;
         const rows: UserExport[] = participantsExportExample;
+        let groupsToExport;
```
reiterl, line 73: Missing type.

reiterl (commented):
Small remarks. Please wait for @bastianjoel review.

vkrasnovyd:
> Ah @vkrasnovyd you need to assign @bastianjoel , if you want his review.

Thanks. I may still need to add a couple more changes based on your custom gender implementation. So until it's finalized, I'll keep it assigned to myself.

vkrasnovyd:
As discussed with Raimund, I've changed value for the 'gender' column from string 'male' to the name of gender with id=0 in the organization repo.
**Reason:** to provide name of one of the default genders that were set for this specific organization.

As disscussed with Loki, **subscription needs to be updated** for this change to work properly.
**Current problem:** organization gender data is not available when I open '{base_url}/{meeting_id}/participants/import' directly and try to export the example. However, csv-example export works if I visit page with organization genders info (for example organization accounts list) before.
Csv-example export in the accounts import page works without such issues.

bastianjoel (approved):

client/src/app/site/pages/organization/organization.subscription.ts
```diff
@@ -32,7 +32,8 @@ export const getOrganizationSubscriptionConfig: SubscriptionConfigGenerator = ()
         additionalFields: [`committee_ids`, `organization_tag_ids`],
         follow: [
             { idField: `mediafile_ids`, fieldset: `organizationDetail` },
-            { idField: `theme_id`, fieldset: DEFAULT_FIELDSET }
+            { idField: `theme_id`, fieldset: DEFAULT_FIELDSET },
+            { idField: `gender_ids`, fieldset: [`name`] }
```
bastianjoel, line 36: It would be better to subscribe this on the account/participant list.

PR #4061, opened 2024-08-28
Propergate matSelect errors into base search selector

### LINKED ISSUE #4028: Meeting creation: UI bug
**Describe the bug**
If the 'Duplicate from* field is a mandatory field for committee administrators for meeting creation, the font and frame will not be displayed in the correct color.

Example Image:
image

**How to Reproduce**
Requirements: A committee admin, a committee, a template meeeting, Activated setting 'Public template required for creating new meeting' under orga settings.
1. open instance as committee admin
2. open committee
3. create a new meeting > check the 'Duplicate from' field.

**Expected behavior**
If no value is selected in the 'Duplicate from' field, the color should switch to the same style as the color of the rewuired field title. (Example in image above)

bastianjoel:
resolves #4028

### DIFF
--- a/client/src/app/ui/base/base-form-field-control.ts
+++ b/client/src/app/ui/base/base-form-field-control.ts
@@ -96,7 +96,9 @@ export abstract class BaseFormFieldControlComponent<T>
 
     public stateChanges = new Subject<void>();
 
-    public errorState = false;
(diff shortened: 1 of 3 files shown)


Elblinator (approved):
function

PR #4073, opened 2024-09-02
Add parsed data cleanup rule

### LINKED ISSUE #3823: User Import - small import improvement (Mail / gender recognition)
**Description**: When importing accounts or participants, there is always the problem that some attributes (e-mail and gender) can have spaces BEFORE and AFTER the value. This leads to problems during import. The conflict must be resolved manually.

Example of the csv file with additional spaces. First mail column has spaces before and after the mail. Second mail has no spaces:
image

**What should happen:**
It would be good if spaces BEFORE and AFTER certain attributes were automatically ignored when uploading the csv file, as these are not logically possible.

Affected attributes:
- E-mail
- Gender 

Info: Issue is part of META issue #3809

vkrasnovyd:
Resolves #3823 

Extended parseInput config with .trim() rule to remove extra whitespaces on the ends of the input values.

### DIFF
--- a/client/src/app/site/base/base-import.service/base-backend-import.service.ts
+++ b/client/src/app/site/base/base-import.service/base-backend-import.service.ts
@@ -176,7 +176,7 @@ export abstract class BaseBackendImportService implements BackendImportService {
             header: true,
             skipEmptyLines: `greedy`,
             quoteChar: this.textSeparator,
-            transform: value => (!value ? undefined : value.trim())
+            transform: value => (!value.trim() ? undefined : value.trim())
         };
         if (this.columnSeparator) {
             papaConfig.delimiter = this.columnSeparator;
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

client/src/app/site/base/base-import.service/base-backend-import.service.ts
```diff
@@ -176,7 +176,7 @@ export abstract class BaseBackendImportService implements BackendImportService {
             header: true,
             skipEmptyLines: `greedy`,
             quoteChar: this.textSeparator,
-            transform: value => (!value ? undefined : value)
+            transform: value => (!value ? undefined : value.trim())
```
bastianjoel, line 179: `value.trim()` could be false while `value` is true. Therefore you should also use `value.trim()` for the check.

bastianjoel (changes requested):

bastianjoel (approved):

PR #4124, opened 2024-09-16
Prevent null in calculated group_ids of assign meetings

### LINKED ISSUE #4085: Bulk action user.update add to meeting is throwing an Error
**Current behaviour:**
If you try to add a participant via bulk action/multi action then the following Error is thrown:  
```
Error: data.group_ids[1] must be integer
```

**Payload:**
```
[  {
    "action": "user.update",
    "data": [      {
        "id": 5097,
        "meeting_id": 101,
        "group_ids": [
          631,
          null  
        ]     
     }]
}]
```

**Response:**
```
{
    "success": false,
    "message": "data.group_ids[1] must be integer",
    "action_error_index": 0,
    "action_data_error_index": 0
}
```

**Wanted behaviour:**
It should be possible to add user via bulk action/multi select to one or more meetings,
if no group was specified then the default group should be used

reiterl:
Resolve #4085 

I don't understand, why the meeting.default_group_id is null in some rare cases, but it leads to an error, because it
sends null in group_ids. So I exclude broken stuff from this array.

### DIFF
--- a/client/src/app/site/pages/organization/pages/accounts/services/common/account-controller.service.ts
+++ b/client/src/app/site/pages/organization/pages/accounts/services/common/account-controller.service.ts
@@ -73,9 +73,10 @@ export class AccountControllerService extends BaseController<ViewUser, User> {
                 return {
                     id: user.id,
                     meeting_id: meeting.id,
-                    group_ids: groupIds?.includes(meeting.default_group_id)
+                    group_ids: (groupIds?.includes(meeting.default_group_id)
                         ? groupIds
                         : (groupIds ?? []).concat(meeting.default_group_id)
+                    ).filter(id => !!id)
                 };
             });
         };


Elblinator (approved):

client/src/app/site/pages/organization/pages/accounts/services/common/account-controller.service.ts
```diff
@@ -73,9 +73,10 @@ export class AccountControllerService extends BaseController<ViewUser, User> {
                 return {
                     id: user.id,
                     meeting_id: meeting.id,
-                    group_ids: groupIds?.includes(meeting.default_group_id)
+                    group_ids: (groupIds?.includes(meeting.default_group_id)
                         ? groupIds
                         : (groupIds ?? []).concat(meeting.default_group_id)
+                    ).filter(id => !!id)
```
bastianjoel, line 79: If there is a reason to always add the default group (which I guess there is) this looks like a very bad approach to "fix" this.

bastianjoel (changes requested):

reiterl:
This pr doesn't solve the underlying problem. So I close it here.

PR #4125, opened 2024-09-16
Fix time menu click area

### LINKED ISSUE #4009: Speaking Times Widget: Layout error
**Describe the bug**
The speaking time widget has a layout bug. The three-point menu and the associated click area no longer overlap correctly. After updating the mdc components, the three-dot menu is shifted to the right. This is clearly visible if you hover over the area with the cursor. 

Example of current state:
image

**How to Reproduce**
The bug can easily be spotted on a staging instance. Please contact me for details. 

**Expected behavior**
A) The three-dot menu and the click area should overlap correctly again. B) The menu should return to its original position to the right of the time value.

vkrasnovyd:
Resolves #4009 

Class 'icon-less-space-to-left' reduced the menu opening dots area width in order to fit all the times into the widget.

This commit restores this area width and reduces its right margin to make the click area overlap the dots correctly without increasing the time block width.


MSoeb (changes requested):
The fix needs some improvements and best a conversation about it. 

1. In the current state the fix leads to to the situation, that the mat cards left to the speaking times widget are smaller. This is not good. Especially on on smaller screen sizes, like tablets, this isn't user friendly. -> The mat card width should not be changed. Changes should only be made inside the cards.

2. The three-dot-menu is still not optimal places. The elements can't easily be dinstingushed if the strcuture levels have a color. It must ne easier to separate the logical elements groups from each other. -> A solution could be to cut off the color a bit. Emanuel and I think, that a time in the style -999:99 should be the max value. The color should be trimmed around this value.

3. The three-dot-menu is still no far from the number. Maybe the second CR can help, but I think, we should place is a bit nearer to the number even if this would lead to an overlapping of the elements.

MSoeb:
Please excuse me for not responding directly to your suggestions. I had only looked at the PR and based my CR on it. That was not correct.

About the implementation: The last picture looks good. The arrangement along a grid is a good idea. However, a few points would need to be considered.

- The overall width of the widget and the width of the other elements to the left of the widget must be maintained. The widget must not become larger, otherwise it will reduce the size of the other elements.

- In the widget itself, the color bar above the numbers should be shortened. The length of the bar is based on a maximum value for the speaking time. This is -999:59. The minus is required as an additional value, as the time can become negative. The aim is to separate and recognize the individual logical elements more clearly. - TLDR: Please shorten the color bar so that a number in the format -999:59 is covered. 

- The idea with the line break sounds good. Please implement it as in the last image. 


Context for the second key point: The three-dot menu appears detached from the associated element or, due to the length of the color strokes, it currently appears that this menu could also belong to another element. I hope that the above changes will be sufficient to make it easier to distinguish the elements that belong together.

bspekker (approved):

PR #4169, opened 2024-09-23
Add default_password to account_list subscription

### LINKED ISSUE #4145: Account/Participant export: Initial Password not always exported
**Describe the bug**
A strange error occurs in the account/participant export; the inital password of some accounts are not exported. I've tried to reproduce it, but I can't. If you need further details please contact me.

**Expected behavior**
Account and participant export should always export the inital password.

reiterl:
Resolve #4145

There is a problem with missing default pw on a prod instance. It seems like export needs these data for exporting it. So I included it in the subscription. The participant_list subscription already includes the "default_password" for this usage.

The autoupdate service needs to restrict the access, if these info should not be seen.

This enlarges the subscription with one field, perhaps it is a performance problem.


Elblinator (approved):
Bug only happens in the account-list 

export works for both lists now

bastianjoel (approved):

PR #4177, opened 2024-09-25
Fix Motion-list CSV export and add new Version to export

### LINKED ISSUE #4050: Fix CSV export from motion with change recommendation
**Describe the bug**
If a motion with change recommendation is exported with the original text then the change recommendation is also exported

**How to Reproduce**
1. Create a motion 
2. Add a change recommendation to the motion
3. Navigate to the motion-list
4. export the motion as CSV (only possible in the original version)
5. look at the exported motion
6. The Change recommendations were also exported

**Expected behavior**
If a motion is exported in the original version then the change recommendations should not be exported 

**Additional information:**
@emanuelschuetze would like to look at the export dialog together with the dev who deals with this bug

Elblinator:
closes #4050 

Added the editorial final version for the CSV export 
=> the PDF now has the two options "final Version" and "editorial final Version", This can be changed back to only one option (If that is wanted which version should be used?)
=> The CSV only has two options ("Original" and "editorial final" Version), are these the only two options wanted for the CSV export?

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/components/motion-export-dialog/components/motion-export-dialog/motion-export-dialog.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/components/motion-export-dialog/components/motion-export-dialog/motion-export-dialog.component.ts
@@ -129,15 +129,7 @@ export class MotionExportDialogComponent extends BaseUiComponent implements OnIn
     @ViewChild(MOTION_PDF_OPTIONS.ContinuousText)
     public continuousTextButton!: MatButtonToggle;
 
-    private _isCSVExport = false;
-
-    protected set isCSVExport(format: ExportFileFormat) {
-        this._isCSVExport = format === ExportFileFormat.CSV;
-    }
-
-    protected get isCSVExport(): boolean {
-        return this._isCSVExport;
-    }
+    public isCSVExport = false;
 
     /**
      * Constructor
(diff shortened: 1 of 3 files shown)


Elblinator:
After talking with @emanuelschuetze:
- The editorial final version was reverted
- Two options are neough for CSV export

- we found a bug, where the crMode was not reset correctly after switching from pdf to csv

client/src/app/site/pages/meetings/pages/motions/components/motion-export-dialog/components/motion-export-dialog/motion-export-dialog.component.ts
```diff
@@ -129,6 +129,16 @@ export class MotionExportDialogComponent extends BaseUiComponent implements OnIn
     @ViewChild(MOTION_PDF_OPTIONS.ContinuousText)
     public continuousTextButton!: MatButtonToggle;
 
+    private _isCSVExport = false;
+
+    protected set isCSVExport(format: ExportFileFormat) {
+        this._isCSVExport = format === ExportFileFormat.CSV;
+    }
```
bastianjoel, line 136: This is weird behavior. Just do the check on assignment. 

Return and assignment values of getter/setters should always be the same.

bastianjoel (changes requested):

bastianjoel (approved):

rrenkert (approved):

PR #4178, opened 2024-09-25
Amount participants visibility in committee/meeting

### LINKED ISSUE #4071: Meeting/Committee menu: Visibility of participants not correctly restricted
**Describe the bug**
In the committees and meetings menu, the visibility of existing participants in a meeting is not correctly regulated in the tool tips. Users who are not allowed to see participants, provided by the group permission in a meeting, can always see the number of participants per meeting in the menus. This is incorrect. The visibility depends on whether a user can see the participants menu in a meeting.

Additionally, the number of accounts in a committee is visible for users who are not admins in the committees listing.

Image 1:
image

Image 2:
image

Image 3:
image

**How to Reproduce**
Requirements: A committee with two meetings. A user as delegate in both meetings who can't see the participant list restricted by group permission.
1. login as user A > open committees menu. Number of accounts per committee is visible. (Image 1 - case 1) > Open the committee detail view. you should see now both meetings with a hint about the user amount in the meeting. (image two - caase 2)
2. open now the meeting list. same behavior here. the user can see the amount of participants in the meetings. (image three - case three)

**Expected behavior**
Case 1: The number of accounts per committee should only be visible for accounts who are allowed to see it. These are Superadmins, Organizationadmins, Account admins and committee admins .

Case 2: Accounts without the permission to see users in a meeting should not see the participant amount. Superadmins, orgaadmins and committee admins should see it. 

Case 3:  The number of accounts per meeting should only be visible for accounts who are allowed to see it. These are Superadmins, Organizationadmins, Account admins and committee admins.

reiterl:
Resolve #4071

### DIFF
--- a/client/src/app/site/pages/organization/organization.subscription.ts
+++ b/client/src/app/site/pages/organization/organization.subscription.ts
@@ -58,7 +58,8 @@ function getMeetingListFollowConfig(
         idField: idField,
         follow: [
             { idField: `committee_id`, fieldset: `name` },
-            { idField: `organization_tag_ids`, fieldset: FULL_FIELDSET }
+            { idField: `organization_tag_ids`, fieldset: FULL_FIELDSET },
+            { idField: `group_ids`, fieldset: [`name`, `permissions`] }
         ],
         fieldset: `list`
     };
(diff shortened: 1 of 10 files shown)


MSoeb (approved):

bastianjoel (commented):
I am against this change. 

The info is still sent to the client and therefore IMHO this issue is not fixed. It would be better to implement https://github.com/OpenSlides/openslides-client/issues/2107 instead.

bastianjoel:
@rrenkert Can we close this?

PR #4180, opened 2024-09-25
Adjust motion poll appearance to YN motion poll method layout

### LINKED ISSUE #4019: Adjust motion poll appearance to ```YN``` motion poll method
**Current behavior**
With the Issue https://github.com/OpenSlides/openslides-client/pull/3979 we acquired the option to only have two options (Yes and No) in a motion poll. 
But the layout is statically made for three options

**Screenshot from a motion poll in the autopilot:**
Screenshot_20240813_163947

**Screenshot from a motion poll in the motion detail view:**
Screenshot_20240813_164103

**Wanted behavior**
In the autopilot:
If a motion poll has the PollMethod.YN then the layout from the buttons should behave like a the buttons from a ballot if it is called as PollMethod.YN
**Screenshot from a ballot in PollMethod.YN:**
Screenshot_20240813_164206

In the detail view: 
If a motion  has the PollMethod.YN then the two buttons should have the same distance from the left and right boarder. The buttons are currently called with the the third option, this option should not be there if it is PollMethod.YN

Elblinator:
closes #4019

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.html
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.html
@@ -65,9 +65,12 @@
                             <div
                                 [ngClass]="{
                                     'yna-grid':
-                                        poll.isMethodYNA || (settings.isSplitSingleOption && displayed_in_autopilot),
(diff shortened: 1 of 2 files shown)


reiterl (approved):
LOTM

MSoeb (approved):

Issue #4159, opened 2024-09-20
Motion with amendments projection - paragraphs are displayed double


MSoeb:
**Current behavior**
The current state of the motion with amendments projection has improvement potential. If a motion has several amendments, with changed in different paragraphs, the projection will show at first the first amendment with all it's changes and than the second and so on. This leads to the strange situation, that the projection of the content start by line 1 till the end of the first amendment and start again by line 1 with the second amendment. Addtitionally the projection doesn't show the number of the amendment nor if amendments are in conflict. This needs to be improved, because in complexe amendment scenarios it is hard to distinguish in the projection which amendment is now shown and it's irritating to see the prjection start again with line 1.

**Wanted behavior**
The projection should follow the same layout logic like the motion detail view. There it is easy to spot amendment conflicts and information about the current shown number of the amendment. 

**Additional information:**
This occurance that the whole motion is shown and not paragraph by paragraph sometimes happens and sometimes is doesn't. I always needed a motion with 3+ paragraphs and creating an amendment which changes several paragraphs was a good way to recreate the problem

Elblinator:
After talking about the concept:
- General motion/amendments projection: the projection should be paragraph driven and not amendment driven, meaning that e.g. the first _paragraph_ should be projected as motion/amendment 1/ amendment 2/ etc. and then the second, third .... 
- the same warn icon should be projected as it is in the motion-detail
- The amendment nr should be projected in the same line as the icon (if there is no nr then only the icon is projected
- The amendment status should not be projected
- The icon and Nr should be indented in the same way as they are displayed in the motion-detail
- If an amendment has no icon and no Nr then the word "Amendment" should be projected 
- If reasonably possible in this issue the submitter from the amendment should be behind the Nr

Elblinator:
For the bug part I'll link the Meta-Issue https://github.com/OpenSlides/openslides-client/issues/2956 with this Issue 
And update the title

MSoeb:
Please don't work on it, until @Elblinator has added a clear reproduction.

Elblinator:
This specific bug is not reproducable anymore 
I'll close this Issue, if the same or a very similar bug reappears please create a new Issue

PR #4199, opened 2024-10-01
Motion with amendments add icon and amendment number

Elblinator:
closes part of #4159 

>- the same warn icon should be projected as it is in the motion-detail
>- The amendment nr should be projected in the same line as the icon (if there is no nr then only the icon is projected
>- The amendment status should not be projected
>- The icon and Nr should be indented in the same way as they are displayed in the motion-detail
>- If an amendment has no icon and no Nr then the word "Amendment" should be projected 
>- If reasonably possible in this issue the submitter from the amendment should be behind the Nr

The first part is not part of an enhancement but a bug in the motion projection


Elblinator:
Notes: 
bottom:12px is okay if the first line of a paragraph is changed

bottom needs to be bigger if a line inside of a paragraph is changed

bastianjoel (approved):

bspekker (changes requested):
Two superfluous line numbers in projector: 
1. Additional line number with empty line appears for first amendment if first amendment only changes up to line x in the motion and second amendment makes changes from line x+1 onwards.
image

2. Line number overlaps with conflict symbol if one amendment spans several paragraphs but the other one does not with the conflict only appearing in the second paragraph.
image

Elblinator:
I removed the doubled line numbers

bspekker (changes requested):
Amendment numbers on projector use different font for italics than in the motion. This causes the colon to not be highlighted in gray completely.

Indication of change recommendations are missing. Sould say "Change recommendation" like it does for amendments without number and symbol.

bspekker (approved):

PR #4206, opened 2024-10-02
Update to the tab navigation of the editor

### LINKED ISSUE #4150: Improve editor controls tab navigation
Currently when navigation with `<tab>` through a form with an `os-editor` one has to navigate over all format control buttons. 

**Wanted behavior**
When navigation with tab only the first format control should be focused. When hitting tab again the editors content should be focused. 
The other format buttons should still be reachable by using the keyboards arrow keys. 

The implementation should follow https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/

reiterl:
Resolve #4150 

Exclude:
 - menu buttons, because here we would need some more logic (ArrowUp,ArrowDown?)

Include: The change of the tabindex at arrow navigation.


reiterl:
I talked to @bastianjoel . The implementation is not complete and should be improved using Attribute-Directives.

bastianjoel:
I don't think this is much better than before either. I will have a look at this when there is time.

bastianjoel (approved):
I think we should just merge this and create an issue to refactor the editor menu. I won't have time for this in the near future. 

@Elblinator Please test if everything works as intended.

Elblinator (commented):
There is a glaring mistake: 
If you undo/redo sth and then the arrow is greyed out then the fields become inaccessable and the whole section is skipped via tab

Then there is another mistake:
The arrow behind the "format_color_text" and "format_color_fill" is inaccessible

@bastianjoel are these Issues to be fixed now or later?

bastianjoel:
@reiterl Please check if the mentioned issues are easily fixable and if so fix them.

Otherwise I would prefer closing this for now.

reiterl:
There was a problem with activating the color choosers with a key pressed, no idea how to fix that. 
As this is a suboptimal solution anyway, I close this and remove the branch.

PR #4247, opened 2024-10-11
Keycloak Auth2/OIDC

boehlke:


client/src/app/site/services/auth.service.ts
```diff
@@ -64,38 +57,44 @@ export class AuthService {
...
+                    break;
             }
         });
     }
 
+    public async checkOnline(): Promise<boolean> {
+        console.log(`auth: Do health check`);
+        let online: boolean;
+        try {
+            await fetch(`system/autoupdate/health`);
```
bastianjoel, line 72: Use http service.

client/src/app/site/services/auth.service.ts
```diff
@@ -64,38 +57,44 @@ export class AuthService {
...
+                case `check-auth`:
+                    console.log(`auth: Check auth`);
+                    this.oauthService.silentRefresh();
+                    break;
             }
         });
     }
 
+    public async checkOnline(): Promise<boolean> {
+        console.log(`auth: Do health check`);
```
bastianjoel, line 69:
```suggestion
        console.debug(`auth: Do health check`);
```

client/src/app/site/services/auth.service.ts
```diff
@@ -64,38 +57,44 @@ export class AuthService {
                 case `new-token`:
                     this.authTokenService.setRawAccessToken(msg.content?.token);
                     break;
+                case `check-auth`:
+                    console.log(`auth: Check auth`);
```
bastianjoel, line 61:
```suggestion
```

client/src/app/site/services/auth.service.ts
```diff
@@ -64,38 +57,44 @@ export class AuthService {
...
+    public async checkOnline(): Promise<boolean> {
+        console.log(`auth: Do health check`);
+        let online: boolean;
+        try {
+            await fetch(`system/autoupdate/health`);
+            online = true;
+        } catch (e) {
+            online = false;
+        }
+        console.log(`auth: health check done, online:`, online, `authenticated:`, !!this.authTokenService.accessToken);
```
bastianjoel, line 77:
```suggestion
        console.debug(`auth: health check done, online:`, online, `authenticated:`, !!this.authTokenService.accessToken);
```

client/src/app/site/services/auth.service.ts
```diff
@@ -110,59 +109,161 @@ export class AuthService {
...
-                this.sharedWorker.sendMessage(`auth`, { action: `update` });
+                // sessionStorage.clear();
+                await this.sharedWorker.sendMessage(<AuthUpdateMessage>{
+                    receiver: `auth`,
+                    msg: { action: `update`, params: null }
+                });
                 this.DS.deleteCollections(...this.DS.getCollections());
                 await this.DS.clear();
                 this.lifecycleService.bootup();
+                debugger;
```
bastianjoel, line 127:
```suggestion
```

client/src/app/site/services/auth.service.ts
```diff
@@ -110,59 +109,161 @@ export class AuthService {
...
+        this.oauthService.logOut();
     }
 
     public isAuthenticated(): boolean {
         return !!this.authTokenService.accessToken;
     }
 
+    public async startOidcWorkflow(): Promise<void> {
+        this.oauthService.events.subscribe(async event => {
+            console.log(event.type);
```
bastianjoel, line 168:
```suggestion
```

client/src/app/site/services/auth.service.ts
```diff
@@ -110,59 +109,161 @@ export class AuthService {
...
-        this.sharedWorker.sendMessage(`auth`, { action: `update` });
+        this.sharedWorker.sendMessage(<AuthUpdateMessage>{
+            receiver: `auth`,
+            msg: { action: `update`, params: null }
+        });
         this.DS.deleteCollections(...this.DS.getCollections());
         await this.DS.clear();
         this.lifecycleService.bootup();
+        // might do a redirect to the logout URL if configured
+        this.oauthService.logOut();
```
bastianjoel, line 159: Are you sure this should happen after the bootup?

client/src/app/site/services/openslides-router.service.ts
```diff
@@ -93,6 +90,7 @@ export class OpenSlidesRouterService {
     }
 
     public navigateToLogin(): void {
+        debugger;
```
bastianjoel, line 93:
```suggestion
```

client/src/app/site/services/openslides-router.service.ts
```diff
@@ -63,9 +63,6 @@ export class OpenSlidesRouterService {
         private updateService: UpdateService,
         private operator: OperatorService
     ) {
-        _auth.logoutObservable.subscribe(() => {
```
bastianjoel, line 66: Also remove from constructor when unused.

client/src/app/site/services/openslides.service.ts
```diff
@@ -25,19 +25,15 @@ export class OpenSlidesService {
      * {@method afterLoginBootup}. If not, redirect the user to the login page.
      */
     public async bootup(): Promise<void> {
-        const online = await this.authService.doWhoAmIRequest();
+        const online = await this.authService.checkOnline();
```
bastianjoel, line 28: If I understand correctly this now always does a request to a health endpoint on bootup. I think that would be unnecessary.

client/src/app/site/site-routing.module.ts
```diff
@@ -7,7 +7,7 @@ import { SiteWrapperComponent } from './modules/site-wrapper/components/site-wra
 
 const routes: Routes = [
     {
-        path: `login`,
+        path: `idp`,
```
bastianjoel, line 10: We might want to add a redirect if we really want to change this.

client/src/app/worker/sw-auth.interfaces.ts
bastianjoel, line 1: Move to a `definitions` folder and rename to `auth.ts`. Same for the other interfaces in this folder.

bastianjoel (changes requested):
As mentioned in one of the CRs please revert `3b1391e`.

This review does not include the worker part as I would like to see the tests passing before I have a look at it. Passing the "build-and-test-dev-image" pipeline is a hard requirement before merging to feature branches. 

Besides from that please fix the linter issues. You can use `npm run cleanup` to do this. After reverting `3b1391e` this should work without manual intervention.

client/src/app/openslides-main-module/services/shared-worker.service.ts
```diff
@@ -71,10 +71,10 @@ export class SharedWorkerService {
      * @param receiver Name of the receiver
      * @param msg Content of the message
      */
-    public async sendMessage<T extends WorkerMessageContent>(receiver: string, msg: T): Promise<void> {
+    public async sendMessage<C extends WorkerMessageContent, T extends WorkerMessage<C>>(msg: T): Promise<void> {
```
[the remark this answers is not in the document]
boehlke: T is now to message type, not only the content type.
So msg contains the receiver. This improves type safety when using the method...

client/src/app/gateways/http.service.ts
```diff
@@ -219,4 +223,29 @@ export class HttpService {
...
+            // Create a blob URL and cache it
+            const blobUrl = URL.createObjectURL(blob);
+            this.blobCache.set(url, blobUrl);
+
+            return blobUrl;
+        } catch (error) {
+            console.error('Error loading image:', error);
+            return null;
+        }
+    }
```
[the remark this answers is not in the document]
boehlke: The main perpose is getting the blob as an object URL. The fetch call might return a cached file but still response.blob() will create a new in-memory representation of the file.

client/src/app/site/pages/login/pages/login-mask/components/login-mask/login-mask.component.ts
```diff
@@ -167,23 +172,68 @@ export class LoginMaskComponent extends BaseMeetingComponent implements OnInit,
...
+            if (url) {
+                const { username, password } = this.formatLoginInputValues(this.loginForm.value);
+                const formData = new FormData();
+                formData.append(`username`, username);
+                formData.append(`password`, password);
+
+                const response = await fetch(url, {
+                    method: `POST`,
+                    body: formData
+                });
```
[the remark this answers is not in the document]
boehlke: I need a more low-level access to the response object that the http service does not provide. Also, I want to send a request that is processed by the app's interceptors...

client/src/app/openslides-main-module/services/shared-worker.service.ts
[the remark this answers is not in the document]
bastianjoel: I don't see how this improves type safety. Yes, you are not able to send messages to a wrong receiver with this but I think as the message type normally contains the receiver name it is normally quite clear which receiver to send it to. 

Even if we want to do something like this I think there are much better ways to achieve something like this by for example extending the shared worker service or providing methods to generate such messages.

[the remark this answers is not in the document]
boehlke: "Yes, you are not able to send messages to a wrong receiver" ... that is how type safety is improved, and I think you strengthen my point when write, that it's "...normally quite clear...". The compiler can enforce clearity.

This solution is not ideal, but an improvement, I think. Let's discuss a better solution.

client/src/app/site/pages/login/pages/login-mask/components/login-mask/login-mask.component.ts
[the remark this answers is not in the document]
bastianjoel: As far as I know you could register a second http provider if you just want to skip interceptors. I would prefer this.

client/src/app/openslides-main-module/services/shared-worker.service.ts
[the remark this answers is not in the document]
bastianjoel: IMHO the added redundancy overweights. I would prefer not doing this in this PR.

client/src/app/gateways/http.service.ts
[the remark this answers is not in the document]
bastianjoel: https://github.com/OpenSlides/openslides-client/pull/2062/files#diff-2b989788fc404d644fa890e422bb5a46f8173007f094b257fe5da603bcf1407fR14

boehlke:
Superseded by #5840

PR #4270, opened 2024-10-16
Add word-break style to keep width in account-detail view

### LINKED ISSUE #3990: Account detail view: issue with long committee and meeting names
**Describe the bug**
In the account detail view, there is a faulty layout where the group column on the right extends beyond the background frame.

It is due to too long, single words that are entered for meeting or committee names. These words are not wrapped and create a widening of the columns to the right over the background frame.

I don't know whether, for example, the number in the column headers are still relevant for widening of the columns. Please check this additionally. 

image

Reproduction:
1. open an instance
2. create a committee and a meeting with names that have 40 letters. Set yourself as admin.
3. open Accounts > open the detail view of your account.

**Expected behavior**
Under no circumstances should the columns extend beyond the background frame to the right. In case of doubt, words should be artificially wrapped so that the column width can be maintained.

reiterl:
Resolve #3990


bastianjoel (approved):

client/src/app/site/pages/organization/pages/accounts/pages/account-detail/components/account-detail/account-detail.component.scss
```diff
@@ -38,4 +38,8 @@ os-user-detail-view h2 {
     .committee-underline {
         border-bottom: 1px solid rgba(153, 153, 153, 0.5);
     }
+
+    .word-break {
+        word-break: break-all;
```
Elblinator, line 43:
```suggestion
        word-break: break-word;
```

Elblinator (changes requested):

client/src/app/site/pages/organization/pages/accounts/pages/account-detail/components/account-detail/account-detail.component.scss
reiterl: done.

PR #4280, opened 2024-10-21
Fix operator reset and wait until they're loaded

### LINKED ISSUE #4045: Login-Screen: unclear error message in console
**Describe the bug**
If you reload the login screen, an error message is displayed in the console.

image

**How to Reproduce**
1. open an instance > stay at the login page without login in
2. open the console
3. reload the page

**Expected behavior**
Console error should not appear.

### LINKED ISSUE #4108: Operator not fully reset after logout
**Describe the bug**
After logout the `operator.service.ts` is not completely reset resulting in issues when switching users. 
For example in #3965 the Dashboard menu point does not disappear when logging in as anonymous after a user with multiple meetings logged out. 
In #3815 another issue is described here: https://github.com/OpenSlides/openslides-client/pull/3815#issuecomment-2345918894

The main problem seems, that `_userSubject` does not get reset on logout. Other properties which do not get reset and could cause issues are `_lastUserId` and `_lastActiveMeetingId`. 

Resetting `_userSubject` will require some adjustments in other places as the `user` property of the operator service throws an error when accessing it while `_userSubject` is empty.

Elblinator:
closes #4108 
closes #4045

### DIFF
--- a/client/src/app/site/services/operator.service.ts
+++ b/client/src/app/site/services/operator.service.ts
@@ -100,13 +100,19 @@ export class OperatorService {
     }
 
     public get isAnyManager(): boolean {
-        return this.isSuperAdmin || this.isOrgaManager || this.isCommitteeManager;
+        return this.isSuperAdmin || this.isOrgaManager || this.readyDeferred.wasResolved
+            ? this.isCommitteeManager
+            : false;
     }
 
     public get knowsMultipleMeetings(): boolean {
         return (
             this.isAnyManager ||
-            (this.isAnonymous ? this.defaultAnonUser.hasMultipleMeetings : this.user.hasMultipleMeetings)
+            (this.isAnonymous
+                ? this.defaultAnonUser.hasMultipleMeetings
+                : this.readyDeferred.wasResolved
+                  ? this.user?.hasMultipleMeetings
+                  : false)
         );
     }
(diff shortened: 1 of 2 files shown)


client/src/app/site/services/operator.service.ts
```diff
@@ -436,7 +442,7 @@ export class OperatorService {
                 this._permissions !== undefined;
         }
         if (this.isAuthenticated) {
-            isReady = isReady && this._OML !== undefined && this._CML !== undefined;
+            isReady = isReady && this._OML !== undefined && this._CML !== undefined && this._userSubject !== undefined;
```
luisa-beerboom, line 445: this._userSubject is never going to be undefined, this._userSubject.value may be.

luisa-beerboom (changes requested):

luisa-beerboom (approved):

MSoeb (changes requested):
There is a problem with existing public access meetings: The meetings can't be opened. I could bring them back to life by setting another meeting as public access and grant viewing permissions to the public access group. After that, the other existing public access meeting can be opened.   

Not sure, if this is needs a change, because we are in test phase and this won't have a negative large impact.

MSoeb (approved):

PR #4284, opened 2024-10-22
Fix amendment wizard routing

### LINKED ISSUE #4275: Fix amendment creation
**Describe the bug**
If a user can only create amendments but not create motion then they are redirected to a wrong page if they try to create an amendment

**How to Reproduce**
1. Create a group which can **NOT** create motions but can create amendments
2. Login as someone from this group
3. (if you used the same browser reload the page with a hard reload)
4. navigate to a motion 
5. Press the plus button to create an amendment
6. you are redirected to a page you are allowed to see 

**Expected behavior**
The page which creates amendments should open and the user should be able to create amendments normally

luisa-beerboom:
Closes #4275


Elblinator (approved):
Code review

MSoeb (approved):
From a functional perspective it works as it should. But I have one small point which could be relevant: Creating (saving) an amendment as a delegate takes longer than as superadmin. I'm not sure, if this is because of the restrictions that limits the actions of a delegate.

PR #4286, opened 2024-10-22
Fix motion create form

### LINKED ISSUE #4285: Delegate unable to create a new motion
**Describe the bug**
As a delegate, I cannot create a new motion on OpenSlides. I triple checked, and the delegates have the permission to create new motions. I get an error message: 

"Error: You are not allowed to perform action motion.create. Forbidden fields: block_id, submitter_ids, tag_ids, agenda_parent_id"

**How to Reproduce**
Steps to reproduce the behavior: preferably on the [demo](https://demo.os4.openslides.com/)

1. Go to a Meeting
2. Click on Motions
3. Add a new motion and click save, but get an error and not being able to add the motion 
4. "Error: You are not allowed to perform action motion.create. Forbidden fields: block_id, submitter_ids, tag_ids, agenda_parent_id"

**Expected behavior**
Being able to add a new motion as a delegate. 

**System information**

- OpenSlides version: [4.1.18]
- Additional version information: [2024-10-17]
- Device: [PC]
- OS: [Windows 11 Pro]
- Browsers: [Chrome, Firefox, Edge]
- Browser versions: [Chrome 130.0.6723.59, Firefox 131.0.3, Edge 130.0.2849.46]

**Screenshot**
OS_bug1

Elblinator:
closes #4285

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-form/components/motion-form/motion-form.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-form/components/motion-form/motion-form.component.ts
@@ -211,12 +211,12 @@ export class MotionFormComponent extends BaseMeetingComponent implements OnInit 
     public saveMotion(event?: any): () => Promise<void> {
         return async () => {
             const update = event || this.temporaryMotion;
-            if (this.newMotion) {
-                for (const key in update) {
-                    if (update[key] === null || update[key].length === 0) {
-                        delete update[key];
-                    }
+            for (const key in update) {
+                if (update[key] === null || update[key].length === 0) {
+                    delete update[key];
                 }
+            }
+            if (this.newMotion) {
                 await this.createMotion(update);
             } else {
                 await this.updateMotion(update, this.motion);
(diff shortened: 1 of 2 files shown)


client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-form/components/motion-form/motion-form.component.ts
```diff
@@ -207,6 +211,11 @@ export class MotionFormComponent extends BaseMeetingComponent implements OnInit
     public saveMotion(event?: any): () => Promise<void> {
         return async () => {
             const update = event || this.temporaryMotion;
+            for (const key in update) {
+                if (update[key] === null || update[key].length === 0) {
+                    delete update[key];
+                }
```
luisa-beerboom, line 217: Probably not a good idea: null's the value we use to signify that a value is to be deleted, removing all keys with that value is going to remove that functionality. Also you can't guarantee that every key is an array.

What I think you really want to do, is removing certain keys depending on whether the user is even allowed to set them according to his permissions

luisa-beerboom (changes requested):

MSoeb (approved):
From a functional perspective it works.

client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-form/components/motion-form/motion-form.component.ts
Elblinator: for the motion.create it is enough to remove every null/empty array from the payload. The permissions do not have to be verified, given that a user can only send null/[] for the fields they are not allowed to change

For the motion.update it is wrong to remove empty arrays or null values

luisa-beerboom (approved):

PR #4287, opened 2024-10-24
Delgations and vote weigh optimizations

### LINKED ISSUE #4154: Vote delegation / vote weight - several optimizations
The vote delegation and vote weight feature needs some optimizations to be more user friendly. I will list the individual requirements below.

**A) User Detail view lacks in a good overview layout, if a user has several vote delegations.** Example image: The user has more than three vote delegations, but after three it is not possible to read them. 
image
--> Solution: Voting delegations are shown in a numbered list below each other. 

**B) Participant list: The total amount of vote delegations is not readable.** 
--> Solution: The suitcase (red mark) should get a number with the total amount of vote delegations per user. 
Image example which suitcase is meant: 
image

**C) Participant list: If a user has vote delegations, the total amount of vote weight is missing.**
--> Solution: The gray vote weight information should be enhanced with information about the total amount of vote weights given per delegation. Example line "Vote weight: 1 plus (amount) voting weight from transferred votes."
Example which line is meant:
image

**D) Participant list > Add new Filter 'Entitled to vote'.** 
The options for this filter should be 1.) 'is entitled to vote' and  2.)' is currently entitled to vote'. 

**E) Enhance information text in Participant list >  gray information text if filter in D is active.** 
If the new filter of D is active, there should be an information about how many rights to vote are given and the accumulated value for vote weights. This values should also change, if multiselection is used. The gray help text on the left side to filter and sort options should display' X of Y (Rights to vote 'value', accumulated vote weight 'value')'. Please take care, that this interacts with vote delegations. If a user has a vote delegation with a vote weight, this should also be accumulated. 
Note: This information should only be displayed for the new filter. For all other scenarios the information should not be visible.
 
**F) Visualization of vote weight in participant list needs to be improved and new help text for vote weight setting is needed**
Current status: All person are displayed with vote weight 1, if the vote weight option in activated. This is irritating and not helpful, because not all users are entitled to vote.
Solution: The vote weight in the participant list should only be displayed for users which are in an entitled to vote group given by the settings of agenda, motions and elections. For all other users, which are not in a corresponding group, the vote weight should not be displayed. Additionally the vote weight setting should get a new help text: "The vote weight will only be shown in the participant list for participants which are in a default vote group".

Elblinator:
fixes A, B, C and F from  #4154


luisa-beerboom (approved):

bspekker (dismissed):
A: list is not retracted by default and overlaps with menu. Add number of principals in participant details: "Voting right received from # principals:"

C: add label: Vote weight: 3 + 1 from delegated votes

MSoeb (changes requested):
Small change required:

The drop-down menu for voting delegations in the user detail view overlaps with letters such as g or p. These letters are cut off at the bottom.

PR #4307, opened 2024-10-30
Create two new participant filters

luisa-beerboom:
See #4154 

Also ensured the poll progress bar would consider the delegation settings => Please test.

@MSoeb I didn't really put much thought into the phrasing in the interface. @Elblinator I know I put you as a code reviewer, but if there is necessary changes in terms of phrasing, could you perhaps make those, idk what would be most understandable to the users.

Should probably be tested for performance as well

### DIFF
--- a/client/src/app/gateways/repositories/users/user-repository.service.ts
+++ b/client/src/app/gateways/repositories/users/user-repository.service.ts
@@ -364,6 +364,9 @@ export class UserRepositoryService extends BaseRepository<ViewUser, User> {
         viewModel.getEnsuredActiveMeetingId = (): Id => this.activeMeetingIdService.meetingId;
         viewModel.getDelegationSettingEnabled = (): boolean =>
             this.meetingSettingsService.instant(`users_enable_vote_delegations`);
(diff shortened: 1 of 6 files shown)


Elblinator (approved):
Code review

bspekker (changes requested):
1. When filter "Can vote for groups" is set, "For groups: total votes" should not count person that is delegated to vote but not in the eligible voting group. (Delegated vote to person in a non-vote-eligible group is currently counted double.)
2. New filters should disappear when setting "activate vote delegations" is not toggled in settings
3. Strings need changing, is still in discussion.

bspekker (dismissed):
As discussed, delete countable vote filter.

Adjust wording:
  - "Can vote for groups" to: "Voting rights" (German "Stimmrechte")
  - "(Vote weight 8; For groups: Total votes 5, Vote weight 11)" to: "(Voting rights 5, Vote weight 11) (delete first vote weight) (German: (Stimmrechte 5, Stimmgewichte 11) )

MSoeb (approved):

PR #4310, opened 2024-10-30
Add default field to subscriptions

### LINKED ISSUE #4236: Clean Non explicit use of default fieldset warnings
As I navigate meeting-list and committee-list on the dashboard (with Firefox) I see
` Non explicit use of default fieldset` warnings in the console.

Perhaps these warnings can also be addressed.

Elblinator:
closes #4236


reiterl:
I couldn't find the reason of the test failure, too.

PR #4325, opened 2024-11-07
Add default field to subscriptions and prettify subscriptions

Elblinator:
unline the PR #4236

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/motions.subscription.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/motions.subscription.ts
@@ -135,6 +135,7 @@ export const getMotionsSubmodelSubscriptionConfig: SubscriptionConfigGenerator =
     modelRequest: {
         viewModelCtor: ViewMeeting,
         ids: [id],
+        fieldset: DEFAULT_FIELDSET,
         follow: [
             { idField: `motion_category_ids`, fieldset: FULL_FIELDSET },
             { idField: `motion_state_ids`, fieldset: FULL_FIELDSET },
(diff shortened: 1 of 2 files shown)


client/src/app/site/pages/meetings/pages/projectors/projectors.subscription.ts
```diff
@@ -61,14 +66,16 @@ export const getProjectorListSubscriptionConfig: SubscriptionConfigGenerator = (
...
                 ]
             },
-            `projector_countdown_ids`,
-            `projector_message_ids`,
-            ...MEETING_DEFAULT_PROJECTOR_IDS_KEYS,
-            { idField: `speaker_ids`, additionalFields: [`meeting_user_id`] },
-            `list_of_speakers_ids`,
-            { idField: `agenda_item_ids`, fieldset: [`item_number`, `content_object_id`] }
-        ],
-        additionalFields: [`reference_projector_id`]
```
reiterl, line 71: This seems to be missing.

reiterl (commented):
At one place an `additionalFields` is gone. Please have a look at it.

client/src/app/site/pages/meetings/pages/projectors/projectors.subscription.ts
Elblinator: I moved it to line 28

reiterl (approved):
Looks okay to me.

bastianjoel (changes requested):
First of all you should use `FULL_FIELDSET`. 

Besides from that using the `FULL_FIELDSET` in most cases is not the correct way to go! For example in `projectors.subscription.ts` you are now subscribing the full meeting information. This is unnecessary. These warnings are there for a reason. Just suppressing them without thinking about if you really need all fields of a subscribed collection makes it unnecessarily harder for us in the future to fix these occurrences.

client/src/app/site/pages/meetings/pages/projectors/projectors.subscription.ts
bastianjoel: `additionalFields` was actually unnecessary. You don't need to subscribe additional fields when already subscribing ALL fields.

PR #4329, opened 2024-11-08
Use isUserPresent to include votingArea template

### LINKED ISSUE #4305: Voting: Change of presence problem - buttons didn't appear
**Describe the bug**
If a user first sets himself to present during an ongoing vote, the voting options are not immediately visible. The user must either open a different menu or reload the website. 

**How to Reproduce**
Steps to reproduce the behavior: An admin and another user (A) as delegate is needed
1. open a meeting
2. open motions > create a motion > create a poll with admin and delegate as groups who could vote
3. open in another browser the instance and log in as user a > open the autopilot. IMPORTANT: User should not be present and the user should have the option to change the presence state.
4. [as admin] start the poll
5. [as user A] you should see now the running poll, but oyu can't vote, because you are not present. -> change your presence state in the user profile menu > you are now able to vote, but the buttons didn't appear. you have to open the motion or relaod the autopilot to see them

**Expected behavior**
If a user changes his presence state during a running poll (or all other voting types) the voting options (buttons) should immediately appear. This should work in all menus, where a user could have the possibility to vote.

reiterl:
Resolve #4305 

I included the error msg here, if isUserPresent is false.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.scss
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.scss
@@ -250,3 +250,7 @@
 .text-right {
     text-align: right;
 }
+
+.text-center {
+    text-align: center;
+}
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.scss
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.scss
@@ -250,7 +250,3 @@
 .text-right {
     text-align: right;
 }
-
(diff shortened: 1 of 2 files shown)


client/src/app/site/pages/meetings/modules/poll/components/base-poll-vote/base-poll-vote.component.scss
```diff
@@ -250,3 +250,7 @@
 .text-right {
     text-align: right;
 }
+
+.text-center {
+    text-align: center;
+}
```
Elblinator, line 256:
```suggestion
```

text-center is defined in the utils.scss

Elblinator (changes requested):

Elblinator (approved):

PR #4334, opened 2024-11-08
Fix bug: second scrollbar appears in chat

### LINKED ISSUE #4302: Chat: Second scrollbar appears
**Describe the bug**
As soon as there are enough messages in the chat window to allow scrolling, an inner and an outer scrollbar appear. The outer scrollbar can be used to move the entire chat window. This is incorrect and is not required.

image

Using the outer Scrollbar:
image

**How to Reproduce**
1. Organization llevel > Settings > superadmin-settings > Enable chat globally
2. Create a meeting
3. open the meeting
4. open chat > create a new chat for the admin with read and write permission
5. add message until an inner and outer scrollbar appear

**Expected behavior**
Desktop mode:
An outer scrollbar should not appear when creating chat messages, as the chat is fully visible in desktop mode and therefore no scrolling is required. 

Mobile / Responsive mode:
Different settings are probably needed for this, which must take effect depending on the resolution or the device used. I tested it on a Pixel 8. Scrolling is required there because the input field is superimposed below other elements.

anbebe:
Closes #4302

### DIFF
--- a/client/src/app/site/pages/meetings/pages/chat/pages/chat-group-list/components/chat-group-list/chat-group-list.component.scss
+++ b/client/src/app/site/pages/meetings/pages/chat/pages/chat-group-list/components/chat-group-list/chat-group-list.component.scss
@@ -3,7 +3,6 @@
     max-width: 1000px;
     max-height: calc(100vh - 160px); // 160px is almost the summary of top bar and livestream header
     height: 95%;
-    margin-bottom: 0px !important;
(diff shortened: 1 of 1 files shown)


Elblinator (changes requested):
appearance is fine 

please remove the margin-bottom alltogether from the chat-group-list-card

Elblinator (approved):

PR #4342, opened 2024-11-12
Add option to projection modified final version

### LINKED ISSUE #4330: Motions: projection of editorial final version doesn't work
**Describe the bug**
The editorial final version of a motion can't be projected. Only the final version is projected. The outcome of the failure depends on two scenarios:

A) If a motion with amendment contains no conflicts and the user made some edits to create an editorial final version, the final version is projected. 
B) If a motion with amendments contains conflicts, which are manually solved, the projection of the (editorial) final version is empty.

In both cases, this means that the final editorial version of the motion cannot be projected.

**How to Reproduce**
1. open a meeting
2. open motions > create a motion with two paragraphs 
3. Change status of the motion to accepted and create a editorial final version > add a new word in the editorial final version, save it and project the motion in the 'final version'. The added word will not be displayed, but the other content is visible.
4. Reset the motion status to 'eingereicht'
5. Create two amendments on the same part of the motion to create a conflict. > Change the status of the amendments to accepted > change the motion status to accepted > create now a editorial final version to solve conflicts > solve them and save it > project the 'final version' > projector will be empty.

**Expected behavior**
In general: If a editorial final version of a motion with or without amendments or change requests is available, this version should be projected with the projection of the 'final version'.

bastianjoel:
resolves #4330

needs https://github.com/OpenSlides/openslides-autoupdate-service/pull/1075

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/view-models/view-motion.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/view-models/view-motion.ts
@@ -354,7 +354,7 @@ export class ViewMotion extends BaseProjectableViewModel<Motion> {
         ];
 
         if (this.modified_final_version) {
-            choices.push({ value: `modified_final_version`, displayName: `Editoral final version` });
+            choices.push({ value: `modified_final_version`, displayName: `Final version` });
         } else {
             choices.push({ value: `agreed`, displayName: `Final version` });
         }


Elblinator (approved):
code review

bspekker (changes requested):
If an editorial final version is available, this version should be projected with the projection of the 'final version'. It is not necessary to offer both options 'final version' and 'editorial version'.

PR #4343, opened 2024-11-12
Add line numbers to empty paragraphs

### LINKED ISSUE #4181: Motions > Editor: empty paragraph line leads to wrong behavior in amendments
**Describe the bug**
If a text in the editor contains an empty paragraph - set with Enter - this has an effect on amendments. Amendments that are placed on an 'empty' line are assigned line 0. In addition, the paragraph set in the editor is not visible in the motion. It is only visible in the editor or in the amendment creation screen.

Image of line 0 amendment:
image

**How to Reproduce**
1. open a meeting
2. open motions
3. create a motion which includes a paragraph set by enter
4. create an amendment to it and choose the empty line
5. --> Amendment should have now line 0 as base.

**Expected behavior**
If a paragraph is set in the editor using Enter, then this paragraph should also be visible in the motion and have the correct line numbering. It should also be possible to create a change request in this line.

**Additional context**
In OpenSlides 3 this was solved in the editor with a no-break space ( &nbsp ). 

Important: In the editor, however, it should still be possible to create a line break with SHIFT+ENTER that does not create a new line numbering. It is only about a line break with Enter.

bastianjoel:
resolves #4181

It should be tested if this impacts old motions which contain empty paragraphs. This is basically the same workaround as described in the issue just added to the line numbering instead of the editor.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/line-numbering.service/line-numbering.service.spec.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/line-numbering.service/line-numbering.service.spec.ts
@@ -552,13 +552,23 @@ describe(`LineNumberingService`, () => {
             }
         ));
 
+        it(`counts empty paragraphs and adds space`, inject([LineNumberingService], (service: LineNumberingService) => {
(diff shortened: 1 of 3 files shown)


Elblinator (approved):

Elblinator (approved):
Code review

bspekker (approved):

Issue #4281, opened 2024-10-21
Fix spinning wheel in login


Elblinator:
**Describe the bug**
If you login as one user navigate to a certain space logout and then login as another user, there is a chance that you cannot log in and you get a spinning wheel

**How to Reproduce**
1. Create two meetings (A and B) and two users (C and D)
2. Add user C to meeting A
3. Add user D to meeting B
4. open a new tap 
5. login as A 
6. log out 
7. login as B
8. spinning wheel
9. reload -> also spinning wheel
10. remove 'login' from the url and load -> you are directly logged in 

**Expected behavior**

**Additional context**

reiterl:
I have tested some more. It seems, that after the first logout that the router is on a wrong url. Not just '/login', but '/11/login'. This leads to some problems, if the next user is not allowed to see meeting 11.

reiterl:
@rrenkert I don't know how to continue here. My fix would possible break the pub access code.

Elblinator:
closed by #4542

Issue #4235, opened 2024-10-09
Fix redirect if a user is locked out


Elblinator:
**Describe the bug**
Sometimes a user is not correctly redirected if they are locked out of a meeting

**How to Reproduce**
1. Have a user (B) in several meetings (uncertain if that is a requirement but it is far more likely to happen with several meetings)
2. With one browser login as user B
3. Navigate to any page (except the autopilot)
4. Make a reload/Make a hard reload
6. With another browser login as an admin and lock out B 
7. The page "You are not supposed to be here." is shown instead of the wanted redirect
8. IF you do a normal reload in the home then the content in the mat-card and the menu left vanishes and you need to reload to be able to navigate somewhere else

**Expected behavior**
B should immediately be redirected

reiterl:
Your issue description mixes user A and user B.

Elblinator:
This Issue will be closed in favour of https://github.com/OpenSlides/openslides-client/issues/4404

PR #4344, opened 2024-11-12
Fix concurrent autoupdate single requests

### LINKED ISSUE #4234: Fix redirect from autopilot if a user is locked out
**Describe the bug**
If a User is locked out while they are on the autopilot they are not redirected to the dashboard/last meeting but every content but the projector vanish  from the autopilot and the menu on the left vanishes as well

**How to Reproduce**
1. Add user A to two meetings
2. Login as A and navigate to the autopilot of meeting One
3. Open new browser login as an admin
4. lock A out of meeting One
5. everything but the projector vanishes 

**Expected behavior**
A should be redirected to the other meeting

bastianjoel:
resolves #4234

### DIFF
--- a/client/src/app/site/services/auth-check.service.ts
+++ b/client/src/app/site/services/auth-check.service.ts
@@ -97,7 +97,7 @@ export class AuthCheckService {
         if (typeof info === `string`) {
             meetingIdString = this.osRouter.getMeetingId(info);
         }
-        if (Number.isNaN(Number(meetingIdString))) {
+        if (Number.isNaN(Number(meetingIdString)) || +meetingIdString <= 0) {
             return false;
         }
         await this.fetchMeetingIfNotExists(+meetingIdString);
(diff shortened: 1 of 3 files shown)


luisa-beerboom (approved):

reiterl:
Why is it needed to build a collection of resolved received here? I don't understand the reason for this code.

Elblinator (changes requested):
The reproduction from #4234 now consistently redirects you to the ```error?meetingId=2&error=Authorization%20Error``` url
-> The redirect to the error page is better than the behaviour before, but is it possible to redirect to the remaining meeting with our current setup?


The reproduction from #4281 is still resulting in the same behaviour (only step 9 now just shows a normal login page)

bastianjoel:
> The reproduction from #4234 now consistently redirects you to the error?meetingId=2&error=Authorization%20Error url
> -> The redirect to the error page is better than the behaviour before, but is it possible to redirect to the remaining meeting with our current setup?

If I understand correctly this is just #4235 which is explicitly not fixed here. 

> The reproduction from #4281 is still resulting in the same behaviour (only step 9 now just shows a normal login page)

In that case I would suggest removing this issue from this PR.

Elblinator (approved):

PR #4354, opened 2024-11-14
Add showReason method and handle p tags in reason

### LINKED ISSUE #4345: Motion > reason: Headline permanently displayed
**Describe the bug**
If a text has been entered in a reason and deleted again, the heading 'Reason' is still displayed in the motion and in the projection.

**How to Reproduce**
1. open a meeting
2. open motions > create a motion. enter some text in the reason field and delete it. save the motion
3. the motion detail view shows now the heading 'reason' without a text. 
4. project the motion
5. open Projector and check the projection. Heading 'reason' is shown.

**Expected behavior**
'Reason' should only be visible, if the reason contains content. Otherwise the reason should not be visible in the motion detail view nor the projection of the motion.

reiterl:
Resolve #4345

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-content/motion-content.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-content/motion-content.component.html
@@ -55,7 +55,7 @@
 
 <div class="motion-content">
     <!-- Reason -->
-    @if (motion.reason) {
+    @if (showReason) {
         <div>
             <h3>{{ 'Reason' | translate }}</h3>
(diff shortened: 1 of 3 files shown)


reiterl:
I looked at the `os-editor`. I didn't find the place, where the empty p is added in it. I used the simple solution to check showReason.

bastianjoel (changes requested):
While I think your change is needed for backward compatibility I think we should also modify the editor to not output empty paragraphs. (For example to fix the motion form currently accepting empty motions)

reiterl:
Remove the 'p' tags, if getText() delivers nothing. In cleanOutput.

Elblinator (approved):
function test

bastianjoel (approved):

PR #4379, opened 2024-11-20
Add VoteWeight documentation and translated search for detail poll view

### LINKED ISSUE #4289: Nominal Poll > documentation issues, information is missing
**Describe the bug**
The nominal polls have some documentation issues in the detail view of a poll, which needs to be fixed or enhanced. 

IMPORTANT: This only effects NOMINAL Votes, NOT non-nominal votes.

1. The entitled users list misses information about vote weight under the participant name. It is only visible in the single votes list
2. The search in the single votes list doesn't support german terms like Ja, Nein or Enthaltung. Therefor german users can't search for them. Only english words could be searched.
3. The pdf export misses information about the vote weight of a the participants which took part in the poll.

**How to Reproduce**
1. open a meeting
2. open settings > participants > activate vote weight
3. open participants > add yourself as superadmin to the meeting 
4. open participants > detail view of the administrator > change your vote weight to 5. If the vote weight field isn't available, you need to change this in the Account menu. 
5. open motions > create a motion > create a poll and choose the admin group as voting group > do vote and publish it > go to the detail view of the poll > in the entitled to vote list you will find the first issue , in the single votes list the second issue and via the three-dot-menu > export the third issue.

**Expected behavior**
1. The vote weight information should also be visible in the entitled users list, just like in the single votes list.
2. The search function should also find german terms Ja, Nein, Enthaltung. 
3. The pdf export should also include information about the vote weight of the participants. It should be placed in a new row between participant and votes. It should be named 'vote weight'. (see the image - marked the area with the red square.)

image

### LINKED ISSUE #4301: Motions > Nominal votes: Add vote weight to entitled users list
**Current behavior**
If a nominal vote is done, the user could open a detail view of the poll with two lists. The entitled users list misses an information about the vote weight, if it is activated in the settings. Also the pdf export of the data should include information about the vote weight, if it is activated in the settings. 

**Wanted behavior**
The poll detail view for nominal votes with activated vote weight should display the individual participant vote weight in the entitled users listing. Text should be 'Vote weight: Number'. If the vote weight is deactivated, the vote weight information should not be displayed. 

Also the PDF export should include information about the individual vote weight. There should be a new row between Participant and Has voted named 'Vote weight'.

IMPORTANT: This should not affect the Single Votes list. The vote must remain anonymous. 

**Additional context**
This issue is connected to #4289.

anbebe:
Closes #4289 
Closes #4301


Elblinator (changes requested):
The vote weight is displayed mostly as expected and the translation is working fine. 

There is an unclear case where the wanted behaviour is not completely clear: 
If a motion poll was nominal and then was anonymized:
**Status quo:** The type of the poll stays nominal but the single votes look like the single votes from a non-nominal poll.
**New behaviour with vote weight:** the vote weight is then still displayed in the entitled users list and the vote weight is exported in the pdf-export

Possibilities for the vote weight: 
A) The vote weight could be hidden and single votes and entitled users list would lookl like the lists from non-nominal votes
B) The vote weight could be displayed and be aligned with the type from the poll (if that is the case then there is another CR: 
If a nominal vote is anonymised and then the page is reloaded then there are Errors running through the console 
```
participant-controller.service.ts:113 ERROR TypeError: Cannot read properties of undefined (reading 'map')
```

@MSoeb Do you want A) or B) to happen or is there a third option which is preferred

client/src/app/site/pages/meetings/modules/poll/components/entitled-users-table/entitled-users-table.component.ts
```diff
@@ -42,6 +45,9 @@ export class EntitledUsersTableComponent {
         return this._isViewingThis;
     }
 
+    @Input()
+    public parent: BasePollDetailComponent<PollContentObject, PollService>;
+
```
bastianjoel, line 50: I would prefer a boolean property `displayVoteWeight`

client/src/app/site/pages/meetings/modules/poll/base/base-poll-detail.component.ts
```diff
@@ -24,6 +24,7 @@ import { BasePollPdfService } from './base-poll-pdf.service';
 
 export interface BaseVoteData extends Identifiable {
     user?: ViewUser;
+    vote_verbose_translated?: string | null;
```
bastianjoel, line 27: See next cr
```suggestion
```

client/src/app/site/pages/meetings/pages/motions/pages/motion-polls/components/motion-poll-detail/motion-poll-detail.component.ts
```diff
@@ -35,7 +35,13 @@ export class MotionPollDetailComponent extends BasePollDetailComponent<ViewMotio
     }
 
     protected createVotesData(): BaseVoteData[] {
```
bastianjoel, line 37: Instead of extending `BaseVoteData` (which is also used in other places) add a new type specific for this use case.

```suggestion
    protected createVotesData(): BaseVoteData[] {
```

bastianjoel (changes requested):

Elblinator:
After a discussion it was decisded that the vote weight should be displayed regardless of "no minal"/"non-nominal"

Elblinator (changes requested):
Please also add the vote weight (in exactly the same manner) to the election-polls

bastianjoel (approved):

Elblinator (approved):

PR #4383, opened 2024-11-22
Update saml attr setting formatting function

### LINKED ISSUE #4378: Saml attribute mapping cut off
**Describe the bug**
When displaying a saml attribute mapping that contains an empty string inside a nested object like the following, the mapping is cut off after the nested object with ":" and the string is filled with random characters fe. "}],".
```js
	"meeting_mappers": [
		{
			"external_id": "M2025",
			"mappings": {
				"structure_levels": [
					{
						"default": ""
					}
				],
```

**How to Reproduce**
1. Login as superadmin.
2. Go to orga settings.
3. Save a saml attribute mapping that follows the rules of backend PR and contains an empty string like above.
4. Reload the orga settings.

**Expected behavior**
The full attribut mapping should be displayed correctly.

**Additional context**
You need to have at least backend PR [2722](https://github.com/OpenSlides/openslides-backend/pull/2722) or a branch containing this PR to be able to save the mapping.

reiterl:
Resolve #4378 

Needs backend https://github.com/OpenSlides/openslides-backend/pull/2722


client/src/app/infrastructure/utils/functions.ts
bastianjoel, line 1: This function has tests. Add a test for the problem from the issue.

bspekker (approved):

rrenkert (approved):

PR #4392, opened 2024-11-25
Rm search-deleted-models presenter

### LINKED ISSUE #4386: Remove search-deleted-models-presenter code
The presenter doesn't exist in the backend any more, yet there is still code to call it in the client.
- A presenter service
- A method that calls it in the history-list

There is also a commented-out osSearchSelectorNotFoundTemplate in the history-list.component.html. That can be removed as well.

reiterl:
Resolve #4386 
It was already removed from the backend. So we need to cleanup here, too.

There are still doc stuff of this presenter in the backend.


luisa-beerboom:
> There are still doc stuff of this presenter in the backend.

That will be gone with OpenSlides/openslides-backend#2732

luisa-beerboom (approved):

Elblinator (approved):

PR #4421, opened 2024-11-28
Fix structure level card in autopilot (right side)

### LINKED ISSUE #4415: Fix structure level time widget
**Describe the bug**
The time wdiget was changed but now there are max two structure levels in one line (in desktop view)

**How to Reproduce**
1. Navigate to settings > LoS > activate  "Default speaking time contingent for parliamentary groups (structure levels) in seconds"
2. Add three structure levels
3. Have at least one participant for each structure level
4. Project content (e.g. a topic)
5. Navigate to the autopilot
6. Put the participants on the LoS (you want three or more structure level on the widget)

**Wanted behavior**
The mobile view should stay as it is
Desktop view there should be three structure levels in one line. (meaning that the text should wrap earlier and a single cell should be as small as it needs to be and as big as it can be

**Additional context**
https://github.com/OpenSlides/openslides-client/pull/4125

Possible solution is to limit the widget width and wrap the text if it's too long. In this example the widget width is 2 times bigger than the time width:
image

Elblinator:
closes #4415

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/speaking-times/speaking-times.component.html
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/speaking-times/speaking-times.component.html
@@ -24,7 +24,7 @@
             </div>
             <ng-container *osPerms="permission.listOfSpeakersCanManage">
                 <button
-                    class="icon-less-space-to-left"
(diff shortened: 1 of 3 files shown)


reiterl (commented):
In my test I don't see a difference.

Issue #4411, opened 2024-11-27
Fix amendment with exactly two words


Elblinator:
**Describe the bug**
B. If you create an amendment on an empty paragraph and this amendment has exactly two words then an additional line is added by this amendment

**How to Reproduce**
Example for the motion text
```
<p>A A A</p>
<p>
</p>
<p>C C C</p>
```

B) 
1. Create a motion with an empty paragraph (see above)
2. Add an amendment on the empty paragraph (here line 2) which adds exactly two words
3. The amendment adds not only your new two words but also an additional line in the presentation
Example for the amendment:
```
<p>BB BB</p>
```

**Expected behavior**
B) The amendment should only add the two words and not add a new line

reiterl:
To A) I see, that the 'p' tags has large margin-top and margin-bottom values. I could fix that. But for some reason there is a `br` tag in the empty paragraph with class `ProseMirror-trailingBreak` in it. This seems to cause the main problem. It is a problem with tiptap/ProseMirror.

reiterl:
To B) I think, I can reproduce the failure.

Elblinator:
To B) 
This specific scenario produces broken html (To find the trail search diffDetectBrokenDiffHtml() in code)
I would recommend  fixing this in it's own issue documented in our big Meta-Diff-Issue: https://github.com/OpenSlides/openslides-client/issues/2956

Elblinator:
Case A was fixed: 
I'll move Case A to the comments

Description:
A. If you have one empty paragraph in the editor then the height of this empty paragraph is too big 
Repo:
A) 
1. Open an editor
2. Add Text which has an empty paragraph
3. The empty paragraph has too much height
Example for the text
```
<p>A A A</p>
<p>
</p>
<p>C C C</p>
```

Expected behavior
A) An empty paragraph should have the same height as the otther paragraphs (which are only one line long)

PR #4425, opened 2024-11-28
Fix the appearance from the editor

Elblinator:
Fixes part of #4411


reiterl (approved):
Okay, that is an idea.

MSoeb (approved):
Solution looks fine.

PR #4428, opened 2024-11-29
Update motion submission date label visiblity

### LINKED ISSUE #4393: Motion > metadata > submission date: even without date the heading is visible for 'normal' users
**Describe the bug**
If no date has been entered under 'Submission date', the term 'Submission date' is still visible for all users which can see motions. This is unnecessary. 

**How to Reproduce**
1. open a meeting
2. open motions > create a motion. Delete the submission date after motion create.
3. open participants > create a user as delegate with the permission to see motions
4. login as delegate and check the motion detail view metadata. 

**Expected behavior**

**Additional context**
If no date has been entered under 'Submission date', the term 'Submission date' should be hidden for all users who do not have the permissions ' Can manage motions' and 'Can manage motion metadata'.

reiterl:
Resolve #4393


Elblinator (changes requested):
The motion needs a reload if the permissions from the participant change while the motion is already loaded
This seems to be a missing subscription, the Edit icon has the same problem, but that was already an Issue. If possible please also fix that part. If that's not possible please clarify, so an Issue for that can be created

Elblinator:
Repo:
- Create a motion
- Give user A the right to change meta_data from motions
- In a different browser as User A open the motion (you should see the mepty submission date -that's correct)
- As admin take the meta_data permission away
- As A the submission date stays visible and the edit icon also stays visible
- Reload the page and both vanish 

The submission date (if empty) and the icon should vanish without the need for a reload

reiterl:
Changing the permission don't have a instant effect on another browser. This would be a large feature.
Please open another issue for this.

reiterl:
Added a change to the permission check of this component.

client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-manage-timestamp/motion-manage-timestamp.component.ts
```diff
@@ -73,6 +84,18 @@ export class MotionManageTimestampComponent extends BaseUiComponent {
         );
     }
 
+    public ngOnInit(): void {
+        this.subscriptions.push(
+            this.operator.permissionsObservable.subscribe(() => {
+                this._change_metadata = this.operator.hasPerms(
+                    Permission.motionCanManageMetadata,
+                    Permission.motionCanManage
+                );
```
Elblinator, line 93:
```suggestion
                this._change_metadata = this.operator.hasPerms(Permission.motionCanManageMetadata);
```

It is enough to check for motionCanManageMetadata. 
motionCanManageMetadata is the important permission for this field. 
motionCanManage is allowed to be false here

Elblinator (changes requested):

client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-manage-timestamp/motion-manage-timestamp.component.ts
reiterl: done

reiterl:
@Elblinator I followed your suggestion and updated the perms in the timestamp component.

Elblinator (approved):

PR #4446, opened 2024-12-03
Add additional_submitter field to motion create, enabled by new setting

### LINKED ISSUE #4405: Motion creation: Add free text submitter field and committee selection field
**Current behavior**
During motion creation it is not possible to add a free text submitter. After creation it is possible to add a free text submitter.

**Wanted behavior**
As admin or as a participant with can manage metadata and can create motion permissions it should be possible to enter a free text submitter during motion create . To implement this, the existing components in motion metadata should be used. This should include the 'Extension' and 'Committee' fields. After integration persons with corrcet permissions mentioned above can enter free text submitters or choose a free text submitter from the committee field. The behavior should be the same like in the motion metadata.

The free text field and the committee field should be placed between submitter and title.
image

important: 
- A setting should be integrated in Settings > Motions, that allows to activate/deactivate the free text submitter field as well as the committee field in the motion create mask. The default value is deactivated. 
- The existing submitter creator field and the free text field should interact with each other. One of the two fields must be filled in. Example: If no text is entered in the free text field, the account creating the motion is set as the submitter. This is the existing behavior. If ONLY the free text field has a submitter and NO submitter is selected in the other field, then the account is not automatically set as the submitter. 
- The new fields 'Extension' and 'committee' should be in on line under 'submitter' field. The extension field should be wider than the committee field. 
- Committe field: Please take care, the box can have long terms. If this happens, a side scrollbar appears, which is is the current state only half visible. This should not happen. Bar should be fully visible.

**Additional context**
The added content (submitters) should behave like the existing component. The added submitters are free text, which will not create accounts.

luisa-beerboom:
Closes #4405
Needs OpenSlides/openslides-autoupdate-service#1090

### DIFF
--- a/client/src/app/domain/models/list-of-speakers/list-of-speakers.ts
+++ b/client/src/app/domain/models/list-of-speakers/list-of-speakers.ts
@@ -28,12 +28,12 @@ export class ListOfSpeakers extends BaseModel<ListOfSpeakers> {
         `id`,
         `closed`,
         `sequential_number`,
+        `moderator_notes`,
         `content_object_id`,
         `speaker_ids`,
         `structure_level_list_of_speakers_ids`,
         `projection_ids`,
-        `meeting_id`,
-        `moderator_notes`
+        `meeting_id`
     ];
 }
 export interface ListOfSpeakers extends HasMeetingId, HasProjectionIds, HasSequentialNumber {}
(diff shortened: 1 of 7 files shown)


luisa-beerboom:
backend is merged already

Elblinator (approved):
code review

bspekker (approved):
New setting and adding additional submitters works with permission 'can manage motions' works.

New issue will be created to allow participants with 'can create motions' and 'can manage metadata' to create motions with submitters other than themselves in the future.

PR #4454, opened 2024-12-05
Show Amendments which delete only whole paragraphs

### LINKED ISSUE #4451: Motion > Amendments: deletion of bullet point/paragraphs is not recognized as change
**Describe the bug**
If a bullet point or paragraph is completely deleted with a new amendment, this deletion is not recognized as an amendment. After saving, it is displayed that no change has been made. The crucial point is, that amendment which are created in this way show have no changed content and therefor the information of the change is not available anymore. 

Here an example for bullet points.

original motion content:
image

view of bullet point before editing:
image

view of bullet point after edit:
image

amendment view after save:
image

**How to Reproduce**
1. open a meeting
2. open motions > create a motion with one paragraph and three bullet points
3. create amendment to the second bullet point and delete the whole content > save
4. Amendment is now empty.

**Expected behavior**
If an amendment deletes a bullet point or a paragraph completely, the amendment should be displayed correctly. Specifically, this means that the bullet point or the paragraph is crossed out in red. 

This is part of the Meta diff Issue #2956

Elblinator:
closes #4451

### DIFF
--- a/client/src/app/domain/models/motions/motion.ts
+++ b/client/src/app/domain/models/motions/motion.ts
@@ -96,7 +96,7 @@ export class Motion extends BaseModel<Motion> implements MotionFormattingReprese
     }
 
     public amendment_paragraph_text(paragraphNumber: number): string | null {
-        return this.amendment_paragraphs && this.amendment_paragraphs[paragraphNumber]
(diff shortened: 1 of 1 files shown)


MSoeb (approved):
Looks good :)

rrenkert (approved):

PR #4455, opened 2024-12-05
Add meeting icons for user detail view

### LINKED ISSUE #3988: Closed meetings: Account detail view needs info about closed meetings
**Current behavior**
All meetings in which a user is participating are listed in the account detail view. There is currently no indication if the meeting is closed or not.

Example - Last meeting is closed without any indicator.
image

**Wanted behavior**
A lock icon should be added in front of the name of a meeting to make closed meetings recognizable. The size of the symbol should be based on the font size of the meeting name. The lock icon should be similar to the one used in the calendar.

**Additional behavior**
Please take care of public and archived meetings. Both types of meetings have individual icons, which should be used in the user details.

anbebe:
Closes #3988

### DIFF
--- a/client/src/app/site/pages/organization/pages/accounts/components/account-main/account-main.component.ts
+++ b/client/src/app/site/pages/organization/pages/accounts/components/account-main/account-main.component.ts
@@ -4,6 +4,7 @@ import { distinctUntilChanged, map, skip } from 'rxjs';
 import { OML } from 'src/app/domain/definitions/organization-permission';
 import { BaseModelRequestHandlerComponent } from 'src/app/site/base/base-model-request-handler.component';
 import { ViewUser } from 'src/app/site/pages/meetings/view-models/view-user';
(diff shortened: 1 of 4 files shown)


Elblinator (changes requested):
- Please move the tooltip to the left side of the icon

- I'm sorry for making a wrong wanted behaviour: 
Please give the icon the same colour as the groups (the colour changes in dark and light mode)

- If a user can see that a meeting is locked from inside but is not a part of this meeting then the link should have the same colour as the groups and should not be a link to the meeting

Elblinator (changes requested):
Wrong behaviour for the superadmin: 
If a superadmin is not a participant of a meeting, then the meeting is not a link (blue) anymore even though the superadmin can still access the meeting

Elblinator (approved):

Elblinator:
Code and Function

PR #4459, opened 2024-12-06
Fix public access settings and group visibility

### LINKED ISSUE #4423: Public Access: optimizations needed
**Current behavior**
The feature 'public access' has small problems if the feature is turned on in the meeting but globally turned off on organization level.

A) Settings > General > System: The checkbox is still clickable. This is inconsistent with other features that can no longer be used when deactivated. 

B) Participants > Groups: The 'Public' group is still available. This also makes no sense if the feature can no longer be used. Additionally the group is only deactivated - no more visible - by switching off the function at meeting level.

**Wanted behavior**
A) When deactivating the feature at organization level, the setting (checkbox and text) in the meeting should be greyed out. However, the checkbox should remain selected if it was selected before deactivation, i.e. if public access was activated. 

B) The 'Public' group should be hidden and deactivated in the group rights matrix when it is deactivated globally. However, the settings should be retained so that they do not have to be set again when the function is reactivated.

anbebe:
Closes #4423

### DIFF
--- a/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
+++ b/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
@@ -226,7 +226,7 @@ export class MeetingSettingsGroupDetailFieldComponent extends BaseComponent impl
                 }
             ]
         });
-        if (this.disabled || this.getWarning()) {
+        if (this.disabled || this.hasWarning()) {
             this.form.disable();
         }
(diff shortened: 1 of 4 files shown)


client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
```diff
@@ -226,7 +226,7 @@ export class MeetingSettingsGroupDetailFieldComponent extends BaseComponent impl
                 }
             ]
         });
-        if (this.disabled) {
+        if (this.disabled || this.getWarning()) {
```
Elblinator, line 229: please change the method title to sth more appropriate for a boolean return value (e.g. hasWarning)

Elblinator (changes requested):
functionally it works!

just wording

Elblinator (approved):

Issue #4452, opened 2024-12-04
Motion > amendment: formatting change in amendment plus original text leads to wrong diff


MSoeb:
**Describe the bug**
Part I: If an amendment change also includes a formatting change in amendment and original text, the diff will be wrong displayed. The whole paragraph will be shown as diff not the relevant parts.

Part II: If a motion paragraph includes for example a bold format in two sentences and an amendment is created in the second sentence with an additional word and a change in the bold content, than the diff is wrong displayed. The whole paragraph will be shown as diff. 

**Part I - Original motion:**
![image](https://github.com/user-attachments/assets/791685fd-657f-4cc8-8355-630c13ed969f)

**Part I - Amendment with red marked diff (amendment plus original text):**
![image](https://github.com/user-attachments/assets/ed543daf-1ff7-4b3e-878f-564628a93b1a)

**Part I - Outcome view:**
![image](https://github.com/user-attachments/assets/9d18a7bf-bbfe-4769-b3b9-1d205656814d)

**Part II - Original motion paragraph**
![image](https://github.com/user-attachments/assets/98d2492c-a3bf-47b5-a857-41ce35c98fc7)

**Part II - changes done in the amendment - changes are red marked:**
![image](https://github.com/user-attachments/assets/ea9deec8-8034-4631-bb98-711096407b8d)

**Part II - Outcome:**
![image](https://github.com/user-attachments/assets/2381e437-8e1f-417d-8060-7f3ac02cb2c7)

**How to Reproduce**
1. open a meeting
2. open motions > create a motion with two paragraphs. Add in the second paragraph a bold formatting of one word in the first and in the second sentence. > save it.
3. Part I - create an amendment in first paragraph and change one word. mark the new word and an original word in bold. > save diff
4. Part I - visual should now be broken like in the example above
5. Part II - create an amendment in second paragraph and change the second sentence as follow: add one new word right before the bold word and add content to the bold word > save diff

**Expected behavior**
If an amendment contains text and fomat changes, the diff visual should clearly display differences and should not show a change in the whole paragraph.

Important: Bold format is just an example. This problem also occurs for italic, underlined, strikethrough and color changes. This should not happen for all of these format changes. - FYI: I've tested all of this format types with Part I. At least for this, same problems occur.

This issue is part of the Meta Diff issue. #2956

MSoeb:
- [ ] changed title
- [ ] extended problems with the counterpart of formatting
- [ ] added a new examples
[...]

PR #4461, opened 2024-12-07
Fix some diff issues when changes include formatting (#4452)

CatoTH:
This fixes *some* scenarios, in which errors happen when textual changes and formatting changes interfere. A complete fix of all possible scenarios would be infinitely more complex, up to not really possible.


bastianjoel (approved):

emanuelschuetze (approved):
tested

PR #4472, opened 2024-12-09
Fix show all changes checkbox state persistance

### LINKED ISSUE #4426: Motion detail view: Amendments Display error
**Describe the bug**
Motions with amendments have display errors in the 'Show all changes' function. If amendments are not displayed in the motion, you can use this function to show the amendments. This works correctly. The error now occurs if you activate the amendments via 'Show all changes', then click on the motions menu and open the motion again, the amendments are visible even though the 'Show all changes' checkbox is not checked. The browser seems to remember the activated setting.  

This also occurs if another menu instead of motions is opened. 

Example - The amendment is visible, but the checkbox is not marked:
image

**How to Reproduce**
1. open a meeting
2. open motions and create a motion with an amendment
3. open the motion detail view and check 'show all changes'. The amendment should now be visible.
4. open motions menu
5. open motion -> Amendment is visibile and 'Show all changes' box is not marked

**Expected behavior**
When switching from a motion with activated 'Show All Changes' function to another menu and back to the motion, the browser should not have remembered the previously set checkmark. When reopening the motion, the amendments should not be visible if the check mark is not set.

Important: It is possible for amendments to be permanently displayed in the motion. This should continue to behave as usual and is excluded from this bug fix.

bastianjoel:
resolves #4426

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/base/base-motion-detail-child.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/base/base-motion-detail-child.component.ts
@@ -113,5 +113,6 @@ export abstract class BaseMotionDetailChildComponent extends BaseMeetingComponen
 
     private destroy(): void {
         this.cleanSubscriptions();
+        this.viewService.showAllAmendmentsStateSubject.next(false);
     }
 }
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
code and function

PR #4475, opened 2024-12-10
Participant create wizard and presence

### LINKED ISSUE #4419: Meeting > User create: issue with present status during user creation
**Describe the bug**
If you add a user to a meeting or create a new one, the presence status is always set to absent. It does not matter whether the presence checkbox was checked in step 3 or not. 

When debugging with Loki, we noticed that the attendance is not sent when referencing. See picture.
image

Something is sent during the new creation, but this has no influence on the presence status. See image.
image

**How to Reproduce**
1. open a meeting
2. open participants
3. create a new user. uncheck and check presence state during creation. > check presence state after creation.
4. Add a new user by referencing an existing user. presence checkbox should stay marked -> after creation the user is not present 

**Expected behavior**
1. when creating or referencing users in a meeting, the presence checkbox should not be checked by default. 
2. if the presence checkbox is checked, the user should be created as present.

reiterl:
Resolve #4419

### DIFF
--- a/client/src/app/gateways/repositories/users/user-repository.service.ts
+++ b/client/src/app/gateways/repositories/users/user-repository.service.ts
@@ -148,7 +148,7 @@ export class UserRepositoryService extends BaseRepository<ViewUser, User> {
         const data = usersToCreate.map(user => {
             const meetingUsers = user.meeting_users as Partial<ViewMeetingUser>[];
             return {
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
function

bastianjoel (approved):

PR #4480, opened 2024-12-11
Use office paste plugin from npm registry

bastianjoel:


bastianjoel:
Problem with bookmarks will be tracked in https://github.com/Intevation/tiptap-extension-office-paste/issues/1

As feedback for the bookmark issue is provided very slowly this PR will just do the switch to the package. Therefore merging now.

PR #4485, opened 2024-12-16
Add notice about running votes for public access users

### LINKED ISSUE #4401: Public Access: running poll needs information
**Current behavior**
Users who have logged in with public access only see the voting title in the autopilot or in the motion (or in an election or in the agenda item) when a vote is in progress. It is not clear that the vote is currently in progress. An additional note would be useful here.

Example in Autopilot:
image

Example in motion:
image

**Wanted behavior**
During ongoing votes, a notice should be displayed for public access users in the voting tile in the autopilot and in the voting tile in a motion, an election or an agenda item.

Notice: 'Voting in progress'

Please take care, that this should be translateable.

anbebe:
Closes #4401


Elblinator (approved):
Code review

MSoeb (changes requested):
In general it does what it should, but if a poll is started, an error appears in the console.

image

1. Login as public user 
2. open Autopilot and console
3. open another browser as superadmin and start a motion poll. (important: public user should have permission to see motions and autopilot)
4. take a look at the public user browser. in the console the error should appear.

MSoeb (approved):
Works fine.

PR #4489, opened 2024-12-16
Fix wrong language after login

### LINKED ISSUE #4462: Wrong language after login
**Describe the bug**
After first login in an instance with German language, the first opened meeting has an English menu structure, even if the browser and the rest of the instance is in German. 

I'm not sure, if it only happens after first login.

**How to Reproduce**
1. login into an instance with cleared website cache
2. open a meeting in -> wrong language.

**Expected behavior**
The correct language should be displayed in the meeting. In the case above it should be German.

bastianjoel:
resolves #4462 

This seems to improve the situation. But I am not entirely sure if it fully fixes it.


Elblinator (approved):
Code looks fine 
Cannot test functionality

MSoeb (changes requested):
Talked about it with Bastian. Fix doesn't work as intended. Language still wrong at first login.

luisa-beerboom (approved):
Code LGTM

MSoeb (changes requested):
Doesn't work. Problem still occurs. Tested with Chromium and Brave.

Additionally something strange appears in both browsers after opening the website: The CPU load rises and fan speed rises.

MSoeb (approved):
Looks good. Tested with Chrome, Brave and FF.

PR #4490, opened 2024-12-16
Use backend directory structure

hjanott:
Since the backend directory structure will change as per https://github.com/OpenSlides/openslides-backend/issues/2604 we need a small change within the github workflows.
This means that the `build-and-test-prod-image` of the build.yml will use the new backend location of the data folder.


peb-adr (approved):
CI should pass, once it sees the new path in backend

PR #4491, opened 2024-12-16
Improve view model list equality checks

### LINKED ISSUE #4303: Motion detail view with amendments: Changing amendment status needs reload
**Describe the bug**
It is possible to change the status of an amendment in the motion detail view, but the change is not displayed. The status switches up, but the old status is still displayed. An update that updates the visulization seems to be missing here.

**How to Reproduce**
1. open a meeting
2. open motions > create a motion with an amendment
3. open the motion in the detail view
4. mark the checkbox 'Show all changes' > try to change the amendment status via the three-dot-menu behind the amendment

**Expected behavior**
Changing the amendmend status via the three-dot-menu should have visual effect. New status should be immediately visible

bastianjoel:
resolves #4303 

Please check if #4303 is still fixed with this. This now also fixes potential errors with change recommendations not being updated.


Elblinator (changes requested):
starting a poll is creating too many autoupdate updates (several in a second)

The state change from the change recommendations and the amendments work fine with this

Elblinator (approved):
besides the function.spec.ts test I see nothing

PR #4492, opened 2024-12-18
Add max per option validator and onkeyon for votes input fields

### LINKED ISSUE #4453: Elections: two issues in cumulative voting
**Describe the bug**
In elections with Yes votes, there is a logical error with cumulative votes. It is possible to create an election that allows more votes per option than the maximum number that can be cast. This is incorrect. The maximum vote value is defined independently of cumulation. The accumulation is ranked behind it. It only specifies the maximum number of votes that can be cast per option. This value cannot be higher than the maximum number of votes.

The backend already seems to detect this correctly, as the value of the maximum number of votes that can be cast is taken into account when the votes are cast.

Additionally the button to send votes doesn't become active if numbers are added. It stays gray, but it can be clicked.

Example of ballot setting - affected settings are marked in red:
image

Example what happened during voting - number is entered, but button is still gray
image

Example - number entered is to high and the dialog show correctly the maximum possible votes:
image

**How to Reproduce**
1. open a meeting
2. open settings > elections > ballots > mark 'Allow to accumulate several votes on one candidate ("comulative voting")'
3. open elections > create an election > add a candidate > create a ballot with voting method 'yes per candidate' > add in 'Maximum amount of votes' 4 > add in 'Maximum amount of votes per option' 6 > select in 'entitled to vote' admin group > save
4. start the ballot and enter a 6 in the number of votes field

**Expected behavior**
The maximum number of votes per option should be based on the maximum number of votes. 'Maximum amount of votes per option' may be either equal to or less than the value in 'Maximum amount of votes'. If a larger value is entered, an error should appear and it should not be possible to create the vote.

In addition, the voting button should become active as soon as a value has been entered.

reiterl:
Resolve #4453

### DIFF
--- a/client/src/app/infrastructure/utils/validators/is-number-range.ts
+++ b/client/src/app/infrastructure/utils/validators/is-number-range.ts
@@ -1,11 +1,11 @@
 import { AbstractControl, ValidatorFn } from '@angular/forms';
 
-export function isNumberRange(minCtrlName: string, maxCtrlName: string): ValidatorFn {
+export function isNumberRange(minCtrlName: string, maxCtrlName: string, errorName: string): ValidatorFn {
     return (formControl: AbstractControl): { [key: string]: any } | null => {
         const min = formControl.get(minCtrlName)!.value;
         const max = formControl.get(maxCtrlName)!.value;
         if (+min > +max || Number.isNaN(+min) || Number.isNaN(+max)) {
-            return { rangeError: true };
+            return { [errorName]: true };
         }
         return null;
     };
--- a/client/src/app/infrastructure/utils/validators/is-number-range.ts
+++ b/client/src/app/infrastructure/utils/validators/is-number-range.ts
@@ -1,6 +1,6 @@
 import { AbstractControl, ValidatorFn } from '@angular/forms';
 
-export function isNumberRange(minCtrlName: string, maxCtrlName: string, errorName: string): ValidatorFn {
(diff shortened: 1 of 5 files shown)


client/src/app/infrastructure/utils/validators/is-number-range.ts
```diff
@@ -1,11 +1,11 @@
 import { AbstractControl, ValidatorFn } from '@angular/forms';
 
-export function isNumberRange(minCtrlName: string, maxCtrlName: string): ValidatorFn {
+export function isNumberRange(minCtrlName: string, maxCtrlName: string, errorName: string): ValidatorFn {
```
bastianjoel, line 3: Might be a good idea to set a default value for that. 

```suggestion
export function isNumberRange(minCtrlName: string, maxCtrlName: string, errorName: string = `rangeError`): ValidatorFn {
```

bastianjoel (changes requested):

bastianjoel (approved):

Elblinator (changes requested):
The min-votes are now in warn-colour if the two other amounts are wrong, they should not be in warn colour
Screenshot_20250106_154204


If Yes as voting method was selected and then max per option is bigger max per vote and then another voting method is selected then you cannot save the balloot
Screenshot_20250106_154107

reiterl:
As min-votes, max-votes and max-votes-per option is one form group it is not easy possible to have the validator on just two of the fields.

Elblinator (approved):
function

both described things are fixed

bastianjoel (approved):

PR #4499, opened 2024-12-27
Update it.po 

Elblinator:
The it.po was updated in the main-repo, this just updates the client as well (https://github.com/OpenSlides/OpenSlides/pull/6859)


bastianjoel:
Why do you need a review for this?

PR #4508, opened 2025-01-06
Allow empty html element at the begining of an amendment

rrenkert:
Relevant if the amendment  starts with a header and you add bfore the header.


Elblinator (approved):
Please wait with merging

PR #4516, opened 2025-01-08
Add allow_to_edit_own_delegation

### LINKED ISSUE #3802: Add allow_to_edit_own_delegation
See https://github.com/OpenSlides/OpenSlides/issues/6223 for context

**Desired behaviour:**

The group permission needs to be added in groups > participants: Under `Can see participants`. The title for the new permission should be `Can edit own delegation`. To edit your own delegation you need to permission to see participants (canSee should automatically be activated)

If a User can edit their own delegation then they should be able to open the edit dialog for themselves and be allowed to set and remove someone in „Voting right delegated to (proxy)“ and they should only be allowed to remove the user in „Voting right received from“
There should be an additional click path to open this dialog for yourself: participant-list > top right three dot menu > option to „Change your delegation“

anbebe:
Closes #3802 

<s>Needs Backend OpenSlides/openslides-backend#2632</s>(merged)


Elblinator (changes requested):
Hide the three dot menu if there is nothing in the menu (e.g. if the person can edit their own delegation but not the delegation from other people)

Elblinator (approved):
functionality

bastianjoel:
@Elblinator There was still an open change request when you assigned me.

bastianjoel (approved):

PR #4519, opened 2025-01-08
Fix alignment of microphone in speakers list

### LINKED ISSUE #4407: List of speakers: listing layout optimization
**Current behavior**
If an active speech is visible in a list of speakers, users who do not have list administration permissions will see an unnecessary layout indentation. The microphone icon and the name are no longer left-aligned. The arrangement probably depends on the view and the options for admins. (Keyword play button). This is not necessary for users without administration permissions.

Here an example:

List of speakers - layout without active speaker:
image

List of speakers - layout with active speaker:
image

**Wanted behavior**
The alignment of the microphone icon including text should, for users without list of speakers management permission, be based on the following speech; in the example the text '1st'. 

For users which can manage the ist of speakers nothing should change.

anbebe:
Closes #4407


emanuelschuetze (dismissed):
Tested. Mic icon is not alligned correcly, see red line in screenshot:

image

The left side of the mic icon should placed at the red line.

Elblinator (approved):
code

MSoeb (changes requested):
I found an error, which occurs with motions that have visibility restrictions. If I project a motion with the workflow 'complex' and add someone to the list of speakers an error occurs in the console. I've checked the behavior on our test instance and it didn't occur. So I assume, that the PR is somehow related to the problem. Could you please check this? - Except of this, the position of the icon looks good. 

Error message:
image

Reproduction:
1. open a meeting
2. open motions. create a motion with complex workflow. don't change state -> project it
3. open console
4. open Autopilot. Add someone to list or push the Play button. Interactions with the list of speakers leads to errors.

anbebe:
I could not reproduce this error. I tried it on the branch and also updated the branch. In both cases I could interact with the list of speakers without receiving an error.
From the error message it seems more to me that there is a problem with the projection of the motion.

MSoeb (approved):
Looks good. Can't replicate error anymore. Maybe my browser was the problem. 

FYI: The mentioned CR from @emanuelschuetze is fixed. Therefor I will dismiss his CR.

PR #4529, opened 2025-01-10
Fix creation of amendments that have no changes

### LINKED ISSUE #4520: Amendment creation mask: Creation button first active after change in editor
**Current behavior**
It is possible to create a new amendment without making any changes. This is possible because the “Create” button can already be clicked without making any changes. This results in unnecessary amendments, which could be avoided.

image

Button in red square is crucial part.

**Wanted behavior**
When creating amendments, the “Create” button should only become active when a change is entered in the editor.

anbebe:
Closes #4520


client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-form/components/amendment-create-wizard/amendment-create-wizard.component.html
```diff
@@ -19,7 +19,7 @@ <h2>{{ 'New amendment' | translate }}</h2>
                     <button
                         mat-button
                         [disabled]="contentForm.value.selectedParagraphs.length === 0"
-                        (click)="matStepper.next()"
+                        (click)="matStepper.next(); checkCanSave()"
```
bastianjoel, line 22: Add a new method. Normally we do not want more than one method call within these event listeners.

bastianjoel (changes requested):

bastianjoel (approved):

PR #4547, opened 2025-01-15
Fix visual warning for deletion of a meeting

### LINKED ISSUE #4514: Protection against accidental deletion of an event
**Description:**
When working on a committee's meeting, the inhibition threshold for accidentally deleting a meeting is no higher than duplicating it.

**Expected behaviour:**
If the "OK" in the deletion-dialog is pressed an additional dialog should open itself and ask if the meeting should really be deleted,, pressing "OK" here should delete the meeting, pressing cancel should not delete the meeting

anbebe:
Closes #4514

### DIFF
--- a/client/src/app/ui/modules/prompt-dialog/components/prompt-dialog/prompt-dialog.component.html
+++ b/client/src/app/ui/modules/prompt-dialog/components/prompt-dialog/prompt-dialog.component.html
@@ -3,7 +3,7 @@
         <h2 mat-dialog-title>{{ data.title | translate }}</h2>
         <mat-dialog-content [innerHTML]="data.content | translate"></mat-dialog-content>
         <mat-dialog-actions>
-            <button color="warn" mat-button [mat-dialog-close]="true" [ngClass]="{ delete: canDelete }">
+            <button color="warn" mat-button [mat-dialog-close]="true" [ngClass]="{ delete: data.deletion }">
                 {{ (data.confirm ? data.confirm : 'OK') | translate }}
             </button>
             <button mat-button [mat-dialog-close]="false">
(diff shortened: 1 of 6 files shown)


reiterl (commented):
I am not sure, if that works well with the translation.

client/src/app/ui/modules/prompt-dialog/components/prompt-dialog/prompt-dialog.component.ts
```diff
@@ -18,4 +18,8 @@ export class PromptDialogComponent {
         public dialogRef: MatDialogRef<PromptDialogComponent>,
         @Inject(MAT_DIALOG_DATA) public data: PromptDialogData
     ) {}
+
+    public get canDelete(): boolean {
+        return this.data.confirm ? this.data.confirm.includes(`delete`) : false;
```
reiterl, line 23: Mmh, what will happen, if 'delete' is translated somewhere above?

anbebe: I think you are right and that it cannot be solved by checking the string but needs something different. I will take a look at it.

client/src/app/ui/modules/prompt-dialog/components/prompt-dialog/prompt-dialog.component.html
```diff
@@ -3,7 +3,7 @@
         <h2 mat-dialog-title>{{ data.title | translate }}</h2>
         <mat-dialog-content [innerHTML]="data.content | translate"></mat-dialog-content>
         <mat-dialog-actions>
-            <button color="warn" mat-button [mat-dialog-close]="true" [ngClass]="{ delete: canDelete }">
+            <button color="warn" mat-button [mat-dialog-close]="true" [ngClass]="{ delete: data.deletion }">
```
reiterl, line 6: Does data always contain the deletion attribute? If no, try `?.` here.

reiterl (commented):
Mmh. one remark. And the interface of the prompt service is growing.

client/src/app/ui/modules/prompt-dialog/components/prompt-dialog/prompt-dialog.component.html
anbebe: No, deletion is not always contained in data but a ? is not necessary in this case since data always exists and the ngClass can handle undefined if deletion is not contained. That's how I understood it so far and it works.

reiterl (approved):
Looks Okay to Me

Elblinator (approved):

PR #4550, opened 2025-01-15
Fix error on first login after password change

### LINKED ISSUE #4143: Password Change didn't work well
**Describe the bug**
If a user wants to change the password different errors occur. 

A) The first login after password change is not possible without a page reload. An error appears 'Fehler bei der Kommunikation mit dem Autoupdate-Service'. 

B) Additionally: Sometimes by changing the password an error ccurs that prevent the user from using the page anymore. A reload is needed. (Not sure about reproduction.)

**How to Reproduce**
1. open user profile
2. change your password
3. try to log in with the new password (Case A)

**Expected behavior**
Changing the password via user profile should change the password, but didn't result in an error message (B). The user should be logout after clicking on 'Save'-Button. Also there should be no error message on the login page after password change. (A)

Additional behavior:
A new gray help text should added above the 'Save' Button. 
Text shoul be: _Note: You will be logged out when you change your password. You must then log in with the new password._
image

bastianjoel:
resolves #4143


Elblinator (approved):
functionality and code

MSoeb (changes requested):
Looks fine, but after password change an error occurs multiple times in the console, which needs to be fixed.

image

Reproduction:
1. open account menu
2. open console
3. open password change menu. change password and save it -> errors occur.

MSoeb (approved):
Looks good

PR #4565, opened 2025-01-22
Add live vote register to motion/assignment-poll-detail

### LINKED ISSUE #3986: Live voting register during long-running polls
**Wanted behavior**
In the case of long-running votes on motions and ballots, a list of voters is provided that shows the status of all entitled voters live. The existing list of entitled users in the detailed view of a vote or ballot need to be expanded to include a live list view so that it is possible to see which entitled voter has voted while the vote is in progress. There is no insight into the individual votes. The live electoral roll is only visible to users who have the right to manage elections or motion polls.

**Additional context**
Further details or questions need to be discussed, if needed. Please contact @emanuelschuetze or me for detailed context and UI concept.

reiterl:
Resolve #3986 

First draft version.
Needs:
https://github.com/OpenSlides/openslides-autoupdate-service/pull/1118
https://github.com/OpenSlides/openslides-vote-service/pull/308

Reuse of the entitled-users-table.


MSoeb:
The first draft already looks good, but some changes are necessary:

1. layout - Live Voting Register
- Please use the existing column design with 'Single Votes' and 'Entitled users'. A separate view, as currently available, is not required. 
- Based on the previous point: The 'Single votes' column is NOT filled during an ongoing vote. The 'Entitled users' list only shows who has already voted.
- During active voting: The text 'No results yet.' should be replaced by the text 'Voting in progress.
- During active voting: A progress bar should be displayed below the text 'Voting in progress'. 
- During active voting: The 'Single votes' column should display a note that no data is available yet.  

2. data visualization:
- The 'Entitled users' list MUST include all persons called to vote. This includes all participants who are in the group(s) called to vote. This also implies participants who are not present.
- Vote delegations: Similar to the previous documentation and recording, transfers of voting rights should also be documented for nominal votes and be visible in the 'Entitled users' list already during the vote. The existing behavior can be reused here.

reiterl:
I chatted with @MSoeb and looked into the code.
I think the autoupdate service should provide a `entitled-users-live` field with the same data as entitled users at stop but live.

reiterl:
The current implementation uses the new `poll.has_voted_user_ids` field. This field provides a list of ids of user who already voted.
The live vote register should show all users, who are allowed to vote. (e.g. including users who haven't voted yet and delegates). The users who are allowed to vote depend on the `poll.entitled_groups` field and the delegates.
So there are some calculations needed and I am not sure how to handle the delegate to part.

reiterl:
I have worked on the different parts:

1.  all users who are allowed to vote are display.
2. delegation handling is also in place.
3. the live vote register into the is in the mat tab
4. in a running vote a new text is displayed
5. add a poll progress component

Please have a look on this.

client/src/app/site/pages/meetings/modules/poll/base/base-poll-detail.component.ts
```diff
@@ -276,6 +285,48 @@ export abstract class BasePollDetailComponent<V extends PollContentObject, S ext
...
+                    map(users => users.filter(user => userIds.has(user.id)))
+                )
+                .subscribe(users => {
+                    const entries: EntitledUsersTableEntry[] = [];
+                    for (const user of users || []) {
+                        const delegateToId = user.vote_delegated_to_id();
+                        entries.push({
+                            id: user.id,
+                            user: user,
+                            voted_verbose: `voted:true`,
```
luisa-beerboom, line 316: Is this supposed to always be `voted:true`?
If no, fix it. If yes, you may as well remove this property.

luisa-beerboom (changes requested):

client/src/app/site/pages/meetings/modules/poll/base/base-poll-detail.component.ts
reiterl: Fixed.

MSoeb (changes requested):
Some changes are required. 

1. Please use the existing column (tab) design with 'Single Votes' and 'Entitled users'. A separate view, as currently available, is not required. (relevant part in image: lower red square.)
2. Additionally more information is needed under the progress bar. Please add information about entitled users and present entitled users with numbers (see red marked area under the progress bar). Please take care, the numbers are not static and should change like the corresponding numbers does. The text should be right-alinged to the progress bar. The color should be gray, just like the words 'orange, schwarz' in the image. The information should be written in german in the following way:
- Stimmberechtige anwesend: [right number of progress bar]
- Stimmberechtigte gesamt: [number of entitled users]

image

MSoeb:
> To point 2. This is **not** part of this issue. I won't solve it. @MSoeb please open a new issue for drilling on the poll-progress.

Yes, this was not previously a requirement. However, this has changed. It is now a requirement. Please integrate it.

Explanation:
Without this new information, there is no explanation as to why the progress bar and the entitled voters list show a different number. In order to be able to understand what these different numbers mean without an explanation (e.g. in the FAQ), this information is absolutely essential. This was not previously foreseeable, but was added as a new, essential requirement for the feature during the last review.

Further information:
The information that is now to be added to the progress bar should always be displayed in the progress bar. For example, even when it is visible in the autopilot.

reiterl:
I think, the logic with the progress bar should be handled in another issue. I will exclude the poll-progress here, because it includes misleading information.

luisa-beerboom (approved):

rrenkert:
@reiterl The additional text requested by @MSoeb is not about progress. It repeats the numbers of the progress bar and the list view to have an explanation what the numbers are about. These number should only be in sync with another.
Can you please integrate the additional text?

rrenkert (changes requested):
The progress was a required element in this view. please readd it.

reiterl:
I talked with @rrenkert about this missing two values. I will add them in the two views. @MSoeb no need for a new issue then.

luisa-beerboom (approved):

MSoeb (changes requested):
Found different points, that needs to be fixed or improved:

1. By opening the entitled users tab (red square in image) during a live poll, an error appears in the console. -> Wanted behavior: Please fix error.

image

image

2. The numbers in the progress bar and in the information beneath it doesn't match.  (see examples in red squares). They don't match, because the lower number doesn't include vote delegations. -> Wanted behavior: Vote delegations should be included in the number 'present people allowed to vote. Both numbers should always match.

image

3. Re-Positioning of the new information (marked relevant content in image). Please place the information more next to the progress bar (marked with green) . It should have more space to the content below. 

image

4. Re-Positioning of 'No data avilable'. The current position is not optimal. PLease please it, if possible, in the middle aligned to the word above. I've marked a possible area with green.

image

reiterl:
One other version of the the live vote register.
to point 1:  fixed the error
to point 2: Used the same way to select to get the present number as in the poll progress component.
to point 3: changed
to point 4: add a text align center and a color gray here ( as the global no-content class)

Please test.

MSoeb (changes requested):
Found two things, that needs to changed: 

1. Entitled users (during voting)
During a live voting the entitled users list (red square in image) isn't visible completely anymore. Only parts of it are visible. --> Whole list should be visible.
image

2. Single votes list: position of 'no data available'
image
Please add above and under 'No data available' space. Currently it doesn't look good. The element should be a bit more separated from the other elements.

reiterl:
To 2. added margin top/bottom to the 'no available data'.
To 1. In my tests the entitled user table component sometimes doesn't show all entries. It is not all the time and it seems to
 be a problem with this component. The observer had delivered the right data, as far as I can say. And it is not a problem with
an error in the entitled users html stuff. Most time I see the problem if I scroll. I will continue to debug here.

reiterl:
To point 1. In Scrolling table component there seems to be something wrong.
The itemSize and the item pixel buffer seems to be part of this problem. Changing the itemSize to 40 seems to fix the display problem. As adding more buffer, calculated from rowHeight.
Anyway in my test it works now. Please review and test.

client/src/app/site/pages/meetings/pages/assignments/pages/assignment-polls/components/assignment-poll-detail/assignment-poll-detail.component.html
```diff
@@ -63,6 +76,19 @@ <h1>{{ poll.title }}</h1>
...
+                    } @else if (poll.isEVoting && poll.isStarted) {
+                        <mat-tab-group>
+                            <mat-tab label="{{ 'Single votes' | translate }}">
+                                <div class="no-content no-data-margins">{{ 'No data available' | translate }}</div>
+                            </mat-tab>
+                            <mat-tab *osPerms="permission.pollCanManage" label="{{ 'Entitled users' | translate }}">
+                                <os-entitled-users-table
+                                    [displayVoteWeight]="displayVoteWeight"
+                                    [entitledUsersObservable]="liveRegisterObservable"
+                                    [isViewingThis]="true"
```
luisa-beerboom, line 88: I think `isViewingThis` is true by default and should never otherwise be set

luisa-beerboom (approved):

MSoeb (approved):
looks good.

rrenkert (approved):

PR #4570, opened 2025-01-24
Add jasmine tests for the theme service

reiterl:
Using spyObj of jasmine to mock most injections.


client/src/app/site/services/theme.service.spec.ts
```diff
@@ -1,16 +1,68 @@
...
+        expect(storageSpy.set.calls.count()).toBe(1);
+        expect(dark).toBe(true);
+        service.toggleDarkMode();
+        expect(storageSpy.set.calls.count()).toBe(2);
+        expect(dark).toBe(false);
+    });
+
+    it(`get currentAccentColor`, () => {
+        expect(service.currentAccentColor).toBe(`#2196f3`);
+    });
```
bastianjoel, line 61: We don't want tests to fail when changing a color.

bastianjoel (changes requested):

client/src/app/site/services/theme.service.spec.ts
reiterl: Okay, I am now using the constants of the service for the test.

bastianjoel (approved):

PR #4571, opened 2025-01-27
Move meta folder to higher level

### LINKED ISSUE #4482: Move the meta folder to a higher level
**Currently:**
The meta repository is currently on level 3 of the openslides-client repo ("/openslides-client/client/src/meta")

**Wanted:**
Best case: Move the meta folder to "/openslides-client/meta"

If the meta folder needs to be enbedded in the "client" folder then a solution would be to just move it to "/openslides-client/client/meta

@bastianjoel is there a technical constraint for the client Repo that needs to meta to be in client?

Elblinator:
closes #4482


peb-adr (changes requested):
Woah, this does not look right.
The submodule (referencing [openslides-meta](https://github.com/OpenSlides/openslides-meta)) should not go away.
Instead of copying all the contents only the "mount point" should move. Then all path references mst change.

bastianjoel:
CLI tools need to be updated

peb-adr:
`grep -r meta` revealed these files which need to be changed.

    client/cli/generate-permissions.ts:const SOURCE = path.resolve(path.join(__dirname, '../src/meta/permission.yml'));
    client/cli/generate-settings-defaults.ts:const SOURCE = path.resolve(path.join(__dirname, '../src/meta/models.yml'));
    client/cli/generate-models.ts:const SOURCE = path.resolve(path.join(__dirname, '../src/meta/models.yml'));

These look like the CLI tools bastian mentioned.
From what I can tell this looks like everything that needs to be changed. If @bastianjoel agrees, I approve

peb-adr (approved):

peb-adr:
LGTM, IMO a client dev should give a final OK though.

bastianjoel (approved):

Issue #4399, opened 2024-11-25
Better Accessibility for blind users using a projector


Jonathan-Matthew:
**Current behaviour**
At our last conference we had a blind user.
He used OpenSlides in Safari with iPhone 12 (IOS 18.1).
He used the integrated VoiceOver-Screenreader and the Fokus Blue 5. Gen Braille display.
He complained that his screenreader got the information from the projector with a large delay. So he had to refresh the side regulary.
Another problem he faced was that visual effects changed the position of the cursor of his screenreader on the website

**Additional context**
Same user and setup as in #4397

bastianjoel:
@rrenkert Do we want to implement this before switching to the projector service? 

I guess we could use aria live regions for this https://www.w3.org/TR/wai-aria/#dfn-live-region
Angular CDK also has support for live announcements which we could use https://material.angular.io/cdk/a11y/overview#liveannouncer

PR #4588, opened 2025-01-27
Replace projector with projector service

bastianjoel:
TODO:
- [x] Remove projector content subscriptions
- [ ] Check accessibility concerns regarding `div` overlaying projectors `iframe` content
  - [ ] Maybe investigate #4399 while at it
- [ ] Fix visual glitching on load
- [ ] Projector edit dialog preview


bastianjoel:
Reimplementing this in a less conflict prone way as this will stay open for some time.

PR #4601, opened 2025-02-03
Add warning icon as svg to pdf export

### LINKED ISSUE #4546: Motion - PDF Export: layout issues with amendments in conflict
**Describe the bug**
If a motion has amendments which are in conflict, the PDF export doesn't work well. The corresponding conflicted text parts starts with 'warning+motion number'. This should be improved.

Example: 
[Motion with conflicts.pdf](https://github.com/user-attachments/files/18423350/Motion.with.conflicts.pdf)

**How to Reproduce**
1. open a meeting
2. open motions
3. create a motion > create two amendment on the same paragraph and create a conflict. > Change status of both amendments to accepted.
4. Open the main motion > export it as pdf > take a look at the conflicts

**Expected behavior**
Instead of showing the word warning, the conflict icon should be printed. This is generic for all languages. Also at least one space should be between the icon and the motion number.

Not sure if it is necessary, but I think, that the text format can be improved to show it in the style like in the motion detail view. (Need to be discussed)

anbebe:
Closes #4546

### DIFF
--- a/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
+++ b/client/src/app/gateways/export/html-to-pdf.service/html-to-pdf.service.ts
@@ -213,6 +213,10 @@ export class HtmlToPdfService {
                 newParagraph = this.createListParagraph(createPayload);
                 break;
             }
+            case `svg`: {
+                newParagraph = this.createSVGParagraph(createPayload);
+                break;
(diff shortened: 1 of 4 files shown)


Elblinator (approved):
fuonction

bastianjoel (approved):

PR #4602, opened 2025-02-03
Fix line numbering layout in motion export with default font

### LINKED ISSUE #4568: PDF Export: Layout problem with custom fonts and line numbering with four digits
**Describe the bug**
In the PDF export, display problems occur if a motion has a line numbering with at least 4 digits and a custum font has been selected. The line numbering and the motion text are then directly next to each other. (See the image example) The distance between line numbering and text varies depending on the selected font.

Example of wrong behavior:
Image

Used fonts:
[Exmaple with Fonts - Inter.zip](https://github.com/user-attachments/files/18517789/Exmaple.with.Fonts.-.Inter.zip)

**How to Reproduce**
1. open a meeting
2. open motions > create a motion > change line start numbering to 4 digits
3. export the pdf -> looks normal
4. open files > add custom fonts for regular, bold und italic as provided by the link > select the fonts for the correct font types. 
5. open new motion > export it as pdf

**Expected behavior**
The aim is to ensure that four-digit line numbering and a custom font do not lead to display problems. There should also be a space between line numbering and motion text. 

This could possibly be achieved by ensuring that the font for line numbering always remains in the standard Fira Sans format and that this is independent of custom fonts.

anbebe:
Closes #4568

### DIFF
--- a/client/src/app/gateways/export/pdf-document.service/pdf-worker.worker.ts
+++ b/client/src/app/gateways/export/pdf-document.service/pdf-worker.worker.ts
@@ -47,7 +47,8 @@ function applyLayout(content: any): void {
  */
 function initPdfMake(data: any): void {
     pdfMake.fonts = {
-        PdfFont: data.fonts
+        PdfFont: data.fonts,
+        LineNumbering: { normal: `fira-sans-latin-400.woff` }
     };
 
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
functionality

bastianjoel (approved):

PR #4610, opened 2025-02-05
Add moderation note pdf export

### LINKED ISSUE #4590: Moderation note: PDF Export
**Current behavior**
Currently the moderation note misses an export function.

**Wanted behavior**
A pdf export option for moderation should be included. The pdf export option button should be placed in each moderation note. This would allow a first easy export.

Here is an initial suggestion where the button could be placed: It should be placed right to the edit button.
Image

Important: The export should NOT be possible in the autopilot moderation note mat card.

reiterl:
Resolve #4590 

Add a service for the PDF export. Use the title of the los as header. A moderation note is always attached to a los. Use MeetingPdfExport, like in topic export.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/moderation-note/moderation-note.component.html
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/moderation-note/moderation-note.component.html
@@ -33,6 +33,15 @@
                                 <mat-icon>edit</mat-icon>
                             </button>
                         }
+                        @if (showModerationNotesExport) {
+                            <button
+                                mat-icon-button
+                                matTooltip="{{ 'Export moderator note as PDF' | translate }}"
+                                (click)="onDownloadPDF()"
+                            >
+                                <mat-icon>picture_as_pdf</mat-icon>
+                            </button>
+                        }
                     </div>
                 }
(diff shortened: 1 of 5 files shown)


bastianjoel (approved):

MSoeb (changes requested):
One thing is missing:

If a user has the permission (via group permission) to see moderation notes, the user should have also the possibility to export moderation notes. Currently this is limited to the permission 'can manage moderation notes'. This should be changed. 

This change should be added to establish export integrity to other exportable content. 

Please take care, that this should not allow export in the Autopilot.

reiterl:
Tested with 'main':
A user with only 'canSeeModerationNotes' can't see the moderation notes.
A user with 'canSeeModerationNotes' and 'canSeeListOfSpeaker' can see the moderation notes. 
As the moderation notes are a subpart of the los the permission 'canSeeLOS' is **required**. 

The new error @MSoeb  reported is orthogonal to the export enhancement and should not be handled here.

MSoeb (approved):
Looks good. Users with permission 'can see moderation notes' and 'can see LoS' can see and download moderation notes. Except for the Autopilot, which is intended and correct behavior. 

PS: My last comment could be ignored. I missed a configuration during testing.

PR #4614, opened 2025-02-06
Fix LineNumbering font in virtual file system

### LINKED ISSUE #4611: Fix motion-export
**Describe the bug**
In a productive system if you try to use the motion export then the followign Error is thrown which blocks the export:
```
Uncaught (in promise) File 'fira-sans-latin-400.woff' not found in virtual file system
```

**How to Reproduce**
1. Start a productive system
2. Make a motion
3. Export the motion (single or multi)
4. The Error is thrown

**Expected behavior**
The export should work

anbebe:
Closes #4611


luisa-beerboom (approved):

Elblinator (changes requested):
The followign comes out when I try exporting (single and multi) 
```Uncaught (in promise) RangeError: Index out of range```

Elblinator (approved):
It works now

Elblinator (changes requested):
If a fone is selected for "Normal" then the export fails

PR #4639, opened 2025-02-13
Fix insertions in empty paragraph of amendment with two words

### LINKED ISSUE #4411: Fix amendment with exactly two words

anbebe:
Closes #4411


bastianjoel (changes requested):
Please add a test

Elblinator (approved):

bastianjoel (approved):

PR #4652, opened 2025-02-17
Add set_as_template to the meeting create form

### LINKED ISSUE #4632: Meeting creation view: Add new checkbox for template
**Current behavior**
The creation screen for meetings currently does not include the option to mark a meeting directly as a template. In addition, a correct error message appears in the mask if you want to create a meeting without an administrator. The error refers to the fact that this is only possible if you define a meeting as a template. This is correct, but the option to create meetings as templates is missing at this point. Overall, it seems inconsistent why it is not possible to specify at this point whether a meeting should be a template or not. 

**Wanted behavior**
In the meeting creation screen, a checkbox is to be added at the last position, which can be used to select whether a meeting is a template or not. This would also directly offer the option of creating the meeting without an admin. 

**Additional context**
The language of the error message should be translated in the corresponded meeting language.

reiterl:
Resolve #4632 

Also add the error message 'Cannot create non-template meeting without admin_ids' to the ErrorMap, so it can be translated.

### DIFF
--- a/client/src/app/gateways/error-mapping/error-map-utils.ts
+++ b/client/src/app/gateways/error-mapping/error-map-utils.ts
@@ -45,7 +45,8 @@ const MeetingCreateErrorMap: ErrorMap = new ErrorMap([
     [
         /Only one of start_time and end_time is not allowed./,
         _(`Start and end time must either both be set or both be empty`)
-    ]
+    ],
+    [/Cannot create non-template meeting without admin_ids/, _(`Cannot create non-template meeting without admin_ids`)]
 ]);
 
 /**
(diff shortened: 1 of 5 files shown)


client/src/app/gateways/error-mapping/error-map-utils.ts
```diff
@@ -45,7 +45,8 @@ const MeetingCreateErrorMap: ErrorMap = new ErrorMap([
     [
         /Only one of start_time and end_time is not allowed./,
         _(`Start and end time must either both be set or both be empty`)
-    ]
+    ],
+    [/Cannot create non-template meeting without admin_ids/, _(`Cannot create non-template meeting without admin_ids`)]
```
bastianjoel, line 49: Maybe this should be changed.

bastianjoel (approved):

bspekker (approved):

PR #4654, opened 2025-02-18
Add not amendment filter to motion-list

### LINKED ISSUE #4638: Motions: Add new filter 'Is not an amendment'
**Current behavior**
If a user wants to create a pdf export of all final versions without amendments, it is necessary to set two filters to get a clean motion list. The user has to select in the amendment filter 'Has amendments' and 'Is no amendment and has no amendments'. This is not intuitive.

**Wanted behavior**
A new filter should be added, that shows ALL motions without amendments. This would help a lot. The filter should be named 'Is not an amendment'. It should be placed at the second position in the red square. 

Image

reiterl:
Resolve #4638


bastianjoel (approved):

MSoeb (changes requested):
Small change needed:

Please change position of the new filter to second position under "Ist ein Änderungsantrag".

image

reiterl:
I changed the order of the amendment filters as requested.

MSoeb (approved):
Looks good

PR #4665, opened 2025-02-21
Improve editor controls tab navigation

### LINKED ISSUE #4150: Improve editor controls tab navigation

Elblinator:
closes #4150

### DIFF
--- a/client/src/app/ui/modules/editor/components/editor/editor.component.html
+++ b/client/src/app/ui/modules/editor/components/editor/editor.component.html
@@ -1,18 +1,32 @@
 <div class="editor">
     @if (editorReady) {
-        <div class="editor-menu">
+        <div
+            aria-label="'Press Enter to reach the editor functions' | translate"
+            class="editor-menu"
+            tabindex="0"
+            (keydown.Enter)="toggleInside()"
+        >
             <div class="button-group">
                 <button
                     class="text"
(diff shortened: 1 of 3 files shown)


bastianjoel:
This does not really implement what is described under "Wanted behavior". 

When writing the issue I did some research on what other editors do and the described way was the behavior I had mostly seen (for example TinyMCE or the Github editor implement this). As this should not only be an accessibility feature but also to make the overall experience better when filling these forms without switching between keyboard and mouse I really think we should stick to these standards. 

My problem with #4206 was that you need to add a repeating scaffolding to every button but besides from that it was exactly what was asked for. Thats why I suggested to add a directive that can be applied to every button. Your PR needs less scaffolding but there is still a lot of repetition. 

I think the whole menu should be cleaned up to add such a feature. Either the buttons should be implemented within a generic component or the menu structure should be defined in a object and then dynamically generated in the template. Maybe it even makes sense to do both.

PR #4686, opened 2025-02-26
Fix line breaks in lists of pdf export

### LINKED ISSUE #4647: Fix line breaks in lists in PDF export of motions
**Describe the bug**
Two line breaks in a list item are shown as a small gap in the motion detail view (similar to a gap between two paragraphs), but as an empty line in the PDF export. This causes the line numbers to be off. 

**How to Reproduce**
1. create a motion with a list
2. in one list item, create two line breaks
3. export PDF of the motion

Motion text:
Image

Motion detail view:
Image

PDF export:
Image

**Expected behaviour**
The two additional line breaks should be visualised in the PDF as a gap and not an empty line as it is in the motion detail view.

anbebe:
Closes #4647


bspekker (approved):

Elblinator:
I tested it with custom fonts which are not broken 

This was the issue with the original issue 

Can you please either state what the problem was or merge this PR? @m-schieder

m-schieder:
I think there was an error when change recommendations in nested lists go over one line.
To test this, #4688 must first be solved.

rrenkert:
waiting for #4688 to be fixed.

PR #4687, opened 2025-02-26
Fix wrong sub settings updates

### LINKED ISSUE #3779: Fix wrong settings updates

anbebe:
Closes #3779

### DIFF
--- a/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
+++ b/client/src/app/site/pages/meetings/pages/meeting-settings/pages/meeting-settings-group-detail/components/meeting-settings-group-detail-field/meeting-settings-group-detail-field.component.ts
@@ -261,7 +261,9 @@ export class MeetingSettingsGroupDetailFieldComponent extends BaseComponent impl
                     this._comparedForm ||
                     (JSON.stringify(form.value) !== JSON.stringify(this._firstValue) && this.setting.type !== `groups`)
                 ) {
-                    this.onChange(form.value);
+                    this.setting.type === `ranking`
+                        ? this.onChange(this.form.controls[`value`].value)
+                        : this.onChange(form.value);
                 }
             });
     }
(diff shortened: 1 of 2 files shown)


bastianjoel (commented):
Not entirely sure but I think there are cases where the backend complains about certain options being saved when another option is disabled. 
@luisa-beerboom could you have a look on this?

Elblinator (changes requested):
I tested it and the settings seem to be in a flat hierarchy in the backend

If you create point of orders then the created PoO are not saved if the main setting is not  off or if the subsetting is off.

These are the payrequests if everything works: 
They are send in this order and I only created one PoO
```
[
    {
        "action": "point_of_order_category.delete",
        "data": [
            {
                "id": 4
            }
        ]
    },
    {
        "action": "point_of_order_category.create",
        "data": [
            {
                "meeting_id": 3,
                "text": "wq",
                "rank": 1
            }
        ]
    }
]
```

```
[
    {
        "action": "meeting.update",
        "data": [
            {
                "list_of_speakers_enable_point_of_order_speakers": true,
                "list_of_speakers_enable_point_of_order_categories": true,
                "id": 3
            }
        ]
    }
]
```


This is the payload if they are not working:
(you can also which ture/false that would be another axample where it isn't working, and turning both false is doing the same) 
```
[
    {
        "action": "meeting.update",
        "data": [
            {
                "list_of_speakers_enable_point_of_order_speakers": true,
                "list_of_speakers_enable_point_of_order_categories": false,
                "id": 3
            }
        ]
    }
]
```

bastianjoel (approved):

Elblinator (approved):

PR #4709, opened 2025-03-06
Add multiselect drag/drop for sorting tree

### LINKED ISSUE #1536: Feature: add multiselect option for Drag & Drop lines (like in list of speakers)
Description: 
Currently, in the list of speakers, it is possible to move several people at once by holding CTRL and clicking on selections. This option should be available wherever there are lists to sort.

Important: Currently, the function still has the bug that when you move only the first selection is displayed and not all the selected options. This should be taken into account when implementing the function. 

Reproduction:
1. open a meeting
2. create an agenda item
3. project the item to standardprojector
4. open participants and create some user
5. go to autopilot and add them to the list of speakers
6. select several speakers by holding CTRL plus left mouse click
7. Drag & Drop them in the list.

anbebe:
Closes #1536

### DIFF
--- a/client/src/app/ui/modules/sorting/modules/sorting-tree/components/sorting-tree/sorting-tree.component.ts
+++ b/client/src/app/ui/modules/sorting/modules/sorting-tree/components/sorting-tree/sorting-tree.component.ts
@@ -122,6 +122,12 @@ export class SortingTreeComponent<T extends Identifiable & Displayable> implemen
     private _model: Observable<T[]> | null = null;
 
     /**
+     * The index of multiple selected elements. Allows for multiple items to be
+     * selected and then moved
+     */
+    public multiSelectedIndex: number[] = [];
+
+    /**
      * Input that defines the key for the parent id
      */
     @Input()
@@ -548,7 +554,10 @@ export class SortingTreeComponent<T extends Identifiable & Displayable> implemen
      *
      * @param event CdkDragStart which emits the event
      */
-    public startsDrag(event: CdkDragStart): void {
+    public startsDrag(event: CdkDragStart, node: FlatNode<T>): void {
+        if (this.multiSelectedIndex.length && !this.multiSelectedIndex.includes(node.id)) {
+            this.multiSelectedIndex = [];
+        }
         this.removeSubscription();
         const draggedNode = <FlatNode<T>>event.source.data;
         this.placeholderLevel = draggedNode.level;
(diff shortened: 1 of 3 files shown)


client/src/app/ui/modules/sorting/modules/sorting-tree/components/sorting-tree/sorting-tree.component.ts
```diff
@@ -559,16 +568,73 @@ export class SortingTreeComponent<T extends Identifiable & Displayable> implemen
...
 
+    /**
+     * Handles a click on a row. If the control key is clicked, the element is
+     * added/removed from a multiselect list /(which will be handled on
+     * dropping)
+     *
+     * @param event MouseEvent.
+     * @param indx The index of the row clicked.
+     */
+    public onItemClick(event: MouseEvent | Event, indx: number): void {
```
bastianjoel, line 579:
```suggestion
    public onItemClick(event: MouseEvent | Event, clickIdx: number): void {
```
Might be better to use more meaningful variable names here.

bastianjoel (approved):

client/src/app/ui/modules/sorting/modules/sorting-tree/components/sorting-tree/sorting-tree.component.ts
```diff
@@ -559,16 +568,73 @@ export class SortingTreeComponent<T extends Identifiable & Displayable> implemen
...
+        if (event.type === `click` || (<KeyboardEvent>event).key === ` `) {
+            event.preventDefault();
+            if ((<MouseEvent>event).ctrlKey) {
```
bastianjoel, line 582: While you are at it - this might be a nice addition to prevent text selection when using multiselect. 
```suggestion
            if ((<MouseEvent>event).ctrlKey) {
                document.getSelection().removeAllRanges();
```

Elblinator (changes requested):
In the call-list, category and agenda:
have four items like this: 
A
B
C
 -> C1 (this is indented)
Select C and C1
Grab C and carry them up

What happens is that only C is gettung dragged up and C1 is not indented anymore

If the are ion the top and you mo ve them down then id it was (C, C1, A, B ---> A, C1, B, C) and C1 stays indented

AND
If you the items are at the top then the sort does not work

Elblinator (changes requested):
1)
A
-> B
C

Cannot be dropped to the end 
The item do not need to be at the beginning,
 but the drop to the end is needed

A
-> B 
--> C
-> D
E
F

If you try to change B, C and D to E then the following is the outcome: 
A
E
-> B
-> C
F
-> D

D should be below C and above F

Elblinator (approved):

bastianjoel (approved):

PR #4748, opened 2025-03-13
Allow YNA to set min/max vote amount

### LINKED ISSUE #4738: Add the min and max option to YNA per candidates
**Current behavior**
Image

**Wanted behavior**
The option min and max option from yes per candidate should be added:
But the default for the max should be the amount of candidates 
The default for min should stay 1

This is how it looks for "Yes per candidate"
Image

reiterl:
Resolve #4738 
First version. Perhaps the form control usage is suboptimal.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.ts
@@ -238,10 +238,11 @@ export abstract class BasePollFormComponent extends BaseComponent implements OnI
     }
 
     private updatePollMethod(method: PollMethod): void {
+        this.contentForm.removeControl(`votes_amount`);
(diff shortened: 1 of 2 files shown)


Elblinator (commented):
Everything works 
But I was too  inaccurate in my Issue: 
The min/max are currently in the election poll and the motion poll 
As I understood the assignment the min/max should only be in the elction poll @bspekker are both wanted or just the elction poll?

bspekker:
The min/max option is only needed for elections with the voting method "YNA per candidate" which doesn't exist for motion polls.

Elblinator (changes requested):
Then please remove the min/max for the motion poll

reiterl:
Only assignment-polls (elections) are allowed to set min/max votes in YNA pollmethod.

bastianjoel (approved):

Elblinator (approved):

PR #4757, opened 2025-03-17
Allow Committee admin to change passwords

### LINKED ISSUE #4643: Password change on account level doesn't work properly for admins with limited scope
**Describe the bug**
If a user is both a committee and meeting admin and has all rights to change personal data of an account, the admin can only change the password inside a meeting. It is not possible in the account menu in the account details.  

However, it is correctly possible in the participant menu within a meeting. This is inconsistent.

error message:
Image

**How to Reproduce**
1. open accounts > create a new user A
2. open committees > create a new committee and add user A as committee admin > create a new meeting and set user A as meeting admin
4. open meeting
5. open participants and add a user to the meeting B
6. login as user A 
7. open Accounts > open Account detail of user B > try to change password of user B > error message above appears

**Expected behavior**
In the case above it should be possible to change the password of user B in the Account menu. Admins which have unlimited access to a user should have to possibility to change the password of an account via Account menu.

Elblinator:
closes #4643


reiterl (approved):
Code okay, this allows committee managers to change the password too.

bspekker (approved):

PR #4773, opened 2025-03-18
Fix committee admin's ability to add user on account level

### LINKED ISSUE #4772: Allow committee admin to add a user on account level
**Describe the bug**
If a committeee admin tries to add a user on the account level then the "Add" button stays disabled

**How to Reproduce**
1. Createa committee and 2 meetings
2. Create a commitee admin (from the just created committee) and a normal user (A)
3. Add the user to one of the meetings
4. login as committee admin
5. Navigate acount-list > A > three dot > "Add to meeting"
6. Try to add A to the other meeting 
7. ==> the ADD button stays disabled

**Expected behavior**
That you can Add A to the other meeting 

In the same way the higher admins are allowed to add A

anbebe:
Closes #4772


Elblinator (changes requested):
The problem is still there 
The problem is in the ADD button and not the "Add to meeting" button

Elblinator (approved):

PR #4783, opened 2025-03-20
Fix edit profile with initial presenter call

### LINKED ISSUE #2493: fix edit "my profile"
Not all information can be changed if the own profile is changed from the menu in the upper right corner and dependent on the layout changes pedending on the available information.

**Current behaviour:**
Information edit:
1. login
2. open the dialog in the upper right corner > my account
3. edit "my profile"
4. Title, Given Name, Surname, intial password and natural person are disabled
5. close and open the edit menu
6. now nothing is disabled and everything can be edited

**Wanted behaviour:** 
If you have to permission to edit your profile, the field should not be disabled

reiterl:
Resolve #2493 

Calls the presenter to set isUserInScope and enable the fields.


Elblinator (approved):

bastianjoel (changes requested):
Not sure about this as it introduces unnecessary presenter calls. 
Wouldn't it be possible to just wait with toggeling `_isEditing` until `updateIsUserInScope` is resolved?

reiterl:
There were changes to the account-dialog (See #4797 ). The problematic fields are not in the dialog anymore.

PR #4786, opened 2025-03-21
Fix change recommendations with nested lists

### LINKED ISSUE #4688: Motion > change recommendations: problems with nested lists
**Describe the bug**
If change recommendations done to nested lists, some errors occur.

Original motion - start point:
Image

Motion with change recommendation: The CR adds ONLY content to the line without a line break.
Image

Problem A - Add more content in a line -> The line will snap to the next line and u
Problem B - Unnecessary dots appears (see the red dots in image)

**How to Reproduce**
1. open a meeting
2. open motions
3. create a motion in the style above
4. add a change recommendation in layer 3, add content in the line and save

**Expected behavior**
Creating a change recommendation in a nested list should need create a line break and it also should not create the unnecessary red dots. 

**Additional information**
The problem in this case seems to be the editor. @m-schieder and @bastianjoel talked about it. 

Important: The wrong behavior leads also to a wrong pdf export. he errors are visible. This should also be tested after the problems are fixed.

anbebe:
Closes #4688

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-diff.service/motion-diff.service.spec.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-diff.service/motion-diff.service.spec.ts
@@ -927,6 +927,36 @@ describe(`MotionDiffService`, () => {
             expect(diff).toBe(expected);
         }));
 
+        it(`handles inserted listitems within nested lists`, inject(
+            [MotionDiffService],
+            (service: MotionDiffService) => {
+                // Hint: Start of list item on second level of list (e.g. in change recommendations)
+                const before =
+                        `<ul><li>` +
+                        `<ul><li><span class="line-number-2 os-line-number" contenteditable="false" data-line-number="2">&nbsp;</span>Zweite Ebene` +
+                        `<ul><li><span class="line-number-3 os-line-number" contenteditable="false" data-line-number="3">&nbsp;</span>Dritte Ebene` +
+                        `<ul><li><span class="line-number-4 os-line-number" contenteditable="false" data-line-number="4">&nbsp;</span>Vierte Ebene</li>` +
+                        `</ul></li></ul></li></ul></li></ul>`,
+                    after =
+                        `<ul><li>` +
+                        `<ul><li>Zweite Ebene` +
+                        `</li><li>Neue Zwischenebene` +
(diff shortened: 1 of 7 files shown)


Elblinator (changes requested):
If you edit the the change recommendation then the class "hidden-li" is not loaded
 
Repo: 
1. Create motion 
2. Create change recommendation in line 4
3. Reload the page
4. Edit the change recommendation
5. The bulletpoints from the motion are visible
6. Reload
7. Display is correct

Motion text: 
```
<ul>
	<li>Lorem ipsum dolor sit amet, consectetuer adipiscing elit.<ul>
	<li>Aenean commodo ligula eget dolor.<ul>
	<li>Aenean massa.<ul>
	<li>Cum sociis natoque penatibus et magnis dis parturient montes</li>
</ul>
</li>
</ul>
</li>
</ul>
</li>
</ul>
```

Change Rec create: 
```
<ul>
	<li class="hidden-li">
		<ul>
			<li class="hidden-li">
				<ul>
					<li class="hidden-li">
						<ul>
							<li>Cum sociis natoque penatibus et magnis dis parturient montes hi</li>
						</ul>
					</li>
				</ul>
			</li>
		</ul>
	</li>
</ul>
```
Change rec edit:
```
<ul>
	<li class="hidden-li">
		<ul>
			<li class="hidden-li">
				<ul>
					<li class="hidden-li">
						<ul>
							<li>Cum sociis natoque penatibus et magnis dis parturient montes edit</li>
						</ul>
					</li>
				</ul>
			</li>
		</ul>
	</li>
</ul>
```

anbebe:
This behaviour is unreliably reproduceable. 
After looking at it with @Elblinator the assumption is that somewhere/somehow processes sometimes take too long but there was no obvious fix for me so far. The additional points disappear when clicking e.g. at the three-dot menu or otherwise forcing a reload of the view. 
@bastianjoel Do you have an idea? @bspekker Is this bug maybe better in a new issue because of its sparse appearance and the main bug fix for bullet lists can be already merged (after approvement)?

bastianjoel (changes requested):
I would prefer a solution that is limited to the change recommendation creation or the editor without changing the diff service. Solving this will be hacky either way but we should prevent adding such workarounds to application parts like the `MotionDiffService` that are already very complex. 

Maybe it is worth investigating how the previous editor handled such cases. I will have another look on this the next week and might provide some requirements on how to solve this.

PR #4789, opened 2025-03-24
Fix different issues around vote

### LINKED ISSUE #4785: Fix vote
**Describe the bug**
Several things are wrong/missing with the vote

**How to Reproduce**
1. Navigate to an election
2. Put at least two candidates on the election list
3. Create an election with Y/N per candidate 
4. Start the election
5. sth visually should change (more in Expected behaviour)
6. Try to give more than one vote -> you are prevented to give more than one vote 
7. Create a different election with Y/N/A per candidate
8. If you send your vote then the dialog does not close after the first time 
9. In general the dialog shows the wrong wording

**Expected behavior**
Image

- [x]  If Available votes is >1 is should be presented
- [x] The Y/N per candidate should also have the "Minimum amount of votes" and "Maximum amount of votes"
- [x] The text for the "Submit selection now?" should again show how many of how many votes you are sending 
- [x] The dialog should not open two times

**Additional context**

Elblinator:
closes #4785

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.ts
@@ -80,10 +80,10 @@ export abstract class BasePollFormComponent extends BaseComponent implements OnI
     }
 
     /**
-     * The flag to allow min/max votes on YNA poll method
+     * The flag to allow min/max votes on YNA and YN poll method
      */
     @Input()
-    public allowToSetMinMaxOnYNA = false;
+    public allowToSetMinMax = false;
 
     public isCreatedList: boolean;
 
@@ -158,6 +158,10 @@ export abstract class BasePollFormComponent extends BaseComponent implements OnI
 
     private get isPollMethodYNA(): boolean {
         return (this.contentForm?.get(`pollmethod`)?.value as PollMethod) === PollMethod.YNA;
+    }
+
+    private get isPollMethodYN(): boolean {
+        return (this.contentForm?.get(`pollmethod`)?.value as PollMethod) === PollMethod.YN;
     }
 
     public get pollMethod(): FormPollMethod {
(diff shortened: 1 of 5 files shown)


rrenkert (approved):

client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.ts
```diff
@@ -300,9 +304,7 @@ export abstract class BasePollFormComponent extends BaseComponent implements OnI
     public showMinMaxVotes(data: any): boolean {
         const selectedPollMethod: FormPollMethod = this.pollMethodControl.value;
         return (
-            (selectedPollMethod === `Y` ||
-                selectedPollMethod === `N` ||
-                (selectedPollMethod === `YNA` && this.allowToSetMinMaxOnYNA)) &&
+            (selectedPollMethod === `Y` || (selectedPollMethod !== `yna` && this.allowToSetMinMax)) &&
```
bastianjoel, line 307:
```suggestion
            (selectedPollMethod === `Y` || (selectedPollMethod !== `YNA` && this.allowToSetMinMax)) &&
```

Should be uppercase ig?

bastianjoel (changes requested):

client/src/app/site/pages/meetings/modules/poll/components/base-poll-form/base-poll-form.component.ts
Elblinator: no yna in lower case is the list option

bastianjoel: Please at least use the enum from `FormPollMethod` to prevent further confusion in the future.

anbebe (approved):
For what reason do we have two enums that are almost identical? (PollMethod and FormPollMethod) Can it be reduced to one? @Elblinator opens a new issue for that topic.

Issue #4566, opened 2025-01-22
Speaking Times widget: Text in moderation note leads to layout error


MSoeb:
**Describe the bug**
The grid layout of the speaking time widget breaks into two columns when text is entered in the moderation note. I don't know exactly how much text is needed for the layout to break. One word is not enough. But I can recreate it with a three-line paragraph. 

Also the matc cards seems to change their position by text change. 

Example - correct visuals:
![Image](https://github.com/user-attachments/assets/0a1d6d23-7fc6-4e28-839f-9e664062119e)

Example - text in moderation note with wrong layout:
![Image](https://github.com/user-attachments/assets/ebdb1ed0-3964-4501-91b3-db69ec96f14c)

**How to Reproduce**
1. open a meeting
2. open setting > list of speakers > parl options > Default speaking time contingent for parliamentary groups (structure levels) in seconds - add '10'
3. open participants > structure levels > create at least 3 structure levels
4. open participants > create three users with different structure levels
5. open agenda > create an item > open the list of speakers add the three users to the list. now the speaking times widget should have three structure levels -> project the item (not the list of speakers)
6. open Autopilot > add "Jemand musste Josef K. verleumdet haben, denn ohne dass er etwas Böses getan hätte, wurde er eines Morgens verhaftet. »Wie ein Hund! « sagte er, es war, als sollte die Scham ihn überleben." in the moderation note. > grid layout should change now.

**Expected behavior**
Inserting a text into the moderation note must not lead to the mat cards being moved or to a change in the speaking time grid layout. The grid should have 3 columns.

bastianjoel:
[#4569](https://github.com/OpenSlides/openslides-client/pull/4569) did not fix the problem.

Elblinator:
I will need some more information in what exactly is amiss. In the following is the same view but with different magnifications.

![Image](https://github.com/user-attachments/assets/2aad94f7-dc3f-460e-9fdb-3032f42a13cf)
![Image](https://github.com/user-attachments/assets/970ff68a-c268-4269-a4ec-48e17eaa8a64)
![Image](https://github.com/user-attachments/assets/d7339f03-d799-41f5-9134-9286eabd1836)
![Image](https://github.com/user-attachments/assets/57b25b50-eae3-4d31-98f7-84a63b2ea6ce)
![Image](https://github.com/user-attachments/assets/96cb6beb-1d16-4704-9a33-ca9cfcf3e4da)
[...]

PR #4795, opened 2025-03-24
Update angular 19

### LINKED ISSUE #4384: Angular 19 update
There is a new angular version!

With the new version `standalone` components are now the default. I think for some parts of our applications it makes sense to use these. Therefore I would suggest to migrate some components, pipes and directives as an example. 

Also there is now a migration to transform constructor based injections to `inject` function calls. This improves extensibility of classes. Therefore I think we should do this change now. 
There are some more transform schmatics regarding signals (we are currently not using them) which we should discuss if it makes sense to apply them with this update. 

Ideally while doing the angular update the eslint version should be updated too as done here: https://github.com/OpenSlides/openslides-client/pull/4084/commits/f14929e03501940cd8dc0b4069b9f60507c80f42. 

TODO (non exhaustive):

- [x] Migrate some components, pipes, directives to standalone (can be done before the update)
- [x] Update Angular
- [x] Update ng2-pdf-viewer to >10.4.x
- [x] Apply further migrations(optional)
- [ ] ~~Replace `ngx-material-timepicker` with new Angular material timepicker https://material.angular.io/components/timepicker/overview~~
- [ ] Apply inject migration
- [x] Update ESLint

bastianjoel:
resolves #4384


bspekker (changes requested):
- [x] If the parliament mode is activatd, the size of the box with speaking times does not adjust correctly: the distance between speaking times should not be fixed.
- [x] Dropdown with checkbox (e.g. supporters in motion create): When checking at least two options there is blank space added in the end of the box.

bastianjoel:
> If the parliament mode is activatd, the size of the box with speaking times does not adjust correctly: the distance between speaking times should not be fixed.

This is not related to the update. See #4566

bspekker (approved):

PR #4821, opened 2025-03-31
Add a max validator to max_votes_amount in same cases

### LINKED ISSUE #4802: Disallow max vote > options
**Describe the bug**
It is currently possible to have more max votes than you have options to select

**Wanted behavior**
Max votes should not be bigger than the amounts of options 

The only exception is if cumulative votes is activated (settings > election) and you have "Yes per Candidate", then max votes should be the same as it is right now

reiterl:
Resolve #4802


bspekker (changes requested):
It is still possible to select more max votes than cadidates for Yes per candidate and No per candidate voting methods.

When cumulative voting is activated, it is also possible to select more max votes than candidates for YN(A) per candidate.

bastianjoel (approved):

bspekker (approved):

PR #4852, opened 2025-04-07
Add checkbox to motion forward dialog

### LINKED ISSUE #4740: Add option to forward motion with/without attachment
This Issue should be worked on after the feature branch motion-forward was merged 
https://github.com/OpenSlides/openslides-client/tree/feature/motion-forward

**Current behavior**
Image

**Wanted behavior**
An option to add the attachments from a motion should be added as a checkbox in the dialog 

The checkbox should be below the status  and above the options to forward to

The text next to the checkbox should be "Forward with attchements"
This checkbox should be false by default

reiterl:
Resolve #4740 

Add checkbox to motion forward dialog which sends a 'with_attachments' to the backend. 
This also needs the backend change of https://github.com/OpenSlides/openslides-backend/pull/2949


bastianjoel (approved):

reiterl:
Merged and tested with the new backend. Forward the attachment worked. @Elblinator please have a look.

MSoeb (changes requested):
The functionality is given. However, the layout and visibility of the function for single and multiple forwarding still needs to be adjusted.

- [x] New position: The attachment export selection should be placed in the table between the identifier and the target meeting. 'Attachments' should appear in the left-hand column. A checkbox should simply be positioned in the right-hand column, aligned in the same vertical line as the lower checkboxes.

- [x] Checkbox: When the dialog is opened, the checkbox has a grey border. The gray area should not appear.

- [x] Logic for single and multiple forwarding: For single forwarding in the motion detail view, the option to forward attachments should only be displayed if there are also attachments availble. For multiple motion forwarding, the option should always be displayed.

reiterl:
Updated the label, too.

MSoeb (changes requested):
- [ ] Forward dialog - position of 'Attachments:' - Attachments should be positioned between the identifier and the target meeting. Not under target meeting

reiterl:
Switched the order of 'attachments' and 'meetings' in the motion forward dialog.

reiterl:
No need for another code review, because of minimal code change.

Elblinator (approved):
code

MSoeb (changes requested):
The dialog in the motion detail view works now as it should.

For multiple selection the dialog does not work as intended. Attachments option is only shown, if a selected motion has an attachment. 

Logic for multiple selection dialog: The attachment option should _always_ be displayed in the dialog. Even if motions are selected that have no attachments.

MSoeb:
Note: There is another issue, that needs to be done, before this issue can be merged. 

https://github.com/OpenSlides/openslides-backend/issues/3008  

Nevertheless, the work here can be finalized. We should only wait for merging. (Discussed and clarified this process with @Elblinator)

reiterl:
If multiple motions are selected in the forward dialog, the with attachment option is always displayed, even when all
motions don't have attachments.
If in multiselect just one motion is selected and it hasn't got attachments, then the with attachment is not displayed.

The motion forward dialog doesn't know, where it is called from at the moment. 
-> needs discussions. @rrenkert @MSoeb

MSoeb (approved):
Spoken with @reiterl and @bspekker about the last change request.

The change request is withdrawn as the effort and benefit of separating the dialogs are not in a reasonable ratio. Furthermore, the current dialog logic - only displaying the attachment option if attachments are also available - follows existing functions that only offer options if they are also available.

Important: As written, the merge should only take place after another issue has been completed.

PR #4867, opened 2025-04-08
Change "FormPollMethod" to "PollMethod"

### LINKED ISSUE #4835: Unify two enums in polls
created for @Elblinator 

**Current behavior**
We have two enums that are almost identical: PollMethod and FormPollMethod

**Wanted behavior**
Check if they can be reduced to one.

**Additional context**
From https://github.com/OpenSlides/openslides-client/pull/4789#pullrequestreview-2713849425

Elblinator:
closes #4835


bastianjoel (changes requested):
The value of `LIST_YNA` is not a valid poll method for example in `poll.ts` field `pollmethod`. 
I guess you cannot just remove this.

PR #4874, opened 2025-04-09
Filter empty forwarded motions from motion list

### LINKED ISSUE #4871: Fix motion-list with (hidden) forwarded motion
**Describe the bug**
If a user can see the motions and is allowed to see their origin then motions are displayed even if they should be hidden 

**How to Reproduce**
1. Create two committees (To and From) with each one meeting
2. Allow "From" to forward to "To"
3. In From:
4. Motion-list > workflow > simple workflow: Allow forwarding of motions
5. Create a motion and forward it to To
6. Go into To everything from now on will be there
7. In Settings > motions: activate "Allow backtracking to forwarded motions"
8. Add a participant  (e.g. Ann)as delegate
9. Allow delegates "Can see origin motion"
10. Go to motions and change the workflow from the forwarded motion to complex
11. Login as Ann
12. Navigate to motion-list

**Expected behavior**
If a motion is in a for the participant restricted state the motion should not be displayed in the motion-list

**Additional context**
In step 10 you could also restrict "submitted" in simple workflow

bastianjoel:
resolves #4871

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/services/common/motion-controller.service/motion-controller.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/services/common/motion-controller.service/motion-controller.service.ts
@@ -39,12 +39,8 @@ export class MotionControllerService extends BaseMeetingControllerService<ViewMo
     /// ////////////////////// Bridge to the dedicated repo
     ///////////////////////////////////////////////////////////////////////////////////////////////
 
-    public override getViewModelList(): ViewMotion[] {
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
If a motion and their amendment were forwarded and the motion is in a state which the user can not see, then the user will not see the motion and the amendment (independently from the amendment's state) 

If the amendment is is a state which the user cannot see then the user can not see the amendment. 

I personally think that is correct: "If the user cannot see the reference motion then they cannot see the amendment"


I tested several cases where the user should/should not be able to see the motion in the motion list.

PR #4876, opened 2025-04-10
Fix problem with poll pdf export and entitled users 

### LINKED ISSUE #4873: Wrong entitled users list in motion poll PDF
* Start new motion poll
* send vote for 2 users
* stop poll
* open entitled users list (see screenshot) -> 2 users have voted (correct)
Image
* make PDF export (see screenshot) -> ALL users listed with "have voted: No" (wrong!)
Image

Additional information:
If you export PDF while voting is still running the entitled users list in PDF is correct.

reiterl:
Resolve #4873 

Only in a poll with status 'started' the live vote register data are available. In the other status the entitled users data should be used.


Elblinator (approved):
code

PR #4880, opened 2025-04-14
4688 fix nested lists

### LINKED ISSUE #4688: Motion > change recommendations: problems with nested lists

bastianjoel:
resolves #4688

### DIFF
--- a/client/src/app/ui/modules/editor/components/editor/editor.component.scss
+++ b/client/src/app/ui/modules/editor/components/editor/editor.component.scss
@@ -94,6 +94,9 @@
                     display: inline-block;
                 }
             }
+            li.os-split-before {
+                list-style: none;
+            }
 
             @include detail-view-appearance.detail-view-appearance;
         }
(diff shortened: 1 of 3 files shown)


bspekker (approved):

Elblinator (approved):

client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-diff.service/motion-diff.service.ts
```diff
@@ -1771,16 +1774,81 @@ export class MotionDiffService {
         }
 
         if (isSplitAfter) {
-            diff = DomHelpers.addClassToLastNode(diff, `os-split-after`);
```
Elblinator, line 1774: please remove `addClassToLastNode` and the things associated with it

PR #4905, opened 2025-04-23
Localize poll result numbers

### LINKED ISSUE #4567: Poll Result visualization: Comma/Point
**Current behavior**
All percentage values for voting and election results are currently displayed in the format 10.123% (an example). A point is used as the decimal separator. 

This is not optimal, as this separation is not common in Europe or in German-speaking countries.

**Wanted behavior**
A) It would be better if we used a comma as the decimal separator to better comply with local European standards. To ensure international standards, we should alternatively offer an option that makes the decimal separator selectable. The user should have the possibility to select the wanted format.

Both enhancement requires new options in a meeting. Therefore a new setting should be integrated under meeting settings > general > new category 'Poll results number format'. The new category should have select options for Decimal separator.

Decimal separator: Comma or Point

bastianjoel:
resolves #4567

### DIFF
--- a/client/src/app/domain/models/meetings/meeting.ts
+++ b/client/src/app/domain/models/meetings/meeting.ts
@@ -30,7 +30,7 @@ export class Settings {
     // TODO: Move to meeting. these are not settings anymore, if the meeting-detail-view
     // in the committee-list-view is finished.
     public enable_anonymous!: boolean;
-    public language!: boolean;
+    public language!: string;
 
     // Jitsi/Livestream settings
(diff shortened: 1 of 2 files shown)


Elblinator:
This PR has nothing to do with the display of motions. 
There seems to be an issue with your instance I'll update it.
PLease test the comma/point from the poll and not the display of the motion

MSoeb (approved):
Looks good.

Elblinator (approved):

PR #4906, opened 2025-04-23
Fix complete change reco content deletion inside list item

bastianjoel:
Fixes an issue when deleting all the content from a change recommendation when editing a list item.


Elblinator (approved):

m-schieder (approved):
OK

PR #4918, opened 2025-04-25
Fix display of invalid supporters

### LINKED ISSUE #4917: Motion supporters: two issues with invalid
**Describe the bug**
Part 1: If a participant who supports a motion is moved to another group that does NOT allow support, then no invalid message will be displayed in the support list. This is incorrect.

Part 2: Based on this, another error occurs. If a participant has transferred their voting right and the support of motions is set as not possible in the voting right delegation settings, this right will continue to be used in the background if the voting right delegation has been withdrawn. In short: Once a voting right has been transferred and then withdrawn again, the settings remain active. However, this is only visible in combination with case 1. 

**How to Reproduce**
Part 1: 
1. open a meeting
2. set everything that is necessary to support motions (group permissions, support motion >0 and workflow). Delegates should be able to support a motion.
3. create two Users A and B as delegates
4. open motions > create a motion and add User A and B as supporters
5. Check now the motion - no invalid supporters should be visible.
6. Now change the group of User B to staff (which has no support permission) 
7. open the motion and check support -> all supporters still valid. Wrong bevahvior. User B has not the permission to support anymore. Therefor the invalid message should appear.

Part 2:
8. Activate now vote delegation AND 'Restrict delegation principals from supporting motions'
9. Change group of User B to delegates
10. Delegate vote right from User B to A. 
11. Check the motion -> User B is not listed as invalid. Correct
12. Undo the vote delegation from user B to A.
13. open the motion. -> No invalid supporters. Correct.
14. Change the group of User B to staff.
15. open Motion and check supporters. -> B is now listed as invalid. This is correct, but in combination with issue 1, makes this so sense, because the bug should occur. This indicates, that the vote delegation is still somehow active in the background.

**Expected behavior**
Issue 1: If a User is changed to a group, which has no support permission and the user is currently a motion supporter, the support should be indicated as invalid. #

Issue 2: If a vote delegation was active and is now undone, the vote delegation should not be active in the background anymore.

Elblinator:
closes #4917

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-meta-data/motion-meta-data.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-meta-data/motion-meta-data.component.html
@@ -43,18 +43,14 @@
             <button mat-button type="button" (click)="openSupportersDialog()">
                 {{ motion.supporters.length }} {{ 'supporters' | translate }}
             </button>
-            @if (validSupporters > 0 && canManage && (allowSupporters$ | async)) {
+            @if (canManage && validSupporters > 0) {
                 <span class="warn">{{ validSupportersText }}</span>
             }
             @if (showSupporters) {
                 <p class="supporters">
                     <os-comma-separated-listing [list]="supportersObservable | async">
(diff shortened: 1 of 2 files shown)


bastianjoel (approved):

MSoeb (approved):
looks good

PR #4933, opened 2025-04-30
Change layout if editorial version is edited

### LINKED ISSUE #4899: Fix margins in editing editorial version
**Describe the bug**
The edit view of editorial final versions of motions uses a margin when the line numbering "outside" is chosen. This is unecessary since the edit view does not show line numbers and the margins left and right should be the same.

**How to Reproduce**
1. open a motion and view in the editorial final version
2. edit the editorial version
3. pick line numbering "outside" and compare to "inside/none"  

line numbering "outside":
Image

line numbering "none":
Image

**Expected behavior**
The margin on the left of the edit field should always be the same as for line numbering inside/none.

reiterl:
Resolve #4899 

Change the layout, if the editorial version is edited. Add an output var to one component, to see if it is edited.


bspekker (approved):

bastianjoel (changes requested):
Why don't you just use `this.viewService.modifiedFinalVersionActionSubject` in `motion-content.component`?

bastianjoel (approved):

bspekker (approved):

PR #4934, opened 2025-04-30
Change color of the buttons in the os-editor

### LINKED ISSUE #4897: Adjust colour of editor buttons
**Describe the bug**
The colour of the buttons in the editor is inconsistent with the rest of OpenSlides since it is the only instance where buttons/text in Openslides is in true black.

**How to Reproduce**
1. open the edit view of a motion/startpage/topic/...

**Expected behavior**
The buttons are in the same grey as all other text in OpenSlides or has the same transparency (usually 0.54 or 0.55).

reiterl:
Resolve #4897

change the color of the buttons of os-editor.

### DIFF
--- a/client/src/app/ui/modules/editor/components/editor/editor.component.scss
+++ b/client/src/app/ui/modules/editor/components/editor/editor.component.scss
@@ -15,7 +15,7 @@
             margin-right: 4px;
             margin-bottom: 4px;
             border-radius: 4px;
-            color: var(--mdc-text-button-label-text-color, inherit);
+            color: rgba(0, 0, 0, 0.54);
             &.text {
                 padding: 0 10px;
             }


client/src/app/ui/modules/editor/components/editor/editor.component.scss
```diff
@@ -15,7 +15,7 @@
             margin-right: 4px;
             margin-bottom: 4px;
             border-radius: 4px;
-            color: var(--mdc-text-button-label-text-color, inherit);
```
bastianjoel, line 18: Did you have a look if there is another variable you could utilize for that instead of hard coding the value?

reiterl: I didn't find any matching theme color, but `--mat-menu-item-icon-color` seem to be the same.
Should use that instead?

bastianjoel (changes requested):
Does not work with dark mode. 

Also the color is way to light. The value provided in the issue is definitely wrong. See for example the upload button in the topic form.

bspekker (approved):

PR #4950, opened 2025-05-07
Fix innerHTML images escaping box

### LINKED ISSUE #4919: Meeting > Start-Menu: Image scaling issues
**Describe the bug**
The image scaling does not work well. If you embed an image on the start page and scale it so that it fits into the matcard, the scaling only works correctly in this view. If you switch to the mobile view, the image is still scaled to fit this view and not the mobile view. The image extends to the right of the screen.

The problem is clearly visible on the start page of a meeting. However, I assume that this is a general problem.

Example with HD screen resolution:
Image

View with a responsive scaling of an iPhone SE:
Image
Image

**How to Reproduce**
1. open a meeting
2. edit the start menu and add a large image. scale it to manually so that it fits the mat card. (tried it with HD screen resolution.)

**Expected behavior**
Image scaling should also work in responsive views. Regardless of the screen resolution, the image should scale correctly so that it can be used responsively.

bastianjoel:
resolves #4919

Also fixes the same issue with topics.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/agenda/modules/topics/pages/topic-detail/components/topic-detail/topic-detail.component.html
+++ b/client/src/app/site/pages/meetings/pages/agenda/modules/topics/pages/topic-detail/components/topic-detail/topic-detail.component.html
@@ -68,7 +68,7 @@
                         @if (topic.text) {
                             <span>
                                 <!-- Render topic text as HTML -->
-                                <div class="detail-view" [innerHTML]="topic.text | trust: 'html'"></div>
+                                <div class="detail-view html-view" [innerHTML]="topic.text | trust: 'html'"></div>
                             </span>
                         }
                     </div>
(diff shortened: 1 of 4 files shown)


Elblinator (approved):

MSoeb (changes requested):
Tested it with an existing meeting a new one.

New one: Uploaded an image and used it as graphic in editor. Works fine.

Existing meeting and existing file: Image is not visible after saving. In edit mode I can add it, but after saving it's gone. Tried to use a new uploaded image. It also doesn't work. I guess the problem is an existing meeting.  --> Fix should be retroactive. 

If you need an example, I can show it to you.

bastianjoel:
This is just a change to the stylesheets. If there is a problem with existing meetings it is not related to this fix or issue.

MSoeb (approved):
Talked with @bastianjoel about it. I've added the error part in another issue #5004 which need to be solved before we build a new staging.

This one looks fine.

PR #5002, opened 2025-05-21
Fix poll detail performance

### LINKED ISSUE #4966: eVoting > Detail view of a poll has different issues
**Describe the bug**
During last congress I've recognized a strange behavior of the poll detail view. I've wanted to check the voting register during a running poll. After a view seconds in the detail view my tab stuck. Also I could not see the persons which have already voted.

Maybe this is related to the progress bar issue. 

**How to Reproduce**
No clear reproduction. I guess it happens during running poll with a lot of users (more than 450 persons).

**Expected behavior**
The detail view should show the persons which have already voted and the tab should not stuck.

bastianjoel:
resolves #4966 

Bug was introduced with #4565


reiterl (approved):
Okay, LGTM.

PR #5009, opened 2025-05-21
Fix poll max votes edit value

### LINKED ISSUE #4986: Ballot: min/max numbers reset if poll is edited
**Describe the bug**
If a voting method with minimum and maximum votes is selected for an election, the set values are reset to 1 when the edit dialog is reopened. 

Edit dialog before save:
Image

information about ballot after saving:
Image

reopened edit dialog: 
Image

**How to Reproduce**
1. open a meeting
2. open elections 
3. create an election with 4 candidates
4. create a ballot with voting method 'yes per candidate', add 3 to max votes. > save
5. edit the poll
6.  -> reopren dialog and check max votes. it is resetted to 1.

**Expected behavior**
Reopening the editing dialog must not reset the set minimum and maximum votes.
IMPORTANT: This applies to ALL selectable voting methods. Please note when correcting.

bastianjoel:
resolves #4986 

There is however still a bug where list poll method is not persisted.


MSoeb (approved):
Works as intended.

Spoken with @rrenkert about the other problem. I will create an Issue for the other problem.

PR #5015, opened 2025-05-22
Dockerfile Rework

Janmtbehrens:
Reworked Dockerfile & Makefiles, so that there is one singular Dockerfile that supports development, testing and production images.
Also adding features like healthchecks and use of restricted app users, as well as fixing some bugs

### DIFF
--- a/client/src/app/domain/models/meetings/meeting.ts
+++ b/client/src/app/domain/models/meetings/meeting.ts
@@ -198,6 +198,7 @@ export class Settings {
     public poll_default_onehundred_percent_base: PollPercentBase;
     public poll_default_group_ids: Id[]; // (group/used_as_poll_default_id)[];
     public poll_default_backend: PollBackendDurationType;
+    public poll_default_live_voting_enabled: boolean;
 
     // SSO
     public external_id!: string;
@@ -448,6 +449,7 @@ export class Meeting extends BaseModel<Meeting> {
         `poll_default_onehundred_percent_base`,
         `poll_default_group_ids`,
         `poll_default_backend`,
+        `poll_default_live_voting_enabled`,
         `topic_poll_default_group_ids`,
         `projector_ids`,
         `all_projection_ids`,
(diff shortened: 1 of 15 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -70,19 +70,14 @@ jobs:
...
-      - name: Check linting
-        run: make run-check-linting
-
-      - name: Check prettifying
-        run: make run-check-prettifying
-
-      - name: Run karma tests
-        run: make run-karma-tests
+      - name: Run Test
+        uses: Janmtbehrens/OpenSlides/dev/actions/build-and-test-service@all-in-one
```
rrenkert, line 81: Please use upstream OpenSlides main here when https://github.com/OpenSlides/OpenSlides/pull/6935 is merged

Dockerfile
```diff
@@ -1,25 +1,74 @@
...
+FROM nginx:1.28.0 as prod
+
+WORKDIR /
+
+# User is given by nginx image
 
 LABEL org.opencontainers.image.title="OpenSlides Client"
 LABEL org.opencontainers.image.description="Web client for OpenSlides which serves as the users main interaction point while using the OpenSlides system. "
 LABEL org.opencontainers.image.licenses="MIT"
 LABEL org.opencontainers.image.source="https://github.com/OpenSlides/openslides-client"
```
rrenkert, line 71: Duplicated labels?

rrenkert (changes requested):

Dockerfile
```diff
@@ -1,25 +1,74 @@
...
+# Playwright Image
+
+FROM mcr.microsoft.com/playwright:v1.52.0-jammy as playwright
+
+WORKDIR /
+
+COPY ./package.json .
+COPY ./package-lock.json .
+ENV CI=1
+RUN npm ci
```
bastianjoel, line 48: I don't think that this should be included in the main `Dockerfile`. The integration tests are not really part of the client.

.github/workflows/continuous-integration.yml
Janmtbehrens: Reference has been changed in newest commit

rrenkert (approved):

PR #5031, opened 2025-06-02
Remove diminish oml warning in account detail

### LINKED ISSUE #5013: Remove wrong error message when editing own admin account
**Describe the bug**
When an accountadmin/orgaadmin/superadmin edits their own account via the account list, they get the following message warning that the edit will change their OML even though the changed fields don't affect these rights:

```
This action will diminish your organization management level
This will diminish your ability to do things on the organization level and you will not be able to revert this yourself.
```

**How to Reproduce**
1. login as accountadmin/orgaadmin/superadmin
2. go to "accounts" and edit your own account
3. edit any field in the personal information card
4. click "save" 

**Expected behavior**
The warning is only shown when a field is changed that actually affects the OML rights: administration roles, committee admin

reiterl:
Resolve #5013 

Rm the diminish oml warning prompt.


bastianjoel (approved):
Not exactly sure but could it be that the issue just asks to add a check if `payload.organization_management_level` is even set? 
If it is undefined the message would appear even though the field would not get changed.

reiterl:
In my tests even a superuser cannot change their omls. So I have removed the whole prompt.
As far as I understand the issue, this was the requested change.

PR #5032, opened 2025-06-02
Translate the default genders in the gender list

### LINKED ISSUE #5012: Accounts: Add missing translations for genders
**Describe the bug**
In the accounts > genders menu the default genders are not translated.

**How to Reproduce**
1. open the account list
2. via the 3-point-menu open "Genders"
3. set your display language to something other than English
--> the default genders are not translated to other languages

**Expected behavior**
The default genders are translated to the language that is currently used.

reiterl:
Resolve #5012 

I didn't translate all gender names, because this could lead to confusion.

### DIFF
--- a/client/src/app/site/pages/organization/pages/accounts/pages/gender/pages/gender-list/components/gender-list/gender-list.component.html
+++ b/client/src/app/site/pages/organization/pages/accounts/pages/gender/pages/gender-list/components/gender-list/gender-list.component.html
@@ -33,22 +33,18 @@
 >
     <div *osScrollingTableCell="'name'; row as gender" class="cell-slot fill">
         <div>
-            @if (gender.id > 4) {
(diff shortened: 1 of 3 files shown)


bastianjoel (changes requested):
This would make the translation of gender names inconsistent. 

When quickly searching for this I found that for example in `user-detail-view.component.html` and `list-of-speakers-entry.component.html`  the gender names get always translated. 
For convenience I would suggest adding a new field `localizedName` to the view model which just handles whether the name should be translated or not. 

While at it please also add a property `isPredefined` to the view model instead of using a magic number.

reiterl:
add the isPredefined in the view. Use it in gender-list, translate all gender names in the gender list.

bastianjoel (approved):

PR #5098, opened 2025-06-12
Add multi select home committee and guest

### LINKED ISSUE #5074: Add multi-select on account-list for home committee and guest
**Current behaviour:**
Currently it is not possible to add accounts to a home committee or give/remove the guest status via multi-select

**Wanted behaviour:**
You should be able to add accounts via multi select 

Used icons: "home_work" and "outbound"
Image

Dialog for home committee only one committee should be selectable
Image

Dialog for external status
Image

reiterl:
Resolve #5074 

Version of the multiselect home committee and guest/external.
Uses the choice Service (like the other multiselect actions) with its caveats.
Doesn't filter the selectable committees.

### DIFF
--- a/client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-list/account-list.component.ts
+++ b/client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-list/account-list.component.ts
@@ -170,11 +170,11 @@ export class AccountListComponent extends BaseListViewComponent<ViewUser> {
 
     public async changeGuest(): Promise<void> {
         const title = this.translate.instant(`Set external status for selected accounts`);
-        const SET_ACTIVE = _(`active`);
-        const SET_INACTIVE = _(`inactive`);
-        const result = await this.choiceService.open({ title, actions: [SET_ACTIVE, SET_INACTIVE] });
+        const SET_EXTERNAL = _(`external`);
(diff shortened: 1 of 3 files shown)


client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-list/account-list.component.ts
```diff
@@ -139,6 +142,43 @@ export class AccountListComponent extends BaseListViewComponent<ViewUser> {
...
+        const result = await this.choiceService.open<ViewCommittee>({
+            title,
+            choices: committees,
+            actions,
+            content: this.translate.instant(
+                `Attention: Accounts will overwrite other home committees if possible and remove the external status.`
+            )
+        });
+        if (result) {
+            if (!result.firstId) {
```
bastianjoel, line 160: Why throw an error when this is not needed for `REMOVE`? 

Additionally, I don't see that this gets handled anywhere. Can this even happen?

bastianjoel (changes requested):

client/src/app/site/pages/organization/pages/accounts/pages/account-list/components/account-list/account-list.component.ts
reiterl: Hmm, yes, I removed the result.firstId error check. This should not happen.

PR #5128, opened 2025-06-20
Move userHeadersAndVerboseNames in the view-user file

### LINKED ISSUE #4990: The userHeadersAndVerboseNames needs cleanup
@bastianjoel wrote in https://github.com/OpenSlides/openslides-client/pull/4912

userHeadersAndVerboseNames in user.constant.ts:
> This needs cleanup. keyof ViewUser would be better here but ViewModels should not be imported in the domain files. As this is just used for import we could place it in some of the modules.

reiterl:
Resolve #4990


bastianjoel (changes requested):
`client/src/app/site/pages/organization/pages/accounts/pages/account-import/definitions/index.ts` or `client/src/app/site/pages/meetings/pages/participants/pages/participant-import/definitions/index.ts` would have been a better location.

PR #5135, opened 2025-06-23
Add change detection push for setData/setting change in motion block slide

### LINKED ISSUE #4997: Projection of motions blocks: sync issues
**Describe the bug**
If a motion block with a few motions is projected, the projection doesn't change after changing the maxium column amount or changes done to meta data of one of the motions in the block. The projection needs a reload to be in sync with the setting.

**How to Reproduce**
1. open a meeting
2. open motions > create 20 motions > put them on a motion block
3. project the motion block
4. open the projection in separate window without putting it in the background.
5. open now in another browser the meeting settings > motions > Projection > Maximum number of columns in motion block projection: Change the value and save. Chek the projection of the motion block. Nothing happens. 
6. open now one of the motions from the block and change recommendation. -> Nothing changes in the projection.

**Expected behavior**
Changing the amount of columns of a motion block or meta data which is part of the projection should have an instant effect on the projection without reloading it.

reiterl:
Resolve #4997

The setData or the setting change didn't push the cd. I updated that.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/projector/modules/slides/components/motions/modules/motion-block-slide/components/motion-block-slide/motion-block-slide.component.ts
+++ b/client/src/app/site/pages/meetings/modules/projector/modules/slides/components/motions/modules/motion-block-slide/components/motion-block-slide/motion-block-slide.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component } from '@angular/core';
 import { getIntlCollatorForLang } from 'src/app/infrastructure/utils';
 import { BaseMotionSlideComponent } from 'src/app/site/pages/meetings/modules/projector/modules/slides/components/motions/base/base-motion-slide';
 import { SlideData } from 'src/app/site/pages/meetings/pages/projectors/definitions';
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/modules/projector/modules/slides/components/motions/modules/motion-block-slide/components/motion-block-slide/motion-block-slide.component.ts
```diff
@@ -91,13 +91,16 @@ export class MotionBlockSlideComponent extends BaseMotionSlideComponent<MotionBl
...
     }
 
-    public constructor(private meetingSettingsService: MeetingSettingsService) {
+    public constructor(private meetingSettingsService: MeetingSettingsService, private cd: ChangeDetectorRef) {
         super();
         this.languageCollator = getIntlCollatorForLang(this.translate.currentLang);
 
         this.meetingSettingsService
             .get(`motions_block_slide_columns`)
-            .subscribe(value => (this.maxColumns = value > 0 ? value : MAX_COLUMNS));
```
bastianjoel, line 100: Normally I would not approve this, as the observable is not unsubscribed. It would be better to provide an observable to the template by modifying the value via `map` to fix this problem.
Anyway, as the projector gets replaced anyway I don't want to bother with this.

bastianjoel (approved):

PR #5153, opened 2025-06-25
Change sorted changes subscription on motion set

### LINKED ISSUE #5044: Amendments: reload necessary to view changes in motion detail view
**Describe the bug**
When another user creates a new amendment the motion detail view gets the info that there is a new amendment but the changes are not shown when ticking "show all changes".

**How to Reproduce**
1. login as meeting admin A and open a motion X in the detail view
2. in a second window: login as a user B with create amendment permissions and create an amendment to motion X
3. as user A: observe how the number of amendments is increased when user B saves the new amendment but ticking "show all changes" does not show the new amendment until a reload

Image

**Expected behavior**
Reloading is not necessary to show all changes.

reiterl:
Resolve #5044

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-view/motion-view.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-view/motion-view.component.ts
@@ -75,6 +75,7 @@ export class MotionViewComponent extends BaseMeetingComponent implements OnInit,
      */
     public set motion(motion: ViewMotion) {
         this._motion = motion;
+        this.onMotionSet();
     }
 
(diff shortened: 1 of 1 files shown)


bastianjoel (changes requested):
The component was just cleaned up. Do not mess it up again with dirty workarounds. 

There are methods like `onMotonUpdated` you should use instead when you need a trigger for updated motions. 
But from the issue description I guess it would be better to fix the observable created in `sortedChangesSubscription`.

reiterl:
I am not sure, why the observables are not working in this context. I tried to debug it and found the following
observation: a set motion of this component happens on after every motion change. I seems like the subscription is no longer working then. This leads to this pr. I have no idea how to continue here, so I closed this pr so that another pair
of eyes look at this issue.

PR #5156, opened 2025-06-26
Allow comulative voting with less options than votes

### LINKED ISSUE #5154: Cumulative elections - creation not possible
**Describe the bug**
Elections with cumulative voting cannot be created because a check is made in the dialog for the value in ‘Maximum amount of votes’ that limits the value to the maximum number of candidates. This is correct for elections that do NOT allow cumulation, but incorrect for elections that allow cumulative voting. With cumulative voting, it is possible to cast a larger number of votes than there are candidates.

Cumulation active - Saving is blocked.
Image

**How to Reproduce**
1. open a meeting
2. open settings > elections > Ballot > activate 'Allow to accumulate several votes on one candidate ("comulative voting")' 
3. open Elections > create a new election
4. open the new election: Add a candidate. Change value in 'Maximum amount of votes.' to 100. -> An error appears, that blocks saving.

**Expected behavior**
The check of the number of candidates for the number of votes in ‘Maximum amount of votes’ must not be executed if ‘Allow to accumulate several votes on one candidate (“comulative voting”)’ is activated in the election settings. The error message is also no longer displayed in this case. If it is not activated, the check should still be made.

reiterl:
Resolve #5154


Elblinator (approved):
function

Issue #5133, opened 2025-06-20
Committee hierarchy: several enhancements/bug fixes for Accounts/participants


MSoeb:
**Current behavior**
New account attributes have also been added with the new ‘Committee hierarchy’ feature. These attributes are currently only available in the account detail view. This is inadequate in several respects. Below are some optimization requests.

Additionally some errors occur.

**Wanted behavior and error description**

- [x] Account list: In the account list, there should be a note for each account indicating whether it has a home committee or an external status. This information is particularly missing if you want to select accounts using multiple selection. It is also not easy to see which home committee an account belongs to. - Suggested solutions: A) An icon is added to the account that shows whether a home committee or an external status has been added. However, it would then still be unclear to which home committee an account belongs. B) The home committee or the external status is entered under the name.

- [x] Account detail view > selection of home committee/external: There is an error with the new attributes. When selecting External, a home committee can still be selected. This is incorrect behavior, as both attributes cannot be active at the same time. Solution: If External is selected, the home committee drop-down menu is grayed out. However, a selected home committee should be retained so that it does not have to be selected again.

- [x] https://github.com/OpenSlides/openslides-backend/issues/3005

- [ ]  The CSV example file for the import should place 'home committee' and 'guest' (should be 'external', see above) should be placed after 'saml_id'

- [x] Participant menu > participant detail view: By opening a participant detail view, an error occurs in the console. I assume, that this happens because of the new attributes of an account. 

![Image](https://github.com/user-attachments/assets/8bc3d08d-83bc-4574-aabc-c9a7505f06f8)

- [ ] Participant menu > participant detail view: New attributes are missing. Solution: Please add the new attributes to the participant view. 

- [ ] Account menu: Add search option for home committee / external

MSoeb:
#5102 <- May be some of the problems will be fixed with this PR.

MSoeb:
#5115 <- Error message should be solved by this.

reiterl:
> Participant menu > participant detail view: By opening a participant detail view, an error occurs in the console. I assume, that this happens because of the new attributes of an account.

This is already reported and worked at in #5115 .

reiterl:
[...]

PR #5171, opened 2025-07-01
Add hc and external to account list in username row

reiterl:
Part of #5133 

Add external and home committee in the small username row in the account list. Implements solution B) of the
point 1. of #5133.


Elblinator (approved):
code

PR #5177, opened 2025-07-04
Change motion poll chart grid

### LINKED ISSUE #2499: Poll results need improvement

reiterl:
Resolve #2499 

In my tests, this patch is an improvement.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.scss
+++ b/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.scss
@@ -15,7 +15,7 @@
 .result-wrapper {
     display: grid;
     grid-gap: 0 2em;
-    grid-template-columns: repeat(auto-fit, minmax(150px, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

bspekker (changes requested):
If the mat card width is 214px, the text goes beyond the card:
 
image

For reference, same size in current stable: 
image

Why 214px? For certain tablets, the poll cards are 214px width by default and we should therefore ensure that polls are displayed nicely on these devices.

bspekker:
Yes, this PR will be closed and the issue needs further conceptual work.

PR #5196, opened 2025-07-08
Update motion export pdf defaults

### LINKED ISSUE #5160: Fix small issues with motion export default selection
**Describe the bug**
There are two small issues with the default selection in the motion export: 1. there is no selection of line numbering and 2. the display of attributes that can't be selected is not correct.

**How to Reproduce**
1. open a meeting with the default settings
2. go to > motions > export 
3. observe the following: 
  - there is no selection for line numbering even though the settings say it should be "outside" (the export with this selection does not contain any line numbers)
  - attributes like motion editors, spokesperson, etc have a checkmark but are greyed out: they should not have a checkmark when they are greyed out (if you click to CSV or XLSX format and back to PDF, the display is correct)

Image

**Expected behavior**
When exporting motions for the first time, the default selection should correspond to the settings.

reiterl:
Resolve #5160 

And unset disabled options.


bspekker (changes requested):
If "outside" is the default for line numbering, it looks like it is selected for the export but if I don't click on it again, no line numbers are exported.

reiterl:
Yes, I can reproduce the problem. I have added the default in another way and cleaned up some code.

PR #5197, opened 2025-07-09
Fix category multi sorting

### LINKED ISSUE #5174: Motions > Categories: Multiple sorting doesn't work
**Describe the bug**
If a user wants to sort multiple motions in a category at once with the 'move' function, it didn't work. An error appears in the console.

Function that didn't work - marked red:
Image

Error message:
ERROR Error: Drop was called without previous drag call

**How to Reproduce**
1. open a meeting
2. open motions and create 4 motions
3. open motions > three-dot-menu > categories: Create a new category named ABC
4. Add all motions to category ABC
5. Open Category ABC detail view
6. Sort motions in ABC > mark motion 1 and 3 by holding CTRL > Click on 'move' > move selected items behind/before last motion --> Error

**Expected behavior**
The move function in the category motion sorting should work correctly.

reiterl:
Resolve #5174

Update the `drop` function. There are three cases, dragndrop single/multi and move button.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/categories/components/category-detail-sort/category-detail-sort.component.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/categories/components/category-detail-sort/category-detail-sort.component.ts
@@ -152,7 +152,6 @@ export class CategoryDetailSortComponent extends BaseMeetingComponent implements
             const selectedChoice = await this.choiceService.open(content, choices, false, actions);
             if (selectedChoice) {
                 const newIndex = selectedChoice.firstId;
-
                 this.sortSelector.drop(
                     {
                         currentIndex: newIndex,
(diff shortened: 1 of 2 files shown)


MSoeb (changes requested):
Additional change required: 'Insert before' and 'Insert behind' doesn't work properly.

In both cases the selected motions aren't placed at the correct position or they don't move at all. 

Example:

Initial sorting - 
image

Selection: 
image

Move - insert behind Antrag 1 - result
image

Move - insert before Antrag 1 - result:
image

In both cases the move doesn't work. 

What should happen:
- Move options hsould place the selected motions at the desired position
- if more than one motion is selected, the motions should be rowed accoring to their position in the listing. Example from above for insert before: New sorting should be Antrag 4, Antrag 6, Antrag 1 and so on.

reiterl:
The `moveToPosition()` function included at least one bug. The choice dialog was missing the calculation from selected item (id) to index in the sorted items list.

bastianjoel (approved):

PR #5216, opened 2025-07-14
Fix vote amount validators

### LINKED ISSUE #5207: Election Dialog > Maximum amount of votes: wrong high value is possible
**Describe the bug**
If an election is created which has minimum and maximum votes, it is possible to enter a higher value for max options if the edit dialog is reopened again after first saving. Values in all election types can than be changed to a vlaue which is not possible.

**How to Reproduce**
1.open a meeting
2. open elections
3. create an election with 4 candidates
4. create a ballot with voting method 'yes per candidate', add 4 to max votes. > save
5. edit the poll > change value for 'Maximum amount of votes' to 5 and save. ->> Wrong behavior.

**Expected behavior**
It should not be possible to enter a higher value in max options than the amount of the available candidates. 

Take care: A higher amount is possible and allowed, if 'cumulative voting' is enabled.

bastianjoel:
resolves #5207


Elblinator (approved):
code and function

PR #5230, opened 2025-07-17
change timer control buttons projector control

### LINKED ISSUE #5225: Timer control buttons should be bigger
The timer control buttons should have the same size as the control buttons in the list of speakers. 

Image
Image

Caspar311:
closes #5225


Elblinator (approved):
code

bspekker (changes requested):
Since the new buttons are bigger the placement of the title and time are weird now. Please make sure that they use the entire vertical space.

PR #5232, opened 2025-07-18
Fix display of amendments in motion detail

### LINKED ISSUE #5044: Amendments: reload necessary to view changes in motion detail view

Elblinator:
closes #5044


bastianjoel (changes requested):
Just add the subscription for the single motion to the combineLatest in `sortedChangesSubscription`.

bastianjoel (approved):

bspekker (approved):

PR #5249, opened 2025-07-23
Show polls on projected agenda-item-list

### LINKED ISSUE #5193: Autopilot: If agenda projected, no poll visible
**Describe the bug**
If the agenda is projected, the Autopilot won't show a running poll. 

**How to Reproduce**
1. open a meeting
2. open agenda > create 1 agenda item
3. open agenda > three-dot-menu > 'project' the agenda
4. open motions > create a motion > create a poll > start poll
5. open Autopilot --> Poll is not visible.

**Expected behavior**
If an agenda is projected, a running poll should also be visible and usable in the autopilot.

reiterl:
Resolve #5193 

There was a hook, which disallowed to see the polls on the agenda item list. I think, this was done with propose.
I have commented it out.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.ts
+++ b/client/src/app/site/pages/meetings/pages/autopilot/components/autopilot/autopilot.component.ts
@@ -45,9 +45,7 @@ export class AutopilotComponent extends BaseMeetingComponent implements OnInit {
     }
 
     public get showPollCollection(): boolean {
-        return (
-            this._currentProjection?.type !== `agenda_item_list` && this._currentProjection?.type !== `wifi_access_data`
-        );
+        return this._currentProjection?.type !== `wifi_access_data`;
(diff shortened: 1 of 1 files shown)


bspekker (changes requested):
When the agenda is projected, the autopilot now always shows the result of the last poll. 
Reproduction: 
1. create a motion
2. create a poll and start it
3. cast a vote
4. stop and publish the vote
5. project the agenda --> the result of the finished poll is visible in the autopilot 

Wanted behaviour: If the poll was finished before the agenda ist projected, the result should not be shown in the autopilot.

reiterl:
As far as I understand the requirements are timestamps needed, when a projection is projected and when a poll is finished/changed it's state. I looked at the model, these data are not there. I see no way to meet the requirements.
My patch is small here (see autopilot showPollCollection). So closed and perhaps someone else has better ideas.

PR #5257, opened 2025-07-24
No live election for us

### LINKED ISSUE #5256: Fix nominal election poll for live_voting
**Describe the bug**
If the setting "Set live voting enabled by default" is activated and you create nominal election poll the following error is thrown: Error: live_voting_enabled only allowed for named motion polls.

**How to Reproduce**
1. Activate  the setting > motions "Set live voting enabled by default"
2. start creating an election
3. set the Voting type to nominal and save
4. The error is thrown

**Expected behavior**
If you create an election poll the marker for live-voting should not be send to the backend

Elblinator:
closes #5256


reiterl (approved):
Looks okay to me.

PR #5261, opened 2025-07-25
Add intervention to speaking times component

### LINKED ISSUE #5050: Parliament mode - Projection of speaking times: Add intervention time
**Current behavior**
Currently the intervention timer is not visible on the speaking time projection. It is only visible in the list of speakers. This not optimal, because this time is crucial from the meeting host to regulate and cut off an intervention.

**Wanted behavior**
The intervention time should also be visible in the speaking times projection. It should be placed at the last position. The displayed time should be the set intervention time. If an intervention is started, the time starts to count down and if the it's stopped, the time should automatically reset to the set default value.

reiterl:
Resolve #5050 

Inputs the los in the speaking time component and calculate a countdown.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/speaking-times/speaking-times.component.html
+++ b/client/src/app/site/pages/meetings/modules/list-of-speakers-content/components/speaking-times/speaking-times.component.html
@@ -35,6 +35,23 @@
             </ng-container>
         </div>
     }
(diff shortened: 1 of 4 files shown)


reiterl:
This pull request includes the os-speaking-times patch so far. I will continue to work on this.

Elblinator (approved):

PR #5264, opened 2025-07-28
Allow to set linebreak with shift+enter in chat msgs

### LINKED ISSUE #5046: Line breaks in Chat Messages
**Current behavior**
If you want to insert a line break in the OpenSlides Chat with the typical keystrokes 'Enter', 'Shift-Enter', 'Alt-Enter', "Ctrl-Enter', the message is sent instead of inserting a line break.
Also when copying text with line breaks, those are transformed to spaces.

**Wanted behavior**
Possibility to insert line breaks with a common keystroke, especially 'Shift-Enter', and in copied text.

**Additional context**
In our meetings we have an anonymous Chat for asking questions about abbreviations/organization/etc. In the Answer Chat we always copy the question, so everyone understands the context of the answer. To separate question and answer in the same message it'd be great to use a line break.
I could also imagine scenarios where there is an organization-chat in which you want to publish a timetable or other structured information where line breaks would be helpful.
Tested on Version 4.2.13 (2025-05-16)

reiterl:
Resolve #5046 

Allow to set linebreaks with shift+Enter in chat messages.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/chat/pages/chat-group-list/components/chat-group-detail-message-form/chat-group-detail-message-form.component.html
+++ b/client/src/app/site/pages/meetings/pages/chat/pages/chat-group-list/components/chat-group-detail-message-form/chat-group-detail-message-form.component.html
@@ -10,7 +10,7 @@
     <textarea
         #chatinput
         autocomplete="off"
-        cdkAutosizeMaxRows="5"
+        cdkAutosizeMaxRows="10"
         cdkAutosizeMinRows="1"
         cdkTextareaAutosize
         matInput


reiterl:
Put in a lot of time into this issue so far. No way to meet the requirement. I will pause here.

bastianjoel (approved):

Elblinator (changes requested):
currently the height is 151px 
it needs to be atleast 200px 


If you add more than five lines then  the text box does not scroll automatically but stays (this behaviour happens for me most of the times but not always, but just adding a new line then a word and the a new line at some point creates this behaviour

Issue #5272, opened 2025-07-29
Motion diff: Issue with copied data in Windows


MSoeb:
**Describe the bug**
When a motion text is copied from a doc document into the OpenSlides Editor, HTML coding is also copied, which is not visible in the editor and cannot be cleaned up with Remove Formatting. If an amendment is then submitted, the entire paragraph is displayed as a diff, not just the relevant section.

I was able to reproduce the error in Windows 11 with Office 365. The error cannot be reproduced in Linux.

Also I've discussed the topic with @bastianjoel. 

File with input data out of the Copy from the doc file:
[Copy.txt](https://github.com/user-attachments/files/21488856/Copy.txt)

**How to Reproduce**
Please contact me for reproduction.

**Expected behavior**
The copying of text should work correctly without additional html code in the background. This should work on Windows 11 (10) operating systems with Office 365 (or older versions).

bastianjoel:
Moved to tiptap paste plugin.

bastianjoel:
Seems to be a problem with the custom tiptap elements we use. Therefore also a issue here.

bastianjoel:
Nevermind.

PR #5280, opened 2025-08-01
Prevent empty spans

bastianjoel:
workaround for #5272


MSoeb (approved):
Works good. 

Tested on Win11 / Office 365 in with a doc and docx file format.

PR #5299, opened 2025-08-06
Agenda export rework

### LINKED ISSUE #5060: Enhanced Agenda export
**Current behavior**
The agenda export is currently only possible to a limited extent. There is a fixed output format for PDF or csv export. The content cannot be customized, as is the case with the motion export, for example. This is not optimal and should be improved.

**Wanted behavior**
The agenda export should be redesigned so that it is visually similar to the motion export and functionally enables a selection of the desired content.

What should be changed and how:

- [x]  "Export as CSV" and "export as PDF" should be removed from the list view
- [x] "PDF" should be removed from the detail view
- [x] A new Option "Export" should be added, the new option should navigate to a new page (like the page for motion export)
   - [x] UI of export menu: The UI/layout should follow that of the motion export (also see comment below). 
   - [x] The export option should be available for the 
       - [x] list-view 
       - [x] the multi-select list view 
       - [x] In the list view > three dot from a single topic
       - [x] the detail view

This Dialog:
- [x] available output formats: PDF and CSV
- [x] export categories: "Content", "Meta information" (only CSV), "Page layout" (only PDF) and "Header and footer" (only PDF)
- [x] If somethings does not exist for the _to be exported_ topic then the option should be disabled (e.g. no topic has an attachments then attachment should be disabled) => even if it is the default

_Content_ elements: 
- Item number (default)
- title (default)
- text (default)
- attachments (as links)  (default, but only available for PDF, hide for CSV)
- moderation notes (default)
- list of speakers
  - for CSV: the amount of open requests to speak (same as in motion)
  - for PDF: the LOS should be exported in the same way as it is presented while you in the client look at it (List of last speakers with the infos there plus the current + waiting speakers with the infos about them)
  - For PDF: Add two buttons. Button 1 = List of speaker (Note: have not spoken yet) and Button 2 = Last speakers (Note: Person that spoken)
- polls (only for PDF, the poll export for PDF should be as the poll export for motions)
- internal commentary

_Meta information_ elements (Important: All elements are only available for CSV export.): 
- duration (default)
- tags
- agenda visibility
- done 
- <s>note</s> 

_Page layout_ elements (only pdf): 
- Table of contents 
- enforce page breaks

_Header and footer_ elements (only pdf): 
- header
- footer
- current date

**Additional context**
- According to new elements above, the pdf and csv exports needs to be enhanced. Please use the existing exports (e. .g single agenda item pdf) as a base.

reiterl:
Resolve #5060 

Draft version so far. UI mostly there, csv export updated.


reiterl:
I think it is a good idea to use the pdfmake table for the lists of speakers and the polls, too. Then the background can be colored with `fillColor`.

reiterl:
Add the non-topic agenda items to the export. The older requirements had that the non-topic agenda item should be excluded. As far as I understand the requirements now, I included them again.

reiterl:
I have added the subitems handling. As far as I can see, I have addressed all open points. Perhaps a first review would be helpful.

Elblinator (changes requested):
In PDF in LoS
- Please remove the structure levels from the speaker names
- Make a space between the comma and the structure level

Make the distance between a topic (if the topic has more than the  title and text and number AND is not just exported because it's the parent of a to be exported topic) 5px bigger

If a topic had a poll once (but the poll was deleted) then the title "polls" is also exported even though there are no exported polls 

If exported with table of contents: 
The following Error is in the ocnsole: 
```
Uncaught (in promise) Page reference id not found
```

If only the subitem (and not the parent) is selected for export then the order in which it is exported is wrong,
 => items: 
1 
2
 2.1
3
if 1, 2.1 and 3 are selected then they are is exported in the following order:
1, 3, (2) 2.1


Please use the margins from  settings -> export for the pdf (left/right/top/bottom)

If in the export options Din A5 is selected then the LoS/ Polls and all divider lines are too long
For polls please just give each otion less space 
For LoS: give every option less space (but state):
 - the start time please give this so little space that it always needs two lines (e.g. 8/1/2025, NEWLINE 1:38:11 PM)
 - do not change the space for the state
 - The curation does not need such a big margin to the right (try 3px for the start)
- Important nuance:  
   - The structure level should in the end still be so big that "Gliederungsebene" can fit in the cell
   - The Speaker should be as big as possible but if that's still rather small in DIN A5 then that needs to happen

Elblinator:
And you should be navigated to the agenda if you pressed exported

Elblinator (changes requested):
- [ ] los and poll needs to have the same length as. the divider llines and the table of content content
	- [ ] the starttime should be in one line if the pdf is exported in DIN A4
- [ ] If only a child is exported then the conntent from the parent is also exported
- [ ] You can press the export button if you can see the agenda but you can only export if you can manage the agenda ( ``` You are not supposed to be here… Authorization Error Required permissions to view this page: agenda_item.can_manage```  -> you should be able to export if you can see
- [ ] the distance between two polls should be bigger

reiterl:
I reworked the updateAvailableExportOptions to use the id fields and subscriped them, Wait for the normal subscription `agenda_list` before update the options just once. In my tests, this worked for all disable options with
minimal changes to the agenda_list subscription and not large front load of all objects. It is a tradeoff.

Elblinator (changes requested):
This CR was actually made by, @MSoeb and co.
 The PDF export should be changed:

- In the LoS
- [x] structure level and state as their own rows should vanish
- [x] The name which was there before (with the structure levels in brackets) should be there again
- [x]  if the speakers has a state, the state should be in the name row in cursive, but in a new line, if the speaker does not has a state this line should not be there
- [x] The Order of duration and start time should be swapped
- [x] In DIN A4 the start time should be as small as it can go 
- [x] Duration should be renamed to Speaking times
- [ ] If a waiting speaker is speaking for a structure level the remaining time should be displayed in the duration


For the rest of the export

- [x] The Line from a topic child should not be displaye
- [x]  The line from a topic should be below the whole topic and not above
- [x] The Line from a topic, the poll and the LoS should be as small as the line from a child is right now
- [x] All topics should have a big more distance to the topic below 
  - only for the actively exported topics (not the parents if a child is exported and not the last topic from the list)
 -  [x] If a motion or a motion block is exported it should look like " 'Topic number' 'Motion number' Motion: 'Motion title (meaning between the motion number and the motion title should be the word "Motion:", same for motion block minus the motion number

reiterl:
I worked on the last CR. I think, I have solved every point so far. It needs a test.

Elblinator:
@emanuelschuetze also said the the divider line between tops now should just be gone

reiterl:
It needs test and code review.

bspekker (changes requested):
It already looks really great!

Two small improvements:
- [ ] in the titles "motion" should be place before the number of the motion ("Motion 03: ..." instead of "03 Motion: ...")
- [ ] titles of subitems need some padding before them (if they have no further content it looks weird that there is no space at all between several titles)

reiterl:
Some open points chatted to me:
- First column in los export should include be wider
- speech state displayed in los should start with a capital first letter
- Internal commentary should be renamed into Comment
- the start time column in los should be as small as possible (It is, no change)

client/src/app/site/pages/meetings/pages/agenda/modules/list-of-speakers/view-models/view-speaker.ts
```diff
@@ -46,6 +46,22 @@ export class ViewSpeaker extends BaseHasMeetingUserViewModel<Speaker> {
...
+            case SpeechState.INTERPOSED_QUESTION:
+                return `Interposed question`;
+            case SpeechState.INTERVENTION:
+                return `Intervention`;
+            case SpeechState.CONTRIBUTION:
+                return `Contribution`;
+            case SpeechState.CONTRA:
+                return `Contra`;
+            case SpeechState.PRO:
+                return `Pro`;
```
bastianjoel, line 61: Add translation markers

client/src/app/site/pages/meetings/pages/agenda/pages/agenda-export/components/agenda-export/agenda-export.component.ts
bastianjoel, line 1: Would have been better to create an abstract class for the exports, as this is now very similar to the motion export. 
But I guess that is too late now.

client/src/app/site/pages/meetings/pages/agenda/modules/list-of-speakers/view-models/view-speaker.ts
reiterl: Done

client/src/app/site/pages/meetings/pages/agenda/pages/agenda-export/components/agenda-export/agenda-export.component.ts
reiterl: Yes, I guess so, too.

PR #5340, opened 2025-08-22
Fix CRs for paragraphs with empty first line

### LINKED ISSUE #5164: Fix motions with empty br at the beginning of p
**Describe the bug**
If an amendment with the motion text ```<p><br>text</p>``` deletes everything then the amendment is on line zero and the whole thing breaks the motion

**How to Reproduce**
1. Create a motion with the following text 
```
<p>text</p>
<p>
<br>more t</p>
<p>lala</p>
```
2. create an amendement to line 2 and delete everything
3. the amendment will start at line zero
4. If you show all changes in the motion then everything up to the broken amendment vanishes and there are errors in the console 

motion:
Image

motion while "Show all changes" is active
Image

**Expected behavior**
The amendment should not start at line zero and the amendment should not break the motion view 

**ADDITIONAL INFORMATION:**
As you can see in the screenshot of the motion the br line has no number, giving this line a number could potentionally already solve the issue 
@bspekker  is this a desired behaviour?

### LINKED ISSUE #5165: Fix CRs gor paragraphs with empty br
**Describe the bug**
If a motion has ```<p> <br> text</p>```

**How to Reproduce**
1. Create a motion with the following text 
```
<p>
<br>first</p>
<p>
<br>secoond</p>
<p>third</p>
```
2. create a CR on line 1 and add a new character
3. The motion will look at like the following 
Image

**Expected behavior**
Line 1 should not appear two times (the empty line should not be there)
Below the "first fg" is a lot of space which should not be there

(see line 3)
Image

### LINKED ISSUE #5355: Motion diff compatibility improvement
Currently we are restricted in doing some changes to the motion diff algorithm as they would make existing motions unusable. One example for this is  #5340. 
While migrations are in some cases possible it is not always trivial what to do in case of a conflict. Also as migrations are run from the backend the diff algorithms is not available and therefore only limited testing on the results is possible. 

To improve this situation we want to save the diff algorithm version a motion was created with to be able to do migrations on motions after changes to the diff algorithm. 

The following things need to be done
 * Add a new `diff_version` field to motions
 * Motion create should set the `diff_version` to version from package
 * Add module `DiffCompat` to `openslides-motion-diff`
   * Compatibility check
   * Method returning `LineNumbering` and `HtmlDiff` module with adjustments for compatibility
   * Migration methods
 * MotionDiff/LineNumberingService adjustments

## DiffCompat
`diff_version` should be specified by the lead motion. Amendments should use the version specified by the lead motion.
Most of the time there will be three ways to migrate a motion text - either keeping the resulting html and changing line numbers, by changing the resulting html or by skipping a change to the diff algorithm. 
In some cases we might want to let the user choose. This might be useful when requiring a change for a bug fix that would otherwise be skipped.

bastianjoel:
resolves #5164
resolves #5165
resolves #5355
reoslves potentionally https://github.com/OpenSlides/OpenSlides/issues/6927

Needs:
autoupdate in main
backend in main


Elblinator:
Do not merge yet

Elblinator (approved):

luisa-beerboom (approved):

luisa-beerboom (approved):
Code looks fine, minor remarks though

MSoeb (changes requested):
Found two Issues:

- <s>Projections of motions not possible (I guess it's because of the new projector service)</s> -> projection have their own repository now
- <s>Found a failure in relation to issue #5165 . If I build a motion in the mentioned way and create a change recommendation to line one, two empty lines appears in the CR editor. The motion in normal view and the edtior of the motion instead show only one line. This also happens in empty line three. -> CR should not have two lines. </s> -> is in the second part of my comment

luisa-beerboom (approved):

MSoeb (changes requested):
Existing motions that contain blank lines without line numbers will be assigned line numbers when the PR is applied. This results in incorrect line numbers in existing motions.

This should not happen. 

I’ve discussed this with Bastian and Raimund.

MSoeb (approved):
Missing line value is fixed. No new line numbering will be added.

Info: The height gap between two paragraphs in this case is equal to the one which will be added in new motions by hitting enter two times. If this will be relevant, we have the option to change it at a later point. For now the problems are fixed and the line numbers are consinstent. So we can go on.

PR #5352, opened 2025-08-27
Add package to app.

rrenkert:


bastianjoel:
What are you trying to achieve with this?

rrenkert:
I'm really not sure if this is the best way, but tried to quickfix it for other devs.

bastianjoel:
I guess this happens when starting the dev setup with a Docker image from current main and then switching branches to a state where the `packages` directory does not exist. 

This happens because `/packages` gets mounted by docker compose. 
If this is the case this should not be merged.

PR #5354, opened 2025-08-28
Inline pdf attachment export for motions

### LINKED ISSUE #4798: Allow to export PFD inside of other PDFs
**Current behavior**
If you export a motion as a PDF and the motion has attachments then the attachments are not exported at all

**Wanted behavior**
If the attachments are PDFs then they should be visually exported as well 

**Questions:**
- Are only PDFs wanted?
- Should there somewhere be an option to say that the PDF should not be exported
- Should the PDF be fully DINa4 or should they be "inside" of a motion

bastianjoel:
resolves #4798 
closes #5274


client/src/app/site/pages/meetings/pages/motions/pages/motion-export/components/motion-export/motion-export.component.ts
```diff
@@ -447,6 +456,12 @@ export class MotionExportComponent extends BaseComponent implements AfterViewIni
         }
     }
 
+    public deselectIncludePdfAttachmentsOption(): void {
+        if (this.tableOfContentChip.selected || this.pageChip.selected) {
+            this.deselectOption(`content`, `includePdfAttachments`);
+        }
+    }
+
```
Elblinator, line 464: remove or use function
Also: here ToC and pageChip is used above Toc and continousText is used

Elblinator (changes requested):

MSoeb (changes requested):
Please make the changes:

If "Include..." is clicked:
- [x] Page layout options Table of contents, Continous Text should be deactivated. (I think, the other options hsould work even with pdf file export.)
- [x] Header and footer: Page numbers should be deactivated. Header and current date for the normal motion pdfs should stay. 

- [x] Option for "Include" should only be available, if a PDF file is attached to a selected motion. (Follows logic of other options)

- [x]  Option "include" should not be a default selection. (If this currently integrated ignore this point.)

Term changes: 
- [x] Attachments --> Attachments (URL)
- [x] Include.. -> Attachments (PDF embedded)

Other problems which can, if possible should be solved:
- [ ] Voting result: Should only be available, if a selected motion have one. If available, not preselected, but open for selection. If not, not possible to select.
- [ ] Change of text version: the text option in content is deselected, if text version is changed. -> Text should stay selected.

For the record:
Other problems which we have seems to solved with this PR. 
- Options which are not available in a motion are correctly not choosable. (Supporters, Motion editors, speaker, recommendation, Category, Tags, Motions block, linked motions). In the live version these options are correctly not choosable, but the visual set them nevertheless as chosen and preselected.

bastianjoel:
> Option for "Include" should only be available, if a PDF file is attached to a selected motion. (Follows logic of other options)

We do not load all attachment data on the motion list. Therefore this is now only disabled when there are no attachments at all.

MSoeb (changes requested):
- [x] Export dialog: If a motion doesn't contain an attachment, the export dialog still shows the marked option, even it is not choosable anymore. 

--> Wanted: All options which are currently not available for export should be in the visual scheme of the Attachments (URL) button. See image below.

Screenshot_20250905_134749

MSoeb (changes requested):
The snackbar currently flickers, display multiple times and shows quite late the info that a PDF is exported. 

Wanted: Only one snackbar with the pdf export text should be shown.

Elblinator (approved):
code

MSoeb (changes requested):
Flickering of snackback during export with attachments happens still.

client/src/app/site/pages/meetings/pages/motions/services/export/motion-pdf-export.service/motion-pdf-export.service.ts
```diff
@@ -52,6 +59,48 @@ export class MotionPdfExportService {
         this.pdfDocumentService.download({ docDefinition, filename, metadata, exportInfo });
     }
 
+    /**
+     * Exports multiple motions to a collection of PDFs including pdf attachments
+     *
+     * @param motions the motions to export
+     */
+    public async exportMotionCatalogWithAttachments(
```
Elblinator, line 67: does this need to be public?
```suggestion
    private async exportMotionCatalogWithAttachments(
```

Elblinator (approved):
code

MSoeb (changes requested):
**Additional information: Tried it on an outdated build. I will do a retest again.**

Heavy performance issue with meetings that contains a large amount of motions with pdf files. If a user tries to export content in such a case the tab will break and a lot of errors with net::ERR_INSUFFICIENT_RESOURCES will occur. 

It's not usable. We have to find a solution for these cases.

Even the debugging crashes.
image

rrenkert:
> Tested with recent version in Chromium and Firefox with 1151 motions with a lot of pdfs. Positive: The new snackbar with a progress bar is very nice.
> 
> * Chromium: After around 450 motions the tab crashes and stops working. I got a Sigill error.
> * Firefox: at 512 motions the action stops and the snackbar freezes.
> 
> Conclusion: We still need a solution for large meetings with a big amount of motions. Currently it's not usable.

With such a huge amount of motions (and attachements) we have to think about a server side implementation of the pdf export. Since we need an implementation of the motion diff algo that is usable in the backend, this is not implemented quickly.

@MSoeb you can try to find a limit of motions on your machine to get a feeling what is possible with the client side solution.

Elblinator:
I restricted the feature for <= 100 motions and <= 100 PDF
@MSoeb please test if this is still too much and needs to be lower

rrenkert (approved):

MSoeb (changes requested):
One change required:

Please add a help text under the "Attachements (PDF embedded)' button with text 'Export limit 100 motions'

Rest works as intended.

MSoeb (approved):
Looks fine. 

For documentation: The new export option is limited to 100 motions / 100 attachments. If one of the options is fullfilled plus 1, the export can't be executed.

MSoeb (approved):
Second test: Missed something by testing. Re-Test. Works fine.

PR #5368, opened 2025-09-04
Hide table border if motion slide meta box background disabled

### LINKED ISSUE #5366: 'Make background color from meta information box on the projector transparent' setting: integrate more projection parts
**Current behavior**
If the setting 'Make background color from meta information box on the projector transparent' is activated, not all content on the projection is transparent. Background of amendment information and a line between motion submitter and change recommender ist missing.  

Image

**Wanted behavior**
The white line and the background of the amendment info (gray) should be transparent too, if the setting 'Make background color from meta information box on the projector transparent' is activated under settings > motions > projection.

bastianjoel:
resolves #5366


MSoeb (approved):
Works fine.

PR #5369, opened 2025-09-04
Change poll projections

### LINKED ISSUE #5364: Change poll projection
**Current behavior**
<details>
<summary>ballot:</summary>
Image

</details>

<details>
<summary>motion:</summary>
Image
</details>

<details>
<summary>topic:</summary>
Image
</details>

**Wanted behavior**
General: 
- The displayed table should start at the same height as the title
- the table should be longer
- the pie shart should be smaller

ballot: 
- nothing different (but the ballot is only interesting if there is only one option (one candidate or list-option)

motion: 
- "Votes" should vanish

topic: 
- "Votes" should vanish
- add a grey line and the valid votes

Image

**Additional context**
This is a kinda continuation of https://github.com/OpenSlides/openslides-client/issues/2499

reiterl:
Resolve #5364 

Changes the poll projections of:

- assignment polls (single candidate)
- motion polls
- topic polls

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.scss
+++ b/client/src/app/site/pages/meetings/modules/poll/components/single-option-chart-table/single-option-chart-table.component.scss
@@ -115,6 +115,18 @@
             border-top: 2px solid rgba(128, 128, 128, 0.2);
         }
 
+        .votesinvalid {
+            font-size: 24px !important;
+            font-weight: 400 !important;
+            line-height: 0.8;
+        }
+
+        .votescast {
+            font-size: 24px !important;
+            font-weight: 400 !important;
+            line-height: 0.8;
+        }
+
         .text {
             &-small {
                 font-size: 12px;


Elblinator (changes requested):
This change request  is from @MSoeb and co:

- [ ] In election if the candidate has structure levels they should be in the next line 
  - if they would go over the "%" from the percentages then they should be ellipsed
- [x] The poll should be bigger (the height should exceed the rows for Yes, No, Abstain a bit 
- [x] Contrairy to the mock-up the "Valid votes" should be the same font-weight as the rest
- [x] The column for the percentages should always have the length of the percentage as if there is one option with three decimal places
- [x] The brackets should vanish around the percentages
- The placments of candidate (only election), poll-data and the pie: 
  - [x] The candidate should be above the poll-data and the pie
  - [x] The poll data should be as much left as it is right now and below the candidate (and the title from the poll) 
  - [x] The poll should also occupy more space to the right as it is right now
  - [x] The pie should be as much right as it can go and should be more up in comparisson to the poll-data (it should start before the pie-data visually starts
     - I think poll-data and pie should have "space-between" 
     - and the poll-data should have a margin-top of 5px or the pie should have  margin-top -5px so that the pie is higher, but feel free to implement that differently from what I just said

reiterl:
> Contrairy to the mock-up the "Valid votes" should be the same font-weight as the rest
In topic poll projection Valid votes is the same font-weight as the options
In the single-option-chart table for sildes the font-weight is same too. 
So fixed.

reiterl:
> In election if the candidate has structure levels they should be in the next line
    if they would go over the "%" from the percentages then they should be ellipsed

I don't understand that. If there is a structure levels line, it cannot "go over" the percentages.

reiterl:
This issue is about poll projections. I suppose, that *non-projection* polls should stay the same.
Including *non-projection* poll displays into this issue would enlarge this issue. 
Excluding *non-projection* poll displays from this issue would lead to more complex code.
I don't see a solution.

Elblinator:
> > In election if the candidate has structure levels they should be in the next line
> > if they would go over the "%" from the percentages then they should be ellipsed
> 
> I don't understand that. If there is a structure levels line, it cannot "go over" the percentages.

if the structure level is longer than the percentages
I am not sure how far out tpercentages are but if they are 500px long then the structure level should be ellipse if they reach that point

reiterl:
In the poll projection of an assignment:
- the candidate name is above table and pie
- the table is a far left as possible
- the pie smaller now, so on the right side of the pie has more space
- As the pie is now smaller the 5px would not fit.

The structure levels of a candidate are not in the data from the autoupdate service, and it is already murky here,
This will need a autoupdate change in my opinion.

reiterl:
Adjusted the pie and the grip-gap between pie and table.

bspekker:
As discussed with @emanuelschuetze a few adjustments are still necessary, please also refer to the mock-ups below:

- [ ] the font should be bigger (for example increased from 28px to 32px)
- [ ] the font-weight should be 500 (instead of 400)
- [ ] The first column containing the yes/no/abstain should be bigger so the empty space to the right is used more. "Gültige Stimmen" should always fit in one line.
- [ ] the results table and pie chart should have the same padding to the left and right in comparison to the divider below the title. The position of the pie chart should be oriented to this line to the right (height is already correct)
- [ ] for assignments: the name of the candidate should be in the first line, the structure level in the second. They should be placed above the table/chart (i.e. not indented). Both name and structure level should be ellipsed if they don't fit in the line (note this is slightly different to previous requirements by Loki)
- [ ] If the title of the election is longer than one line, it should be ellipsed

Mock-up for motion poll projection:
1920x1080 projektor view
Mock-up for assignment projection:
1920x1080 projektor view (1)

reiterl:
Analog elections have a different projection, I won't work on the here.
Updated the font-weight, font-size and first column min-width in the single-option-chart-table for slides.

reiterl:
Add ellipsis-overflow class to title and subtitle of a poll slide. This is the requirement. Not only assignment poll slides. 
I talked with @bspekker about it.

bspekker (approved):
Looks good!

PR #5371, opened 2025-09-04
Fix motion export default setting of text for all versions

### LINKED ISSUE #5317: Fix motion export default setting of text for all versions
**Describe the bug**
Text is preselected in the export dialogue only for the diff version. Changing to other versions, deselects text. 
(The text will still be exported. Selecting and unselecting and then exporting will export without text.)

**How to Reproduce**
1. Open the export dialogue: motions > 3-point-menu > export
2. observe that diff version and text is selected
3. select original version
4. text is not selected

**Expected behavior**
Text should always be selected by default for all text versions.

reiterl:
Resolve #5317 

Remove the callback which deactivate the text field.


reiterl:
I will look at this stuff again. @bastianjoel said, that perhaps a single amendment, could be a problem.

reiterl:
Tested some more, there is an error with disabled and selected fields now.

Elblinator (approved):
I tested the functionality together with @bspekker

PR #5372, opened 2025-09-04
Send publish and anonymize in single request

### LINKED ISSUE #5315: Fix flickering in live voting projection at stop, publish & anonymize
**Describe the bug**
When stopping, publishing & anonymizing a nominal vote with live voting, the projected votes flicker from running state to published nominal votes to the anonymized poll result with chart. It is hard to reproduce and see this artifact with very few voters but it is visible for noticable time with 100 voters.

 This video shows the flickering with 16 voters:

https://github.com/user-attachments/assets/1f8d308e-9283-40f0-9b5c-dce3addecb26

**How to Reproduce**
1. create a poll with nominal votes and activate live voting
2. start the poll and cast several votes (proxy votes or other magic)
3. project the single votes 
4. stop, publish & anonymize the poll and look at the projector 

**Expected behavior**
The published results is not visible and the projection goes directly from running vote to anonymized results chart.

bastianjoel:
resolves #5315


Elblinator (approved):
code

PR #5402, opened 2025-09-12
Fix Projection of Live Voting: visual bug (title) 

### LINKED ISSUE #5400: Projection of Live Voting: visual bug (title)
**Describe the bug**
The The poll title contains a capitalised ÄÖÜ the dots from the letters vanish

**How to Reproduce**
1. navigate to motion-detail
2. create a nominal and live voting poll qith ÄÖÜ as title
3. project the poll as Single votes
4. the ÄÖÜ are cut of in the projection

**Expected behavior**
The Ä¨ÖÜ should be completely readable

Elblinator:
closes #5400


luisa-beerboom (approved):
Code looks good, though there's one caveat I'd like to add:

This looks at first glance like it may push the table downwards a bit. Idk if it actually is the case since I haven't tested, I don't know how far it is. It should therefore be checked if this breaks the dynamic table formatting by pushing the table down so far that it flows out of the slide at the bottom, because it thinks it still has space and doesn't format yet.

If the table remains unaffected, however, this LGTM

PR #5403, opened 2025-09-14
Adding buttons to sort assignment candidates by first/last name

### LINKED ISSUE #5391: Enable sorting candidates by first/last name
I'm always frustrated when I have to sort the candidates by first/last name in an election using drag and drop.

I want a button to chose if i want to sort the candidates by first or last name.

xela1601:
# Enable sorting candidates by first/last name

Resolves #5391 

Since, I was not really able to test this with a local setup, it would be nice if the one who does the review can do this.


Elblinator (changes requested):
- a subscription is missing so that the new data is updated without a reload
- The current buttons should behave and look like hte sort options in different lists (e.g. participants)
   - You should press "sort and then the sort menu opens where you can press "Given name" or "Surname" 
   - the sort should be possible to sort upwards and downwards

luisa-beerboom (changes requested):
This doesn't fulfil the linter style requirements.

Quoting our readme:
```
Code can be cleaned and aligned automatically using `npm run-cleanup`.
This will take care of code alignment, import sorting and quotation marks.
To execute this inside the docker container, you can either use `make run-cleanup` while the client
container is already running or `make run-cleanup-standalone` if it's not.
```
Alternatively - if the latter doesn't work - you can try to run `npm run cleanup` in the `openslides-client/client` folder

xela1601:
Hey there @Elblinator and @luisa-beerboom 

First of all - thanks for taking the time for the review.
And sorry for my late reply.
Let's first talk about my local setup.
I cloned the main repo and all the git submodules and i also have setup openslides-client to use my fork.

> You can read the install here: https://github.com/OpenSlides/OpenSlides/blob/main/DEVELOPMENT.md ( the most important things are: `git clone --recurse-submodules git@github.com:OpenSlides/OpenSlides.git` `git submodule update --init` `make run-dev`
> 
> To pull every service to main you need to run "$ make services-to-main" on the main-repo


ok so i was trying to set this up on my macOS device exactly as described in the readme but i have the following problems with the described steps:

so first of all the `sed -i` command in line 23 of the Makefile fails:

```bash
$ make run-dev
sed -i "1s/.*/go 1.25.0/" dev/docker/workspaces/*.work
sed: 1: "dev/docker/workspaces/a ...": extra characters at the end of d command
make: *** [build-dev] Error 1
```

I found out that there is a difference in behavior when using `sed` with the `-i` flag on GNU sed vs. BSD sed (https://www.baeldung.com/linux/gnu-bsd-stream-editor)

So this can be resolved by detecting the os and building the full command dynamically based on that:

```bash
UNAME_S := $(shell uname -s)
ifeq ($(UNAME_S),Darwin)
	SED_INPLACE = -i ''
else
	SED_INPLACE = -i
endif

build-dev:
	sed $(SED_INPLACE) "1s/.*/$(GO_VERSION)/" $(DOCKER_PATH)/workspaces/*.work
	chmod +x $(SCRIPT_PATH)/makefile/build-all-submodules.sh
	$(SCRIPT_PATH)/makefile/build-all-submodules.sh dev
```

I think this is a kind of change that more people in the dev team or even future devs could benefit from, as the number of devs using mac-os devices is increasing. If you're interested i can do another PR for this :)

Back to the original problem:

When i run `make run-dev` 3 of the containers are exitting with errors:

The postgres and the redis container are logging:

`exec /usr/local/bin/docker-entrypoint.sh: exec format error`

the datastore-writer container logs `exec ./entrypoint.sh: exec format error`


> To set up your repo for the client see: https://github.com/OpenSlides/OpenSlides/blob/main/DEVELOPMENT.md#work-in-submodules

So now about setting up only the client:

I run these commands within the openslides-client repo:

```bash
docker build . -t client-dev --build-arg CONTEXT=dev
docker run -it -v ${PWD}/client/src:/app/src -p 4200:4200 --rm client-dev
```

The container started on port 4200 logging:
```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2025/09/28 15:33:31 [warn] 1#1: duplicate extension "js", content type: "application/javascript", previous content type: "application/javascript" in /etc/nginx/nginx.conf:20
nginx: [warn] duplicate extension "js", content type: "application/javascript", previous content type: "application/javascript" in /etc/nginx/nginx.conf:20
```

when i try to connect to it via `localhost:4200` (btw. the `DEVELOPMENT.md` file says it should run on port 4201) I get a connection reset:

```bash
curl localhost:4200
curl: (56) Recv failure: Connection reset by peer
```

What did work was to run the `make run-dev` script in the `openslides-client` repository.

But in this case i was not really able to do any testing in the UI, because it complained about using http and not https:

image

xela1601:
> * a subscription is missing so that the new data is updated without a reload

It's very hard for me as being completely new to this codebase to find this out by myself.
Probably you're talking about the usage of the `updateSubscription()` method of `BaseUiComponent`, right?
What needs to be subscribed to what in this context? Or maybe - to challenge me - you can give me some resources that point me to the right direction? 
Would also be nice to understand the bigger picture of this subscription-based architecture in this codebase.

xela1601:
@Elblinator 

> * The current buttons should behave and look like hte sort options in different lists (e.g. participants)

I was not able to find any kind of sorting button in the participants list. Would be helpful if you can point me to the exact piece of code that you are referencing :)

Elblinator:
I am sorry for the confusion with the setup.
You need to run ```make run-dev``` in the main-repo, and wait until your instance is build. You can navigate to your instance using https://localhost:8000/
If you run this comand directly in the client no other service ist started and hence, you have an empty client.

To work with the client (change stuff etc.) you need the steps from 
https://github.com/OpenSlides/OpenSlides/blob/main/DEVELOPMENT.md#work-in-submodules
After you did that you need to open the repositoy in an editor of your choice and then change the code

xela1601:
> I am sorry for the confusion with the setup.
> 
> You need to run ```make run-dev``` in the main-repo, and wait until your instance is build. You can navigate to your instance using https://localhost:8000/

Yeah - that's exactly what I did. Unfortunately some of the containers exited right away as described above. :(

xela1601:
> That would be very lovely! The PR needs to be in the main repo

-> https://github.com/OpenSlides/OpenSlides/pull/6993

Elblinator:
Sorry for the rather short CR (change request), I'll eloaborate more, if anything is still an unknown just ask away :) 

- a subscription is missing so that the new data is updated without a reload
Our subscriptions are requests to the autoupdate (AU) to give the client the information it should display. After the subscription is made to the AU the autoupdate is giving the client every change in information. For example you look at a motion (https://demo.openslides.org/13/motions) and then someone changes the submitter then, two people decide they want to support the motion and then the status is changed. All these changes would (and will) change the appearance of  the motion automatically
Going to an easier side for the next explanation: https://demo.openslides.org/13/mediafiles 
If you look into the dev-console (f12) you will see the following
```[autoupdate] new request: mediafiles_list:subscription``` and ```autoupdate] from streams: 322884471 - mediafiles_list```
In the ```new request``` you open the subscription to the AU, so you tell the AU which news it should give you if it has some for you 
In the ```from streams``` you get the info from the AU you requested and are allowed to see
If you look deeper in what the console is giving you there you will be able to see which fields you requested and which info you got returned.
Now to the code side of things: 
Almost all subscriptions you can find in the files ```sthsth.subscrition.ts``` (ther are some exceptions)
To find the correct file you should search for the string between "new request:" and ":subscrition"
So for the subscrition in the mediafiles list "mediafiles_list"

Our subscritions are build like this: 
ids/idField = the thing you are looking at
fieldset = the fields you want to subscribe to 
follow = somethings in side of the thing you also want to look at  (inside of the follow is a list of dict
As you can see the subscribtion are shallow and if not stated otherwise only one level deep.
If you describe to something which is impossible for the AU to give you, there is a snackbar error visible for you (you do not need the console)

To figure out which field(s) need a subscription you can use try and error, look at the models.yml (https://github.com/OpenSlides/openslides-meta/blob/main/models.yml)  to figure out which fields you can subscribe and look at different subscriptions and what they sbscribe

Elblinator:
Also your setup problem is neither ignored nor forgotten, we will answer as fast as we can

Janmtbehrens:
Hey @xela1601, I've tried to run the commands you have described and largely got the same output as you. 

> ```shell
> docker build . -t client-dev --build-arg CONTEXT=dev
> docker run -it -v ${PWD}/client/src:/app/src -p 4200:4200 --rm client-dev
> ```

I got the same error as you did. The solution is to add `--target "dev"`  to the build command. As it stands, the build command will build all context stages, including production and testing, which is probably what causes the duplication errors. If you specify the target as being dev, you'll see the build will only take half as long and the subsequent running of the container will not cause errors.
Curling `localhost:4200` caused the same error to happen though. I don't know if this is intended behaviour or not (I don't usually work on the client), but as @Elblinator mentioned, client requires a lot of other containers to be up and running to work correctly, so it might reject connections if it's not setup via `make run-dev` inside the main repository. I'm just guessing here ^^'

> But in this case i was not really able to do any testing in the UI, because it complained about using http and not https and did not show any login form

I had the same issue here, but resolving why `make run-dev` won't work in the main repository should also be a workaround for this issue. That being said, I haven't been able to recreate the container errors you mentioned. I'm going to try a deeper dive asap

Janmtbehrens:
After googling for this error here

`exec /usr/local/bin/docker-entrypoint.sh: exec format error`

it seems to heavily indicate that there is an architecture-based problem. We've tried setting up OpenSlides on a Raspi System with ARM64 Architecture, which should be the same that mac uses, and it worked fine. 
Seeing as the problem is likely some incompatibility with your particular system architecture and OpenSlides, we can't really help debug this anymore, sorry. 
But of course we will still answer any questions you might have and would love to hear if you found a working solution @xela1601 :)

Elblinator:
Hey @xela1601, 
How is it going? Can we help in any way?

xela1601:
I closed this by accident. this PR is continued in https://github.com/OpenSlides/openslides-client/pull/6160

PR #5428, opened 2025-09-23
Add subscription fields

Elblinator:
part of #4236

meeting.constants.ts
change from default to []
=> PROJECTOR_LIST_SUBSCRIPTION and  PROJECTOR_LIST_MINIMAL_SUBSCRIPTION call this 
minial as far as I can see does not need anyfield from there,
the list is getting the info elsewhere

agenda.subscription
chagned from none to ```[`id`,  `meeting_id`]```
I also tried to include ```attachment_ids``` but I could see no difference in the behaviour when removing it

assignment.subscription
ASSIGNMENT_LIST_SUBSCRIPTION:
idField: `assignment_ids`,  fieldset: FULL_FIELDSET, 
and 
idField: `assignment_candidate_ids`, fieldset: [],

the list is also called in the detial view, hence, the FULL_FIELDSET for assignment_ids
and the info from assignment_ids` is called over other ways (assignment_id, id, meeting_id, meeting_user_id, weight)

mediafiles.subscription 
 ``` { idField: `mediafile_ids`, fieldset: FULL_FIELDSET },``` and ```follow: [{ idField: `mediafile_id`, fieldset: FULL_FIELDSET }]``` are redundant and (at least from what I could gather) only one of them is needed
But I am also not 100% sure about that
The other two Full sets are needed.

participants.subscription ```=> PARTICIPANT_DETAIL_SUBSCRIPTION``` is rarely called and, hence does not need the full set info
In the detail view of a participant the participant_list is called

design.subbscription
the fields are the ones that are needed for the design to work

### DIFF
--- a/client/src/app/domain/models/meetings/meeting.constants.ts
+++ b/client/src/app/domain/models/meetings/meeting.constants.ts
@@ -24,9 +24,9 @@ export const MEETING_MEDIAFILE_USAGE_ID_KEYS = [
 
 export type MeetingDefaultProjectorIdsKey = `default_projector_${ProjectiondefaultValue}_ids`;
 
-export const MEETING_DEFAULT_PROJECTOR_IDS_KEYS = PROJECTIONDEFAULTS.map(place => {
-    return { idField: `default_projector_${place}_ids` as MeetingDefaultProjectorIdsKey, fieldset: [] };
-});
+export const MEETING_DEFAULT_PROJECTOR_IDS_KEYS = PROJECTIONDEFAULTS.map(
+    place => `default_projector_${place}_ids`
+) as MeetingDefaultProjectorIdsKey[];
 
 export type ViewMeetingDefaultProjectorsKey = `default_projectors_${ProjectiondefaultValue}`;
 


client/src/app/domain/models/meetings/meeting.constants.ts
```diff
@@ -24,9 +24,9 @@ export const MEETING_MEDIAFILE_USAGE_ID_KEYS = [
 
 export type MeetingDefaultProjectorIdsKey = `default_projector_${ProjectiondefaultValue}_ids`;
 
-export const MEETING_DEFAULT_PROJECTOR_IDS_KEYS = PROJECTIONDEFAULTS.map(
-    place => `default_projector_${place}_ids`
-) as MeetingDefaultProjectorIdsKey[];
+export const MEETING_DEFAULT_PROJECTOR_IDS_KEYS = PROJECTIONDEFAULTS.map(place => {
+    return { idField: `default_projector_${place}_ids` as MeetingDefaultProjectorIdsKey, fieldset: [] };
+});
```
bastianjoel, line 29: Change this back to as it was and instead use `MEETING_DEFAULT_PROJECTOR_IDS_KEYS` in `fieldset` or `additionalFields` in subscriptions.

bastianjoel (changes requested):

PR #5467, opened 2025-10-07
Change display of analog polls

### LINKED ISSUE #5451: Fix analog poll in detail view
**Describe the bug**
In the detail view from the poll and from the belonging content object  "Invalid votes" and "Total votes cast" needs to be smaller 

**How to Reproduce**
1. create an analog motion poll/election poll with invalid votes and a number with "Total votes cast"
2. look at the poll in the detail view

Image

 
**Expected behaviour**
The strings (and their numbers) should be as small as the rest from the text

The strings in the best case scenario stay big if you look at the projected result

reiterl:
Resolve #5451


emanuelschuetze (approved):
tested

Elblinator (approved):

PR #5470, opened 2025-10-07
Agenda Export: problems in single agenda item export 

### LINKED ISSUE #5464: Agenda Export: problem in single agenda item export
**Describe the bug**
If a single agenda item is exported without a table of content, an error may occur in a specific sequence whereby the table of content is also exported. The error occurs if multiple export with table of contents was used prior to the single export. This setting is retained in the background and executed in the single export, even though it was not selected or is not selectable at all.  

**How to Reproduce**
1. open an agenda
2. create two agenda items
3. open the agenda list > export the list with the table of content
4. open an agenda item and do a single export. --> Table of content will be exported, even if it's not selected (selectable)

**Expected behavior**
- [ ] Bug should not occur. Table of content should not be automatically exported in single export

**Additional behavior**
- [ ] Deleted the category "Seitenlayout" from the single agenda item export options. They aren't relevant.

Image

Elblinator:
resolves #5464


MSoeb (approved):
works as intended.

PR #5496, opened 2025-10-15
Use tree-weight to sort motions for call list in exports

### LINKED ISSUE #5461: Fix call list export with child elements
**Describe the bug**
If a call list has child elements, the export (in CSV and PDF) places the child elements in the wrong order. 

**How to Reproduce**
1. open the call list: meeting > motion > menu > call list
2. define some child elements (not to the first/second motion)
3. export the call list --> the order in the export is different to the one in Openslides

call list in OS:

Image

export:

Image

**Expected behavior**
Child elements should be listed under called with directly after their parent item in the call list.

reiterl:
Resolve #5461 

Use a different weight to sort the call list. A comment in the motion said, the 'tree-weight' is the
sorting weight for call lists. Use this. Tests show a slightly better behavior in the exports.


bspekker (approved):
function

Issue #5446, opened 2025-09-29
Change user icon and add feedback for present state


bspekker:
**Current behavior**
The present state in the user icons in the top right is very small and hard to see for participants. Additionally, if the users can't set the present state themselves, it is difficult for them to see if they were set to present by their admins. Therefore the icons should be changed and a feedback incorporated for the present state.

**Wanted behavior**

- [ ] The icons in the top right corner should be changed to "checkmark" if the user is present and to "person" if the user is not present (and while navigating the dashboard)
- [ ] If the setting "Allow users to set themselves as present" is deactivated, the user menu should show if the user is present (no info necessary if they are not present). It is read-only. See mock-up for placement and icons.

The user is present but can't set themselves to not present 
<img width="328" height="144" alt="Image" src="https://github.com/user-attachments/assets/2de9f3e1-827f-40f4-8742-bb8f53281391" />

bspekker:
We discussed that the icons could be changed and the checkmark could be added to the person icon if user is present.
Please make a suggestion.

PR #5498, opened 2025-10-16
Update account-button, add small green badge if present

reiterl:
Resolves part of #5446 

As showcase for how the badge solution could look. @bspekker please have  a look at it.


reiterl:
Ah, okay, good to know. @Elblinator also wants to create a version of the account button.

PR #5501, opened 2025-10-20
Small changes to the account button

### LINKED ISSUE #5446: Change user icon and add feedback for present state

reiterl:
Resolve #5446 

- icons to left of the username
- other icon 'done'
- change display if not allowed to change present

### DIFF
--- a/client/src/app/domain/models/meetings/meeting.ts
+++ b/client/src/app/domain/models/meetings/meeting.ts
@@ -141,6 +141,8 @@ export class Settings {
     public motions_export_follow_recommendation!: boolean;
     public motions_hide_metadata_background: boolean;
     public motions_create_enable_additional_submitter_text: boolean;
+    public motions_enable_restricted_editor_for_manager: boolean;
+    public motions_enable_restricted_editor_for_non_manager: boolean;
 
     public motion_poll_ballot_paper_selection!: BallotPaperSelection;
     public motion_poll_ballot_paper_number!: number;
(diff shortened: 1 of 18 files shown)


client/src/app/site/modules/global-headbar/components/account-button/account-button.component.html
```diff
@@ -8,9 +17,6 @@
             }
         </div>
     </div>
-    <div class="fake-fab-button flex-center">
-        <mat-icon class="icon-16" color="primary">{{ isPresent ? 'how_to_reg' : 'person' }}</mat-icon>
-    </div>
```
Elblinator: This should still be there but not with color="primary" and only the person icon

client/src/app/site/modules/global-headbar/components/account-button/account-button.component.scss
```diff
@@ -53,6 +53,9 @@ a {
     line-height: $button-size;
     text-align: center;
     cursor: pointer;
+}
+
+.fake-fab-button-border {
```
Elblinator: no new class needed but ```background: white;``` should be removed
also for the person icon

Elblinator (changes requested):

bspekker (approved):

PR #5558, opened 2025-11-06
Add orga setting to disable forward with attachments

### LINKED ISSUE #5540: Add restriction for forwarding of motion attachments
**Current behavior**
If a motion has an attachment and it is forwarded, it is currently always possible for the meeting administrators to forward attachments to other meetings they have no permissions for. In some cases this is not desired and should be restricted on the orga level.

**Wanted behavior**

- [ ] A new setting is needed on the orga level: "Attachments to motions cannot be forwarded"  (end of "General"), by default it is not enabled
- [ ] If the setting is enabled, attachments will never be forwarded and the motion forward dialog in a meeting will never offer the option to forward attachments.

reiterl:
Resolve #5540

Needs:
https://github.com/OpenSlides/openslides-backend/pull/3192
https://github.com/OpenSlides/openslides-meta/pull/342


Elblinator (approved):
every thing seems to be in order

Elblinator:
@reiterl please merge this yourself bc backend (and meta) are necessary for the client

PR #5604, opened 2025-11-17
Fixed line breaks in submitters #5343

### LINKED ISSUE #5343: Line breaks in submitters needs white spaces to work
**Describe the bug**
Line breaking in submitters only works with spaces.
Image
Both regular submitters and text submitters have this behaviour.

**How to Reproduce**
1. Create a motion
2. Add a text submitter with a really long name without spaces

Alternatively:
1. Create a user with a really long name that has no spaces (f.E. by giving an extremely long first name)
2. Create a motion using said user as submitter

**Expected behavior**
There should be forced line breaks at the point of cutoff.

**Additional context**
We already have automatic linebreaks if there is spaces. This issue is probably a matter of a single line of css.
It may be worthwhile to also check other places for the same type of bug

Borreg0:
Closes #5343

### DIFF
--- a/client/src/app/site/pages/meetings/modules/projector/modules/slides/components/motions/modules/motion-slide/components/motion-slide/motion-slide.component.scss
+++ b/client/src/app/site/pages/meetings/modules/projector/modules/slides/components/motions/modules/motion-slide/components/motion-slide/motion-slide.component.scss
@@ -31,7 +31,7 @@
 .metatable {
     width: 100%;
     font-size: 120%;
-
+    word-break: break-all;
     td.min {
         width: 1%;
         white-space: nowrap;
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
code

MSoeb (changes requested):
In general the PR breaks long words and doesn't let them flow under the motion text card. That's fine. 

But one thing needs to be changed. 

In the current live version, line breaks in names do not occur within a word. The name is not broken within the word. The word breaks to the next line. (See example) The line break is therefore before the word that can no longer be displayed completely in the current line. -> This behavior should persist.

Example of live version - intended behavior:
image

PR Version of line break  wrong behavior:
image

Goal:
In cases where a word can be moved completely to the next line, the word should appear in its entirety on the next line and should not be broken. The line break should follow the rule of the example of the current live version.

Elblinator (changes requested):
meta is not supposed to be updated like that

PR #5610, opened 2025-11-18
Added aria label and focus on cancel button

### LINKED ISSUE #5602: Screenreader add global search and cancel for global search
**Current behavior**
The search icon from the global search is missing the ARIA-label and the global search dialog is not focusing the cancel button (and hence not reading it out loud)

**Wanted behavior**
- The global search icon should have an ARIA-label
- The cancel button from the global search should be focusable

**Additional:**
Focusable mean that if you use tap to navigate through a page the tap will at some point reach that place.

The global search:
Image

The cancel button is the X in the corner
(Before taking the screenshot I hovered over the icon, that's why it seems focused, the shadow indicates focus a lot of the times)
Image

Borreg0:
Closes #5602

### DIFF
--- a/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.html
+++ b/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.html
@@ -6,7 +6,7 @@
     <div class="flex-center">
         <!-- Button to open the global search -->
         @if (isSearchEnabled) {
-            <button mat-icon-button (click)="openSearch()">
+            <button mat-icon-button role="Search" (click)="openSearch()">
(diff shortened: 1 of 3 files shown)


bspekker (changes requested):
Two minor things:
1. the cancel buttom that can be in focus now should also have an aria label
2. the search bar should still be the first focus in the dialogue and not the cancel buttom

bspekker (changes requested):
Currently after closing the global search, the focus jumps to the beginning of the page. This is irritating.

The focus after closing the search menu should be on the search icon.

bspekker (approved):

Elblinator (approved):

PR #5613, opened 2025-11-19
Mimicked styles to avoid interferences

### LINKED ISSUE #5599: Motion-list change how motion title are read by Screenreader
**Current behavior**
The motion title in  the motion list are embedded in a div and therefore not read as importance elements

Image

**Wanted behavior**
Without changing the appearance the motion titles should be a h1

**Additional:**
I would recommmend changing the div to a h1 and then mimicking the current style.
Important: The elipsis for long titles should still work

Borreg0:
Closes #5599

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.html
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.html
@@ -114,7 +114,7 @@
                 <os-detail-view-navigator [model]="motion"></os-detail-view-navigator>
             }
             <div class="column-title innerTable overflow-hidden">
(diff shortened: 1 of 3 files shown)


Elblinator (changes requested):
- The title should be directly in the h1 (h1 > title)
- if possible change the span from the motion number to ng-content or something similar which is removed if rendered
- add an aria-labelled by to "span class="favorite-star"" which describes that this symbolizes the favorite
- add an aria-labeled-by for the h1 with a function which returns a string " 'is favorite' + motion number + motion title"
If there is no number that part should be removed
If the motion is not a favourite then that part should be removed
if you need help with how to do the code look at: https://github.com/OpenSlides/openslides-client/pull/5576/files

bspekker (changes requested):
The title is too strong:
image

It should be visually the same as the status quo:
image

Elblinator (changes requested):

bspekker (approved):

Elblinator (approved):

PR #5651, opened 2025-12-01
Long submitters names now break

### LINKED ISSUE #5343: Line breaks in submitters needs white spaces to work

Borreg0:
closes #5343


Elblinator (approved):
code

PR #5661, opened 2025-12-05
Colors changed

### LINKED ISSUE #5521: Fix text colour in global header in dark mode
**Describe the bug**
When the global header colour is changed, in the light mode the text colour will switch to a light or dark colour depending on the brightness of the colour. When switching to the dark mode, the text in the global header is always white/light.

**How to Reproduce**
1. set OpenSlides to light mode and change the global header colour to a bright or dark colour and observe the text in the header: the text colour changes
2. set OpenSlides to dark mode and change the global header colour and observe again: the text colour doesn't change 

(3. for comparison repeat both steps and change the primary colour.)

**Expected behavior**
The behaviour should be the same as for a colour change in the primary colour: if the header is light, text should be dark and if the header is dark, the text should be light.

Borreg0:
Resolves #5521

I found another issue while fixing this one:

As the button "save" shares the same colour as primary, if the theme is set to light and the primary color is also white, the button becomes invisible. I propose to change the color to "accent"
Screenshot_20251205_115332

Change with accent (not implemented).
Screenshot_20251205_115701

### DIFF
--- a/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.scss
+++ b/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.scss
@@ -11,6 +11,11 @@
     display: flex;
     flex-direction: row;
     padding-left: 26px;
+    .mat-icon,
+    os-account-button,
+    .display-name {
+        color: var(--theme-primary-contrast-500);
+    }
 }
 
 .background-headbar {


bspekker (approved):
Looks good!

Agree, that the button should have the accent colour, especially since it is like that in every other dialogue. I will create a new issue for it.

Elblinator (approved):

PR #5681, opened 2025-12-12
motion forward: filter target meetings with end_time

### LINKED ISSUE #5148: Motion forward > target meeting list: Visibility of meetings
**Current behavior**
The motion forward dialog currently displays all possible meeting, in which a motion can be forwarded plus past meetings. It makes no sense to display past meetings, because it's not a practical usecase to forward a motion to a meeting in the past. Additionally, if all past meetings would be displayed, the list can be very long, if past meetings are not archived. 

**Wanted behavior**
The target meeting list should not display past meetings (meeting dates in the past) or dateless meetings.
The list should only show timeless, current and future meeting.

reiterl:
Resolve #5148

Add a filter method to the target meetings.
Exclude meetings with current time  > end_of_day(end_time ).

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/components/motion-forward-dialog/services/motion-forward-dialog.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/components/motion-forward-dialog/services/motion-forward-dialog.service.ts
@@ -186,7 +186,7 @@ export class MotionForwardDialogService extends BaseDialogService<
 
     private endAndActiveMeetingFilter(meetingData: GetForwardingMeetingsPresenterMeeting): boolean {
         return (
-            Number(meetingData.id) !== this.activeMeeting.meetingId &&
+            +meetingData.id !== this.activeMeeting.meetingId &&
             meetingData.end_time &&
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/pages/motions/components/motion-forward-dialog/services/motion-forward-dialog.service.ts
```diff
@@ -174,6 +183,16 @@ export class MotionForwardDialogService extends BaseDialogService<
         }
     }
 
+    private endFilter(meetingData: GetForwardingMeetingsPresenterMeeting): boolean {
+        const referenceTime = meetingData.start_time ?? meetingData.end_time;
```
Elblinator: I do not know any compination where a meeting has only an end_time or only a start_time. 
Why do you not just use "meetingData.end_time?

reiterl: I use meetingData.end_time now only.

Elblinator (approved):
code

PR #5699, opened 2025-12-30
Add 1sec delay to main tooltip

### LINKED ISSUE #5670: Give tooltips in `os-global-headbar` a 1sec delay
**Current behavior**
The tooltips in the  `os-global-headbar` are appearing immediatly if the cursor is moving over them

**Wanted behavior**
The tooltips should appear only after the cursor rests for 1 second over the icon

**Additional context**
example of one tooltip ion the headbar
Image

Elblinator:
resolves #5670


bspekker (changes requested):
Works so far, except for tooltip for tile view in motions and new folder in files.

bastianjoel (approved):

PR #5701, opened 2026-01-02
Adjust motion-list

### LINKED ISSUE #5655: Motion-list adjustment
**Current behavior**
In the desktop view the motion titles are just as long as you can see in the screenshot and they are overlaps by the meta info box even if the box is empty  ^1 and ^2

**Wanted behavior**
- If a meeting has no category, no motion blocks and no tags then the mask to select them should be hidden (categories/blocks/tags do not need to be assigned, they just need to exist)  ^1
- If a person cannot change these meta information and they click on that area then they should be navigated to the motion
- If a person cannot change the meta information and cannot see the LoS then the whole rest of the line should navigate to the motion-detail

^1
Image

^2
Image

Elblinator:
resolves #5655 

IMPORTANT:
If a user cannot see the LoS ALL LoS icons are now hidden

### DIFF
--- a/client/src/app/site/pages/meetings/modules/meetings-component-collector/projectable-list/components/projectable-list/projectable-list.component.html
+++ b/client/src/app/site/pages/meetings/modules/meetings-component-collector/projectable-list/components/projectable-list/projectable-list.component.html
@@ -41,16 +41,21 @@
         </div>
     </div>
 
-    <!-- Speaker -->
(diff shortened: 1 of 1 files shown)


luisa-beerboom (approved):
Generally looks good. Just one small point that I am unsure on. If my thoughts are unjustified, consider the code approved.

PR #5739, opened 2026-01-13
All label for global search

### LINKED ISSUE #5727: Add aria label for global search
**Wanted behaviour:** 
if the global search is focused with tab something like "Global Search" should be told to the user

Elblinator:
resolves #5727 

Also removed the span, because that makes it harder for the screenreader to reach the content and I did not see any visual difference with and without span


reiterl (approved):
LGTM

PR #5740, opened 2026-01-13
Fix Globalhead bar color

### LINKED ISSUE #5725: Global headbar items invisible in dark mode
**Describe the bug**
As described in #5521, the text color will change to white when activating dark mode, and if the global headbar color is also set to white the global headbar items will be invisible. The expected color calculations will not work in this case. 

light mode
Image

dark mode
Image

**How to Reproduce**
1. Change the global header color to white
2. Switch to dark mode
3. Organisation name, account name, and search button are now invisible

**Expected behavior**
Make items visible when in dark mode regardless the background color

Elblinator:
resolves #5725


reiterl (approved):
As `#000000`  is `black` it is more correct from the code side.
I'm not sure why it should work.
Code looks good to me.

PR #5773, opened 2026-01-23
Categories and Motion Block names will break now

### LINKED ISSUE #5745: Categories and motion blocks need line breaks
**Describe the bug**
Similar to what happened in #5343 , very long categories or motion block titles don't break and go behind the motion mat card:

Image

**How to Reproduce**
1. create a motion
2. create a category and a motion block with a very long title
3. select this category and motion block to your motion from step 1

**Expected behavior**
The category and motion block labels have a line break so the don't intercept with the motion card.

**Additional context**
Please also check that titles without space also wrap at some point.

Borreg0:
Resolves #5745


bastianjoel (dismissed):

Elblinator (approved):
Code

MSoeb (approved):

PR #5817, opened 2026-02-03
Change participants and accounts example file data

### LINKED ISSUE #3822: Import Pages - csv example file improvement

reiterl:
Resolve #3822

### DIFF
--- a/client/src/app/site/pages/meetings/pages/participants/export/participants-export-example.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/export/participants-export-example.ts
@@ -1,51 +1,15 @@
 export const participantsExportExample: any = [
     {
-        title: `Dr.`,
         first_name: `Max`,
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
code

Elblinator (approved):
function

PR #5819, opened 2026-02-03
Change follow recommendation dialog of motion block

### LINKED ISSUE #2943: Motion-block: Change dialog-text for "Follow recommendation for all motions"-action
**Current behaviour:**
The dialog (if you press on the icon) has a wrong wording:
008 B

**Reproduction:**
in a meeting:
1. create motion block
2. activate a recommender
3. create a motion 
4. select a motion block and a recommendation (e.g. "Annahme") for the motion
5. got into the motion block 
6. press in the top right area on the double tick 
7. the dialog should open

**Wanted behaviour:**
The wording should be changed to: "Are you sure you want to overwrite the state of all motions with recommendations? 
Motions which have a final state will not be changed."

<- I would split it in two paragraphs.

reiterl:
Resolves #2943 

I have changed the two labels.


Elblinator (approved):
function

PR #5833, opened 2026-02-06
Use ViewSpeaker.topic for sort numbering in contribution view

### LINKED ISSUE #3618: Contributions: Sort order Numbering (improvement needed)
**Description:**
The 'Numbering' sorting in contributions needs to be adjusted. The elements in the list are not sorted equally. The content is first sorted by module and then within the module. This is not optimal.

**What should happen:**
Sorting ' Numbering' should sort across all modules and structure the elements according to their title without creating module clusters.

Additionally: 
If the user isn't allowed to see an item from the contribution-list, then this item should be completely hidden (see comment for more context)

reiterl:
Resolve #3618 

Another try to fix this issue.
This uses ViewSpeaker,topic for the numbering sort. The topic attribute is the string displayed in blue in the
contribution list.


Elblinator (approved):
beautiful \o/
(code and function)

Issue #5551, opened 2025-11-03
Linter warnings


Elblinator:
contitunation of: https://github.com/OpenSlides/openslides-client/issues/4885

- [ ] @typescript-eslint/no-empty-object-type
- [ ] no-async-promise-executor
- [ ] max-len 160/120
- [ ] no-case-declaration

PR #5842, opened 2026-02-09
Clean-up Linter warnings

Elblinator:
part of #5551


Elblinator:
Also remove getNumber() and swapped the two cases where it's used with getIdentifier()

PR #5861, opened 2026-02-10
Change assignment results in screen reader

### LINKED ISSUE #5598: Screenreader change how assignment results are read out loud
**Current behavior**

Image

How the table is spoken out loud (not verbatim but close):
'third cell yes, fourth cell no, fifth cell Abstain. one Klara Ziegel rot 42 percent 6 21percent 3 28 percent 4 2 Tim Ziegel....

**Wanted behavior**
One line should be spoken as one unit 
"Poll Options [the options]. Read out loud in the order they placed"
"Name, structure level: the structure level. Yes: 6 votes 42%. No:  3 votes 21%. Abstain: 4 votes 28%"
"Name, ...."
Please use the actual percentages not the abbreviations.

This Issue will be probably need some more loops if worked on

**Reproduction:**
1. activate vote delegations in the setting (participant-detail -:> delegate setting)
2. set yourself to present
3. delegate the vote from the opther participant to yourself (this make the next steps easier)
4. create an election with two or more candidates
5. create a poll where you and the other participants can partace in (I would recommend a Y/N/A poll but the others should work too)
6. cast your votes
7. end and publish or just end the poll
8. now the results should be read as described

Elblinator:
resolves #5598

### DIFF
--- a/client/src/app/site/pages/meetings/pages/assignments/modules/assignment-poll/components/assignment-poll-detail-content/assignment-poll-detail-content.component.html
+++ b/client/src/app/site/pages/meetings/pages/assignments/modules/assignment-poll/components/assignment-poll-detail-content/assignment-poll-detail-content.component.html
@@ -5,20 +5,20 @@
                 <div class="row" role="row" tabindex="-1" [class]="classOptionAmount">
                     <div
                         #btn
-                        aria-label="{{ ariaLabel('place') }}"
                         class="cell"
                         osArrowNavigation
                         role="columnheader"
                         tabindex="0"
+                        [attr.aria-label]="ariaLabel('place')"
                         [buttonList]="buttonElements"
                     ></div>
                     <div
                         #btn
-                        aria-label="{{ ariaLabel('name') }}"
                         class="cell"
                         osArrowNavigation
                         role="columnheader"
                         tabindex="-1"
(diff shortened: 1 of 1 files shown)


client/src/app/site/pages/meetings/pages/assignments/modules/assignment-poll/components/assignment-poll-detail-content/assignment-poll-detail-content.component.html
```diff
@@ -1,127 +1,134 @@
...
-                                            {{ 'Votes' | translate }}
-                                        </span>
-                                    }
-                                </th>
+        @if (hasResults && canSeeResults && !shouldShowChart && (hasLoaded | async)) {
+            <div aria-label="Results-table" class="assignment-results" role="grid" tabindex="-1">
+                <div class="row" role="row" tabindex="-1" [class]="classOptionAmount">
+                    <div
+                        #btn
+                        aria-label="{{ ariaLabel('place') }}"
```
bastianjoel: It would be better to use proper attributes for binding data. https://angular.dev/guide/templates/binding

For aria you need to use attr prefix like this: `[attr.aria-label]="..."`

bastianjoel (approved):

PR #5862, opened 2026-02-10
Redirect user after login to last page

### LINKED ISSUE #5763: URL: Deepl-Link redirection
**Current behavior**
If a user gets a link of a motion, it is not possible to open it directly after login. The first page that opens is the dashboard or a meeting starting page. 

**Wanted behavior**
It should be possible to redirect the user after login to the URL the user wants to open.

Example: 
https://demo.openslides.org/18/motions/9 <- user gets this link and click on it. The login page now opens. Users log in. If he has the permission to see the content of the link, a redirection should happen to the content. 

Same should happen for for guest users if public meetings setting is enabled.

bastianjoel:
resolves #5763

Currently untested with saml login.


Elblinator (approved):
code

MSoeb (changes requested):
Deep link for account admin doesn't work. The account admin always sees a calendar with no meetings. 

Additionally I haven't tried it with a SAML config. This will follow.

Elblinator:
> Deep link for account admin doesn't work. The account admin always sees a calendar with no meetings.

@MSoeb The deep-link works 
But the account admin probably has no meeting, and is not allowed to see/be in the meeting

MSoeb (changes requested):
Deep link redirection with SSO Accounts doesn't work. Tried it with orgaadmin and normale delegate. In both cases the redirection doesn't work.

Tested with deepl link to motion:
Normal delegate will open the meeting, not the motion. 
Orgaadmin will open the dashboard.

What should happen: Redirections should work as intended.

bastianjoel:
@MSoeb I would suggest merging this now as I don't see that we will ever fix this for SSO due to lack of dev setup. 

We should look into this again after the keycloak integration.

MSoeb (commented):
Talked with @bastianjoel about the PR and how we handle SSO logins. 

We'Ve made the decision, that SSO redirect will be solved in a seperate issue.

PR itself works fine.

PR #5865, opened 2026-02-11
Fix fuzzy match participants and accounts

### LINKED ISSUE #2601: Enhancement: Fuzzy match for Participants search
Very small but helpful enhancement for my organization would be a fuzzy match for accented characters like `é` to show up when searching for `e`. US English keyboards do not have easily accessible accent marks like European keyboards and require a few confusing workarounds to pull up these characters. 

Names like `Renée` don't come up with the current exact match which is typed out when pulling up `Renee`, making it appear a participant is not present, or is a cause of confusion between tech support and participants when in-person where we can't copy and paste. Seating a backup delegate mid-debate with a similar accented name ended up taking a minute longer than was comfortable as it was as a vote was about to launch.

Borreg0:
Resolves #2601


bastianjoel (changes requested):
Besides from the comments I think your changes are quite far away from fitting the requirements. 

As far as I can see you also only do the normalization just one sided which means that when providing an input with special chars you won't be able to find such entries.

bastianjoel:
I do not see you working on this for quite some time. Also this is far from an acceptable solution. Therefore I am closing.

Issue #5856, opened 2026-02-10
Diff issue in amendments with ordered/unordered lists


MSoeb:
**Describe the bug**
Amendments in an ordered/unordered list result in an incorrect diff. If new content is entered at a lower level, the diff generates a new line. This is incorrect. 

Amendment text:
<img width="348" height="328" alt="Image" src="https://github.com/user-attachments/assets/55b3e10d-4b4e-41cb-9c69-20375ad02476" />

Amendment output with wrong diff:
<img width="458" height="435" alt="Image" src="https://github.com/user-attachments/assets/2fb8362b-90f0-4659-b358-0688b469121b" />
 

**How to Reproduce**
1. open motions
2. create a new motion with code 1 linked below.
3. create an amendment to it in line 1-5 with Code 2 linked below.
4. --> Error with wrong new line will appear.

Code 1:
> <ol>
	<li>Testzeile 1<ol>
	<li>Testzeile 2<ol>
	<li>Testzeile 3</li>
	<li>Testzeile 4</li>
</ol>
</li>
<li>Testzeile 5</li>
</ol>
</li>
<li>Testzeile 6</li>
</ol>

Code 2:
> <ol>
	<li>Testzeile 1<ol>
	<li>Testzeile 2<ol>
	<li>Testzeile 3 Mehr Inhalt</li>
	<li>Testzeile 4</li>
</ol>
</li>
<li>Testzeile 5</li>
</ol>
</li>
</ol>

**Expected behavior**
If an amendment if done to a lower level list point, the diff should work correctly.

PR #5871, opened 2026-02-12
Motion diff issue with nested list and text change

reiterl:
#5856 

Add an unittest to show the issue in the Diff service with nested list. 
The unittest is at the moment red.


packages/openslides-motion-diff/src/diff/index.spec.ts
```diff
@@ -822,6 +822,29 @@ describe(`MotionDiffService`, () => {
             expect(diff).toBe(expected);
         });
 
+        it(`handles inserted text within nested lists`, () => {
```
Elblinator:
```suggestion
        it.skip(`handles inserted text within nested lists`, () => {
```

Elblinator (approved):

Issue #5863, opened 2026-02-11
motion diff: amendment and nested list (dot in text)


reiterl:
**Describe the bug**
The motion diff of a motion with a nested list and a dot and an amendment with a change to second level of the
nested list is an all deleted (red) and everything new added (green).
I also see an warn message in the console:
```
Added sequential number mapping view without meeting id
```
 and the amendment hasn't got a title.

**How to Reproduce**
1. Go the motion list in a meeting
2. Create a new motion with following text: `<ul><li>Ebene 1<ul><li>Ebene 2.1</li><li>Ebene 2.2</li></ul></li></ul>`
3. Create an amendment to this motion with  `<ul><li>Ebene 1<ul><li>Ebene 2.1 foo</li><li>Ebene 2.2</li></ul></li></ul>`. The `foo` is the only change
4. Look at the motion detail view 

**Expected behavior**
It should work like:
motion text:  `<ul><li>Ebene 1<ul><li>Ebene 2A</li><li>Ebene 2B</li></ul></li></ul>`
amendment text: `<ul><li>Ebene 1<ul><li>Ebene 2A foo</li><li>Ebene 2B</li></ul></li></ul>`
So it should only show the `foo` in green (as change).

**Additional context**
I think it is a problem with the HTML Diff stuff

reiterl:
Also see #5856 for another problem in this context.

Elblinator:
The end of this bug is that the diffDetectBrokenDiffHtml(...) is detecting broken html that's why the complete original text is deleted and the new amenment text is inserted. The program simply doesn't know what the correct union of that is.

From what I could see is that the html is wrongly calculated

MSoeb:
I'e reopened the issue, because the problem still occurs and is not solved.

PR #5872, opened 2026-02-12
Add a test to the diff service about nested lists and dots

reiterl:
Adds a test for #5863

This test is failing, because the problem isn't fixed.


packages/openslides-motion-diff/src/diff/index.spec.ts
```diff
@@ -822,6 +822,25 @@ describe(`MotionDiffService`, () => {
             expect(diff).toBe(expected);
         });
 
+        it(`handles changed text within nested lists (part 2)`, () => {
```
Elblinator:
```suggestion
        it.skip(`handles changed text within nested lists (part 2)`, () => {
```

Elblinator (approved):

Issue #5855, opened 2026-02-10
Prettify mandate check part 2


Elblinator:
**Describe the bug**
continuation of https://github.com/OpenSlides/openslides-client/issues/5785

I talked with @MSoeb about the Issue and serveral things we missing in the issue description

The mandate check is not clearly structured, some adjustments should be made

**Expected behavior**

generic wanted behaviours:
- [ ] Even in small views there should not be a horizontal scrollbar, the content should be made to fit/wrap
- [ ] The progressbar should be bigger (see the projected live votes for a n example of a bigger progressbar)
- [ ] if there is enough space the progress bar should use more space (see mock-up)

Gender list:

- [ ] The entries should have more height, see screen shot 
- [ ] The different entried do not need to have alternting colours, but there should be a very light line deparating them (see mat-divider in the participants three dot menu for an example)
- [ ] If there is enough space the from the genders should be further aprat (see mock-up)

Participant list:

- [ ]  in the desktop view the  present/ Absent title should be directly above the participants list. 
The list of participants names should be more left not the other way around

**Additional context**
<img width="1920" height="993" alt="Image" src="https://github.com/user-attachments/assets/7bcbd4d3-73e2-4d8b-8559-13bc97f4526a" />

Elblinator:
everythign is done

PR #5875, opened 2026-02-13
Fix Progress bar width

Borreg0:
Resolves a part of #5855


MSoeb (changes requested):
We need to make a few changes: 

1. When languages are switched, the size of the bar always changes and the distance to the text on the left also changes.

To fix this problem, the bar should be narrower overall so that it remains in the same position even when languages are switched.

Use the Russian translation as a reference, as it has the longest text length.

2. Small screens: The content should still be visible on small devices even with the bar. The mat cards and the content should fit in a format, which fits the screen space. 

Example of unreadable content.
image

3. Alignment of gender numbers: The alignment of the sub gender numbers should be in line with the overall value.

See: The numbers left of the red line should be in line aswell.
Ausrichtung

MSoeb (changes requested):
Some additional changes are required.

**1. Matcard on small screens like smartphones:** The matcard should not be scrollable **horizontally** on small devices. The matcard should fit the screen size. I've tested it with a Pixel 8 and on a iPhone SE. I can slightly be moved. Here a video example of it.

https://github.com/user-attachments/assets/c2873f92-b0bf-4aef-be35-cbabfd186a17

**2. Total value and numbers of genders under it:** The values should be aligned like in the mockup. 

Affected area:
Image

**3. Problem with the text to the left and right of the progress bar:**
There are several issues occurring simultaneously here. First, switching languages causes the progress bar to shift. This is clearly visible when you change the language from German to Russian. Second, the spacing between the progress bar and the text on the right is too small, which doesn’t look good visually. 

To solve both problems, I would suggest making the progress bar slightly shorter overall. This would create more space on the left and right side, which, in my opinion, would resolve the issue.

Borreg0:
I am not sure if the class bar-height is useful. Please take a look.

MSoeb:
I discussed the pull request and the requirements with @emanuelschuetze and @Eich-hoernchen (Martin) last week. During our conversation, we identified a few points that need to be changed and will likely require some testing with immediate feedback. To streamline the process, we’ve decided that Martin will take over the rest of the work. He’ll be able to work directly with you on-site to make the necessary adjustments. 

Martin will get in touch with you.

Elblinator:
Please reassign me when the coding and styling is mostly done

PR #5931, opened 2026-03-02
Add meeting id to subscription

### LINKED ISSUE #5929: Prevent warning in console
**Describe the bug**
Since Rel-DB there are several warnings in the console which were not there before: 
```
Added sequential number mapping view without meeting id Proxy(Qo) {_model: t, viewModelUpdateTimestamp: 1772461695383, getTitle: ƒ, getListTitle: ƒ, getVerboseName: ƒ, …} poll
```

**How to Reproduce**
1. in a meeting create any poll 
2. the warning appears once for every poll in the meeting

**Expected behavior**
The warning should not be thrown

Elblinator:
resolves #5929


bastianjoel (changes requested):
Probably there is a subscription with a missing subscription to `meeting_id`. 

This is not a valid fix.

PR #5948, opened 2026-03-05
Add embedding option in text editor

### LINKED ISSUE #5902: Embedding of motion/assignment/topic links in home/topic text
It should be possible to link to other topics, motions or assignments in topic and home texts.

The editor should get a new button or the add link popup should be extended where a user can choose from a list. 
An embedded link should look like in the ABK field in motions (e.g. [motion/4]). 

For this to work in the texts the texts need to be processed to parse those tags and request the titles of these entries via single requests on the autoupdate.

Borreg0:
Closes #5902


Elblinator (changes requested):
- The new feature should still be in the dialog 
- Both options should have a keyboard_arrow_down/keyboard_arrow_up icon (and the function of such)
- Both titles should have the same colour and NOT the disabled colour
- "Insert/edit link" should be changed to  "Insert/edit external link"
- The option to add Home is missing right now
- The option to add more than one object is not working  
  - If you look at a motion with this functionality you can add a string or several motions. 
  - if you use this function:
    - then you need to change the title to "Link a topic/motion/assignment"
    - AND after you added one and then press the option again you need to add a NEW link. Right now after adding one you are visually one space behind the link, and if the option is pressed again the old link is change to the new link. Only if you are on the link (cursor directly on the last/first letter or in between) the link should be changed.
 - Remove the done/clear option and if an object was selected then update the string in the selected item box directly
 - The issue with the updated link if you visually are at a different place should be fixed regardless of whether or not the dialog can create one vs several links

Elblinator (changes requested):
A) The links form a topic are currently not working corretly. 
Please remove any non topic from the topic selection (e.g. a motion if it's on the topic-list)

B) give the element with the class "reference" the function to open/close the menu (in cotrast to just the keyboard_arrow_up/_down

C) Disable this function for the motion editor (ANY editor which is in the motion/amendment view

D) If you navigate into the dialog with only key inputs (without the mouse) you focus a wrong element in the beginning.
After implementing B) that should again have changed, The wanted behaviour is:
 1. after opening the dialog
If you have both option (see C) )
 2. Then your focus should be on The first title/arrow and if enter is pressed then the option should open
 3. If you press tab then you should navigate to other title/arrow and if enter is pressed then the option should open etc.
 4. The next tab should navgigate to the buttons

If you have just one option (see C) )
2. The focus should be immediately on the url input field
3. Tab goes to the "Text to display" input etc.

E) If you inserted an object via your new function and decide that you want to use the other url function, then you have to close and open the dialog again and then you can insert a url. Please make a suggestion on how to prevent the needed closing of the dialog

client/src/app/ui/modules/editor/components/editor-link-dialog/editor-link-dialog.component.ts
```diff
@@ -45,14 +184,91 @@ export class EditorLinkDialogComponent {
     }
 
     public save(): void {
-        if (this.link.href && !/^[a-zA-Z]+:\/\//.test(this.link.href)) {
-            this.link.href = `http://` + this.link.href;
+        if (this.externalUrl) {
+            if (!/^[a-zA-Z]+:\/\//.test(this.link.href)) {
+                this.link.href = this.externalUrl.includes(`http`) ? this.externalUrl : `http://` + this.externalUrl;
```
bastianjoel: This is to add a protocol if missing. Your change here makes no sense. 
foo.com -> http://foo.com
http.com -> http.com

bastianjoel (changes requested):
I've stumbled over too much things I don't like too quickly here. Therefore stopping my review. 

Please take a look at your code again and cleanup unnecessary code, debug statements and check for obvious bugs before submitting it for review.

Elblinator (changes requested):
- Please also add the clear option for the Insert/edit external link
   - this clear option should also just be present if  "Link a topic/motion/assigment" is present (aka hide this optioon if it's not present)
- Hide the "keyboard_arrow_down" icon when only "Insert/edit external link" is present
-  "Link a topic/motion/assigment"  should only be present in home and the topic creating/editing 
   -  I saw it in assignemts/participantss/moderation note. It's possible that I forgot to check other places, please make sure to only present it at home and topics
- If the "Link a topic/motion/assigment" option is opended there is a warning in the console*^1, please prevent that warning 


*^1 [only part of the waring: 
```
editor-link-dialog.component.html:111 
  It looks like you're using the disabled attribute with a reactive form directive. If you set disabled to true...
```

PR #5997, opened 2026-03-16
Fixed motion.reason width

### LINKED ISSUE #5982: Fix text width of reason in amendments
**Describe the bug**
In amendments, the width of the reason is restricted by "Show entire motion text" which is not nice especially on small devices.

Image

**How to Reproduce**
1. create a motion
2. create an amendment with a long reason
3. view the amendment detail and change the width of your screen to observe how and when the text breaks

**Expected behavior**
The reason uses the entire width of the motion mat card and is not restricted by other elements. (compare with reason in normal motions)

**Additional context**
Pay attention that very long words don't run out of the mat card.

Borreg0:
Closes #5982


Elblinator (changes requested):

Elblinator (approved):

bspekker (changes requested):
Words shouldn't break in the middle when there is enough space for them on the next line.

Image

Please also have a look at the current behaviour for reasons of regular motions: They use the full width of the motion mat card.

bspekker (approved):

Elblinator (approved):

client/src/assets/styles/reason.scss
```diff
@@ -0,0 +1,5 @@
+.motion-text p {
+    word-break: keep-all !important;
+    overflow-wrap: break-word;
+    white-space: collapse;
+}
```
bastianjoel, line 5: This is messing with the motion texts while the fix was only intended for the motions reason. We have complex things in place there. Why does something like this go through review @Elblinator?

PR #6042, opened 2026-03-27
Add meeting time zone field

### LINKED ISSUE #5949: Change timestamps from client to CET
**Current behavior**
Currently the client sends the time stamps as UTC 
`start_time: 1772319600`

Additionally a meeting should have a selected timezone
(https://github.com/OpenSlides/openslides-backend/issues/3387)

**Wanted behavior**
The client should have a new field while creating a meeting which lets you select the time zone and after the meeting is created that field is not changable but presented with the same restrictions the name is restricted

And the send timestamps should be transformed to the selected timezone

reiterl:
Resolve #5949 

Needs https://github.com/OpenSlides/openslides-backend/pull/3438

- add time zone field in meeting edit dialog
- use Intl to get the supported time zone list
- add time zone field to meeting model


reiterl:
There is a patch of the  os4 go repo needed. Allowing to generate timezones fields. 
datastore/dsmodels/gen_models/main.go L.307 add "timezone"

reiterl:
Needs https://github.com/OpenSlides/openslides-backend/pull/3438

I include the field and the timezone for the calculations of meeting.start-time and meeting.end-time. I added the
meeting_time_zone field to import. meeting_start_time should be 'YYYY-MM-DD', so I didn't modify the stuff here.
As discussed, no meeting.update with timezone and no organization_update with timezone.
I didn't find the hooks where 'created*' are added, so I suppose the backend does that.

reiterl:
@luisa-beerboom can you have a look on it?

client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/components/meeting-edit/meeting-edit.component.ts
```diff
@@ -122,18 +124,21 @@ export class MeetingEditComponent extends BaseComponent implements OnInit {
...
         this.createForm();
+        this.initTimezones();
         this.init();
 
         if (this.isCreateView) {
             super.setTitle(ADD_MEETING_LABEL);
         } else {
             super.setTitle(EDIT_MEETING_LABEL);
             this.meetingForm.get(`language`)?.disable();
+            this.meetingForm.get(`time_zone`)?.disable();
```
luisa-beerboom: No editing yet?

luisa-beerboom:
> As discussed, no meeting.update with timezone and no organization_update with timezone.

Didn't we only discuss this for organization?

client/src/app/site/services/time-zone.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Injectable({
+    providedIn: `root`
+})
+export class TimeZoneService {
+    private organizationRepo = inject(OrganizationControllerService);
+    private presenter = inject(GetValidTimezonesPresenterService);
+
+    public getTimeZone(): string {
+        return 'Europe/Berlin';
```
luisa-beerboom: Why is this hardcoded?

client/src/app/site/services/time-zone.service.ts
```diff
@@ -0,0 +1,43 @@
...
+        const year = value.getFullYear();
+        const month = value.getMonth();
+        const day = value.getDate();
+        const timezone = tz ?? 'UTC';
+        return new TZDate(year, month, day, timezone);
+    }
+
+    public transformFromTS(value: number, tz?: string): Date {
+        const timezone = tz ?? 'UTC';
+        return new TZDate(value * 1000, timezone);
```
luisa-beerboom, line 72: Is this intended to recalculate the timestamp and does it do that?

client/src/app/site/services/time-zone.service.ts
reiterl: Ah, this was used for testing.

client/src/app/site/services/time-zone.service.ts
reiterl: This used in the localized date range pipe to create a date for the display. In my test it worked.

client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/components/meeting-edit/meeting-edit.component.ts
reiterl: In the issue https://github.com/OpenSlides/openslides-client/issues/5949 the description
said, no time zone editing after the creation of the meeting. I followed that. Is there a good reason, to allow the meeting-edit view to update the time zone?

luisa-beerboom: Idk, I wrote the backend to allow for it since that's the idea of what should happen, that I got during various discussions.

For potential reasons I would imagine belated editing, political changes and the like, though @MSoeb can probably judge that type of thing better than I can.

reiterl: The new requirements is that organization and meeting time zone are allowed to be updated. See the related issue. I start to work on that.

reiterl:
Added:

- Allow to update time_zone in the meeting-edit form
- Allow to update time_zone in the organization-setting form

AFAIK this is complete. so @luisa-beerboom please review and @Elblinator please test.
To test this the backend pull request with the presenter (s.o) is needed.

Elblinator (dismissed):
Please add a search field in the selection mask

bspekker (changes requested):
There are some open issues:
- [x] The dropdown menu to select a time zone should be searchable (cf. participant selection in list of speakers etc)
- [x] If a meeting has a time zone different from the organisation time zone and you open a motion that shows the meeting time zone, a reload will reset the depicted time to the organisation time zone. 
**-> Wanted behaviour:** The meeting time zone should always be shown when it deviates from the org time zone (maybe cause is a missing subscription?)
- [ ] If the time zone is changed in the organisation setting, some weird things happen: Time zones in the dashboard are not depicted correctly for meetings that have the previous time zone (no time zone is shown) and the new org time zone is shown even if that is not necessary because it is the default as the org time zone. 
**-> Wanted behaviour:** The time zones are only depicted when they differ from the org time zone.
- [ ] Additionally, sometimes an artefact appears: when you click on a meeting, the depicted time zones very shortly change to something else before loading the meeting. Maybe this is solved by the previous point.

Last point: Personally, I'm not a big fan of this long list of time zones with so many entries and the long form of the entries. Do you have a different list that is maybe sorted by and starts with "UTC+XX ..." etc. or is limited only to the ~24 unique time zones? For the depiction in the dashboard and motions, a short format (only something like "UTC+2", "CET", or "GMT+4") would be sufficient (and the dropdown can have a longer version).
If you can't find an easy solution for this right now, we can also have a further look at this at a later time.

reiterl:
> The dropdown menu to select a time zone should be searchable (cf. participant selection in list of speakers etc)

That could be larger work . At the moment all possible time zones are display. Perhaps we can exclude some them. The problem with 'UTC+1' is that you run into problem with summer/winter time and other "political" problems.  "Europe/Berlin" is sometimes CET or CEST for example. 

> If a meeting has a time zone different from the organisation time zone and you open a motion that shows the meeting time zone, a reload will reset the depicted time to the organisation time zone.

That could be a subscription problem, I will look at it.
Note: The timezone is displayed, when browser time zone and used time zone for the calculation differs.

> If the time zone is changed in the organization setting, some weird things happen: Time zones in the dashboard are not depicted correctly for meetings that have the previous time zone (no time zone is shown) and the new org time zone is shown even if that is not necessary because it is the default as the org time zone.

The time zone is displayed, when browser time zone  and used time zone differs. Changing organization time zone can lead to such stuff. 

> Additionally, sometimes an artifact appears: when you click on a meeting, the depicted time zones very shortly change to something else before loading the meeting. Maybe this is solved by the previous point.

The time zones needs to be fetched with a presenter, perhaps that is the reason? Can you give a step to step
of this problem?

In the real world time zones are a politic problem, that stuff is messy. The lib I use understands "Europe/Berlin",
etc. Perhaps we limit the possible time zones to 6-8 and that is it. Then we wouldn't need the search field.

Possible:
- UTC
- Europe/Berlin
- ...

reiterl:
@bspekker If you like the idea of limiting the possible tzs, perhaps provide a list. 
I will work on the subscription problem.

reiterl:
I could use "Etc/GMT+2", etc also as tz. Perhaps a mix of some cities (2-3) and the Etcs and UTC?

reiterl:
> For the depiction in the dashboard and motions, a short format (only something like "UTC+2", "CET", or "GMT+4") would be sufficient (and the dropdown can have a longer version).

Shorten the displayed timezone in the date display in a meeting. I use tzName with option "short" for this.
Also add the short time zone to the localized date range pipe.

client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/components/meeting-edit/meeting-edit.component.ts
```diff
@@ -265,6 +274,7 @@ export class MeetingEditComponent extends BaseComponent implements OnInit {
         };
 
         rawForm[`language`] = [this.orgaSettings.instant(`default_language`)];
+        rawForm[`time_zone`] = [this.timeZone.getOrganizationTimeZone(), Validators.required];
```
luisa-beerboom: Is it right to always initialize the edit with orga timezone? Shouldn't it be meeting timezone?
Or is this replaced later?

client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/components/meeting-edit/meeting-edit.component.ts
```diff
@@ -405,7 +419,7 @@ export class MeetingEditComponent extends BaseComponent implements OnInit {
 
     private enableFormControls(): void {
         Object.keys(this.meetingForm.controls).forEach(controlName => {
-            if (this.isCreateView || controlName !== `language`) {
+            if (this.isCreateView || controlName != `language`) {
```
luisa-beerboom: Why doesn't strict equals work here?

client/src/app/site/services/time-zone.service.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ORGANIZATION_ID } from '../pages/organization/services/organization.service';
+import { OrganizationControllerService } from '../pages/organization/services/organization-controller.service';
+
+@Injectable({
+    providedIn: `root`
+})
+export class TimeZoneService {
+    private aktiveMeetingRepo = inject(ActiveMeetingService);
+    private organizationRepo = inject(OrganizationControllerService);
+    private presenter = inject(GetValidTimezonesPresenterService);
```
luisa-beerboom: Any reason why these aren't injected via constructor?

client/src/app/ui/modules/meeting-time/meeting-time.component.ts
```diff
@@ -11,6 +12,8 @@ import { PipesModule } from '../../pipes';
     imports: [CommonModule, PipesModule]
 })
 export class MeetingTimeComponent {
+    public timeZoneService = inject(TimeZoneService);
```
luisa-beerboom: Again: Why manual injection? Is this a new style guideline I missed?

client/src/app/ui/pipes/localized-date-range/localized-date-range.pipe.ts
```diff
@@ -1,44 +1,60 @@
...
     pure: false,
     standalone: false
 })
-export class LocalizedDateRangePipe extends FormatPipe implements PipeTransform {
-    public override transform(value: any, dateFormat = `PPp`): any {
+export class LocalizedDateRangePipe implements PipeTransform {
+    private timeZone = inject(TimeZoneService);
+    private cd = inject(ChangeDetectorRef);
+    private inputDate = inject(DateFnsConfigurationService);
+    public formatter = new FormatPipe(this.inputDate, this.cd);
```
luisa-beerboom, line 13: What necessitated the move away from extending the `FormatPipe`?

luisa-beerboom:
> Last point: Personally, I'm not a big fan of this long list of time zones with so many entries and the long form of the entries. Do you have a different list that is maybe sorted by and starts with "UTC+XX ..." etc. or is limited only to the ~24 unique time zones? For the depiction in the dashboard and motions, a short format (only something like "UTC+2", "CET", or "GMT+4") would be sufficient (and the dropdown can have a longer version).
> If you can't find an easy solution for this right now, we can also have a further look at this at a later time.

@bspekker 
I could write an entire essay, about why it isn't just ~24 unique time zones and why every single one of these countless zones is probably necessary, even if they have the same rules right now (unless we want to region-lock our time handling accuracy). So limiting the amount of entries really isn't sensible IMO.

As for making access easier:
1. Sorting them alphabetically is already a pretty good solution IMO, as they're automatically sorted by continent
2. There'd be the option of sorting them by current UTC offset ("current" because that changes due to daylight savings), but I fear that'd make the list less readable not more.
3. We could place selected time zone and/or browser time zone at the top by default, that'd be relatively easy and would probably be optimal for most normal cases
4. The entire thing could be treated as an expandable tree as defined by the compartments of the time zone names, but that would require quite a bit of UI design work

I personally like a combination of 1 and 3.

If you want to limit the list regardless, feel free to define whatever compromise list you'd like, but I'm going to point out right now, that you'd be excluding anyone, not exactly covered by that list, from having accurate time interpretation and security from political changes (in relation to time zones).

client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/components/meeting-edit/meeting-edit.component.ts
reiterl: This is initialized for the create form, the values are replaced later.

client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-meeting/components/meeting-edit/meeting-edit.component.ts
reiterl: typo

client/src/app/site/services/time-zone.service.ts
reiterl: No, changed into injected with constructor.

client/src/app/ui/modules/meeting-time/meeting-time.component.ts
reiterl: This will also be changed to constructor injection.

client/src/app/ui/pipes/localized-date-range/localized-date-range.pipe.ts
reiterl: Because the signature of the transform method changes (add the timezone) so I used delegation here.

luisa-beerboom (approved):

reiterl:
I used a list-search-selector and some glue code, to add a search field to the time zone fields. 
It still have some caveats. Perhaps I can fix them later.

reiterl:
Ah, the list-search-selector has a glitch, I can reproduce the caveat with main and the participant gender field.
This would be stuff for another issue.

bspekker (approved):
Looks good!

I can't reproduce the issues when loading a meeting anymore, so we have to look out if it still happens in the future.

Thank you for the remarks on different lists for time zones!
I like the current compromise of having the short version in the dashboard and motions and the longform in the dropdown. The search also helps to find your target tz easier than scrolling through the entire list.

client/src/app/site/services/time-zone.service.ts
bastianjoel: @luisa-beerboom `inject` is the preferred way to add dependencies.

PR #6061, opened 2026-04-02
Remove color while pasting text

### LINKED ISSUE #5961: Editor: remove assumed colour also for dark mode while inserting text
**Current behavior**
The editor in some cases copied the background colour and font color 
I saw last week that this also happens in dark mode and is relevant for our customers

**Wanted behavior**
the standard font and background colour should also be removed from the editor

**Additional context**
The same for the light mode is here: 
https://github.com/OpenSlides/openslides-client/pull/5859

MoonCubes:
Resolves: #5961
I'm not really happy with hard coding a background color,
so I think it might be best if all formatting in general (or just color information) were removed from pasted text,
since it mostly causes annoyances
(even tough that could also remove wanted formatting).


Elblinator:
> I'm not really happy with hard coding a background color,
so I think it might be best if all formatting in general (or just color information) were removed from pasted text,
since it mostly causes annoyances
(even tough that could also remove wanted formatting).

there were discussions about this as the wanted behaviour. 
And we decided that if the is extra highlighting or font colour then removing weverything with the past ist extremely annoying. So we only want to remove the default colours

Elblinator (approved):
Function

Elblinator (approved):
code

PR #6113, opened 2026-04-22
Update injectable providedIn from a module to 'root'

### LINKED ISSUE #5970: Injectables provided in module deprecated
We are currently using some `@Injectable` decorators that have the option `providedIn: <Module>`.
This is deprecated and should be replaced.

reiterl:
Resolve #5970 

Updates the following parts:
agenda
committee
mediafiles
meeting
motion
organization-tag
participants
projection
projector
theme
topic

It needs to be reviewed and tested. I tested it myself, here all my tests passed. 
There is still a lot to do here.  (39 classes must be improved).

### DIFF
--- a/client/src/app/gateways/export/csv-export.service/csv-export-for-backend.service.ts
+++ b/client/src/app/gateways/export/csv-export.service/csv-export-for-backend.service.ts
@@ -1,7 +1,6 @@
 import { inject, Injectable } from '@angular/core';
 import { BaseViewModel } from 'src/app/site/base/base-view-model';
 
-import { ExportServiceModule } from '../export-service.module';
 import { FileExportService } from '../file-export.service';
 import {
     BackendCsvColumnsDefinition,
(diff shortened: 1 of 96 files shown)


bastianjoel (changes requested):
Why does this only solve part of part of #5970?

I do not like doing this in multiple steps as this requires a lot of testing.

reiterl:
It involved only parts of the issue, because I have finished only half of it.  But okay, I can try to do it as whole.

reiterl:
I think, I have found every injectable providedIn module and replaced it, I had some problems with the autoupdate service and polls, which I also have with main.

bastianjoel (approved):

bastianjoel:
From a code perspective the is fine if it works. 

However - I want to point out that it might had been better to provide some of those services for route groups. This would have been closer to the current behavior as services could get torn down in some cases which might lead to different behavior after reinitializing. If problems arise while testing I will insist on changing this for all modules that are loaded lazily by their dependents.

bspekker (approved):

Elblinator (approved):

Issue #6114, opened 2026-04-22
Clean-up account-dialog-main


Elblinator:
- The componente should extend BaseModelRequestHandlerComponent and not BaseMeetingModelRequestHandler
   - then it should also implement onShouldCreateModelRequests, with hideWhenDestroyed
  - getSubscriptions should be removed
   - and as standalone:true (so the standalone part just removed)

reiterl:
@Elblinator Can we close here? Everything except the standalone removal are fixed.

Elblinator:
Yes we can close here

PR #6118, opened 2026-04-23
Cleanup of the account dialog main component

reiterl:
Resolve part of #6114

Use a different method to get the Subscription with hideWhenDestroyed also use a more general handler.

It doesn't include the standalone removal. This would lead to dependencies import /export problems with a broken mix of standalone and module components.


Elblinator (approved):
function

bastianjoel (approved):

Issue #2504, opened 2023-07-07
amendments: change recommendation diff doesn't work


MSoeb:
**Description:** Recommendations with crossed out lines done in amendments are not displayed. The line will just disappear without a diff-hint. This isn't optimal. It is not easy to recognize, what content is crossed out. 

**Original amendment:**
![Screenshot_20230707_105842](https://github.com/OpenSlides/openslides-client/assets/105491030/eb605ae2-a8e3-471c-a2f8-5e3d1e7c7461)

**Crossed out sentence:**
![Screenshot_20230707_105933](https://github.com/OpenSlides/openslides-client/assets/105491030/9fb643cf-8b9b-44b0-a6b8-16fc41e1e96b)

**Reproduction:**
1. open a meeting
2. open motions
3. create a motion
4. add an amendment to the motion
5. add a change recommendation to the amendment content by deleting a word or more content.

**What should happen:**
The diff should display in the amendment the crossed out line done by the change recommendation. With this change users could easily spot diffs. Also the diff would be visible in the projection.

Note: The issue is linked to #2503

CatoTH:
<img width="1484" alt="Bildschirmfoto 2023-08-27 um 16 44 52" src="https://github.com/OpenSlides/openslides-client/assets/533440/fa2268b4-9ed3-4df1-8717-90fd594bce2f">
I could unfortunately not really replicate this right away... in the diff view, the striked out text appeared striked out, as in the screenshot. It was not really recognizable that this has changed in comparison to the original amendment, but it was striked out. Is there something particular to take into consideration when creating the change recommendation to trigger this?

bastianjoel:
I think it is important that something was added to the end of a paragraph of the amendment. 

So for example with motion text:
```html
<p>Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim.</p>
<p>Donec pede justo, fringilla vel, aliquet nec, vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus.</p>
```

[...]

PR #6168, opened 2026-05-11
Fix out of scope amendment change recommendations

### LINKED ISSUE #6157: Change recommendation: Doesn't work with completly deleted text in amendments
**Describe the bug**
It is not possible to create a change recommendation for an amendment whose text has been completely deleted. The process will crash with the following  error.

```
TypeError: Cannot read properties of null (reading 'parentNode')
```

**How to Reproduce**
1. open a meeting
2. open motions
3. create a motion
4. create an amendment and delete all. 
5.  Save amendment
6.  open it
7. create a change reco to deleted part -> crash

**Expected behavior**
It should be possible to create a change reco to an amendment which only has deleted text

bastianjoel:
resolves #6157

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-change-recommendation-controller.service/motion-change-recommendation-controller.service.ts
+++ b/client/src/app/site/pages/meetings/pages/motions/modules/change-recommendations/services/motion-change-recommendation-controller.service/motion-change-recommendation-controller.service.ts
@@ -201,25 +201,32 @@ export class MotionChangeRecommendationControllerService extends BaseMeetingCont
         lineNumberedParagraphs: string[],
         lineRange: LineRange
     ): any {
-        const consolidatedText = lineNumberedParagraphs.join(`\n`);
-
-        const extracted = amendment
-            .services()
-            .diff.extractRangeByLineNumbers(consolidatedText, lineRange.from, lineRange.to);
-        const extractedHtml =
-            extracted.outerContextStart +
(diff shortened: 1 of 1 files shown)


Elblinator (approved):
cöde

MSoeb (changes requested):
In general: Change reco work again. nice :)

Nevertheless a small change is needed.

After adding a change reco, the diff view won't show the whole amendment text anymore. 

Example of diff view:
Image

Original amendment:
Image

Wanted behavior: The diff view should show the whole amendment not only the line with the change reco.

What have I done?
I've added to the first line of the amendment a change reco.

bastianjoel:
This is because change recos on all crossed out text should be expanded to the full paragraph line range. This is a different issue and as far as I know there is already an issue open for that.

Elblinator:
I think the Issue @bastianjoel means is: https://github.com/OpenSlides/openslides-client/issues/2504

Given that that Issue has "needs discussion" and that we want to produce a release soon I would remove that part 100% from here and pull it into the discussion from the other issue

MSoeb:
Talked with @bastianjoel about his point. We've made the decision to solve the mentioned problem with another issue. (https://github.com/OpenSlides/openslides-client/issues/6205)

But, during testing another error occured which needs to be solved: Change recommandation in the scenario above won't appear in the final version. @bastianjoel will create an issue for this.

MSoeb (approved):
The general problem is solved. The further problem, as described in my last comment, will be solved in another issue.

Therefor I approve this issue.

Issue #6129, opened 2026-04-28
Parliamentary Mode: Wrong Display of Speaking Times


MSoeb:
**Describe the bug**
When Parliament Mode is enabled, speaking times are displayed incorrectly in the List of Speakers if accessed via a specific path.

Here is an example of the error:
<img width="833" height="370" alt="Image" src="https://github.com/user-attachments/assets/57800410-e61f-4cc2-9511-b821a0dc24fc" />

An error message also appears in the console.
<img width="351" height="307" alt="Image" src="https://github.com/user-attachments/assets/7548e79b-45d8-4eb5-805d-3f179d4f71c3" />

**How to Reproduce**
1. open an openslides instance with a meeting, that has the parliament mode with demo data
2. open participants > 3-dot-menu > contributions
3. open now an agenda item, that have a list of speakers with speakers on it. 
4. .> 3-dot-menu > List of speakers

If this doesn't lead to the error on first try, do a reload at start of step 2. 

**Expected behavior**
Speaking times should be displayed correctly.

**Additional context**

PR #6178, opened 2026-05-12
Prevent Error in parliamentary mode

Elblinator:
part of #6129

I do not know where the nananana comes from I could not reproduce that behaviour but I could prevent the errors in the console


bastianjoel (approved):

bspekker (approved):
looks great!

PR #6184, opened 2026-05-13
Fix time component 

### LINKED ISSUE #6156: Motion forward adjust with time_zone
**Describe the bug**
If you try to forward a motion then the following Erros are in the colnsole: 
```
ERROR TypeError: Cannot read properties of undefined (reading 'time_zone')
    at MeetingTimeComponent_ng_template_3_Template (meeting-time.component.html:16:9)
```

**How to Reproduce**
1. Allow your committee to forward to itself
2. Create a meeting (with or without time_zone)
3. activate the forwarding in the motion workflow
4. try to forward a motion
5. the dialog opens
6. you cannot select any meeting
7. The Error is in the console

**Expected behavior**
You should be able to select some meeting and the Error should not be thrown

Elblinator:
resolves  #6156


bastianjoel (approved):

bspekker (approved):
very pretty

PR #6186, opened 2026-05-13
Remove all colour while pasting

### LINKED ISSUE #6180: Editor: Pasting with background color error
**Describe the bug**
Pasting text with background color, i.e. from an OpenSlides motion reason in dark mode, doesn't work correctly. Span elements are still copied. These elements will lead to problems with motion diffs. 

**How to Reproduce**
1. open a meeting
2. turn to light mode
3. create a motion ^1
4. turn to darkmode
5. copy the text from the first motion (do no copy source code or the text in editor, copy the displayed text in the motion detail
6. create a new motion with the copy paste text
7. create an amendment (light or dark mode does not matter) which does not remove the text which was there but adds new 
8. The amendment display will be wrong: 

Image

 ^1```
<p> Platz </p>
```

**Additioonal info:**
Apparently the ```ClearTextColorPaste``` does not remove empty spans, which are created if the font and background colour is removed

**Expected behavior**
The motion text should be saved without empty spans 
so that amendments can be created without the wrong appearance

**Additionally:**
@MSoeb: For documentation: During discussion with @emanuelschuetze , @bspekker , @bastianjoel and me, the decision was made to re-add the italic button for limited motion editor.

Elblinator:
resolves #6180

### DIFF
--- a/client/src/app/ui/modules/editor/components/editor/extensions/clear-textcolor.ts
+++ b/client/src/app/ui/modules/editor/components/editor/extensions/clear-textcolor.ts
@@ -1,6 +1,5 @@
 import { Extension } from '@tiptap/core';
 import { Plugin } from '@tiptap/pm/state';
-import tinycolor from 'tinycolor2';
 
 export const ClearTextcolorPaste = Extension.create({
     name: `clear-textcolor-paste`,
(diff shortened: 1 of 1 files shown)


bastianjoel (changes requested):
I will not approve such changes. 

You always have the option to remove styling via the button or just paste without style. Before completely restricting the user to paste colored text we should discuss other options.

Elblinator:
This will be fixed differently hence I'll close this PR

Issue #6152, opened 2026-05-07
[vote] Implement new vote service


bastianjoel:
- [ ] Create forms
  - Content object specific form dialogs 
    - [x] Topic
    - [x] Motion
    - [x] Assignment
      - [x] Hide Selection/per Candidate forms if only one candidate available
      - [x] Score should be available for single candidate
  - [x] Analog polls
  - [x] Base form
  - [ ] Edit forms
  - Config specific forms
    - [x] Selection
    - [x] Approval
    - [x] Rating approval
    - [x] Rating score
    - [x] Add option for select polls to display table instead of chart
- [x] Meta info component
- [x] Poll progress
  - [x] Selection
  - [x] Approval
  - [x] Rating approval
  - [x] Rating score
- [x] Voting
  - [x] Selection
  - [x] Approval
  - [x] Rating approval
  - [x] Rating score
  - [x] Submit vote button validations
  - [x] Display propper message when user has delegated their vote
- [x] Result view #6250
  - [x] Selection
  - [x] Approval
  - [x] Rating approval
  - [x] Rating score
  - [x] 100%-base - Details see internal document in `Entwicklung/interne Projekte/Darstellung Abstimmungsergebnisse`
  - [x] General options should be displayed above the totals
- [x] Detail views #6251
  - [x] Topic
  - [x] Motion
  - [x] Assignment
  - [x] Config specific single votes list
  - [x] Base component
- [x] Result pdf export
  - Content object specific adjustments
    - [x] Topic
    - [x] Motion
    - [x] Assignment
  - Config specific export
    - [x] Selection
    - [x] Approval
    - [x] Rating approval
    - [x] Rating score
- [x] Ballot paper pdf export
  - [x] Selection
  - [x] Approval
  - [x] Rating approval
  - [x] Rating score
- [x] Meeting settings defaults
  - [x] Add new category for voting
  - [x] Topic
  - [x] Motion
  - [x] Assignment
  - [x] Pass to forms #6418
  - [x] Add `motion_poll_projection_name_order_first` to reset meeting settings
- [ ] Add integration test for vote service

Issues that should be solved while at it:
- #6150 
- #6049
- #6035
- #5828
- #5480
- #5425 
- #5134
  - [x] Add detail view for topic polls
  - [x] Add approval polls to topics
- #3795
- #3620
- #3025
- #2714
- #2583
- #2499
- #2412
- #6308

Maybe:
- #2323

Elblinator:
[...]

PR #6187, opened 2026-05-13
Improve agenda and poll detail pdf

emanuelschuetze:
Agenda pdf:
- use nowrap in identifier column 
- move identifier of subitems in title column
- remove "Table of contents" h2 title

Poll detail pdf:
- use nowrap in index column
- reduce column width of vote-weight and has-voted
- use getFullName instead getShortName in user tables

### DIFF
--- a/client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
+++ b/client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
@@ -74,6 +74,7 @@ export interface TocLineDefinition {
     identifier: string;
     title: string;
     pageReference: string;
+    level?: number;
     style?: StyleType;
     fillColor?: string;
 }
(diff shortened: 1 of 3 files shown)


bastianjoel:
Might be a good idea to create an issue for the poll related changes and add it to the list in #6152. Otherwise your changes are likely to be dropped after the new vote service.

client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
```diff
@@ -74,6 +74,7 @@ export interface TocLineDefinition {
     identifier: string;
     title: string;
     pageReference: string;
+    level?: number;
```
Elblinator, line 77:
```suggestion
    isTopicSubitem?: boolean;
```

Elblinator (changes requested):
please also create issue

PR #6188, opened 2026-05-13
Show correct name in committee detail if forwarding is active

### LINKED ISSUE #5075: Show correct amount in "Receive motions from" and "Forward motions to"
context: 
<s>https://github.com/OpenSlides/openslides-autoupdate-service/issues/1216</s>
FOR TESTING NEEDED: https://github.com/OpenSlides/openslides-autoupdate-service/pull/1365

**Current behavior**
(correct) list view

Image

detail view: 
from committee admin
Image

from superadmin:
Image

Edit view: 
from committee admin:
Image

superadmin:
Image

**Wanted behavior**

- [ ] in the detail view: the complete list from "forward motions to" and "Receive motions to" should be shown
- [x] in the edit view:
the committee admin should see every meeting which they can see/get from the autoupdate.
There are two cases in which the committee admin can see a committee
1) the committee admin is directly or indirectly the committee admin from the committee
2) the committee can already forward to this committee while the committee admin is not committee admin from that committee

Case A should always be not diabled (currently the case)
Case B should be disabled (that is missing)

The payload to the backend should not deviate from the current payload. 
Even though the committees are not shown to a committee admin they are being send to the backend. If they are not send the backend returns an (correct) Error message.

Elblinator:
resolves #5075

### DIFF
--- a/client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-view/components/committee-detail-view/committee-detail-view.component.html
+++ b/client/src/app/site/pages/organization/pages/committees/pages/committee-detail/modules/committee-detail-view/components/committee-detail-view/committee-detail-view.component.html
@@ -62,10 +62,13 @@
                 {{ committee.description }}
             </div>
             <!-- forward motions to -->
-            @if (committee.forward_to_committees.length > 0 && isOrgaAdmin()) {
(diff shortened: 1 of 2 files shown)


bastianjoel (approved):

bspekker (approved):
Detail view works as intended.

PR #6217, opened 2026-05-20
Add Committeelink to meeting

### LINKED ISSUE #6055: Add clickable committee link on the head of the toolbar
**Current behavior**
If a person has more than one meeting or has any higher admin role (super/orga/account/committee) then if you are in a meeting a link to navigate directly to the committee from your meeting would be rather useful.

**Proposal**
Before the meeting name could be the "layer" icon and the icon could be a hyperlink to navigate directly to the committee-detail
Image

Or the meeting name could be the hyperlink (but the string being the meeting name would be a bit confusing there)

**Additional context**
If a user has only one meeting and no special admin role (see above) then they cannot see the committee and therefore should not be directed anywhere

This is a feature which is wanted by @m-schieder and me and potentionally more people

**DISCUSSION:**
What is wanted needs to be specified

Elblinator:
resolves #6055

### DIFF
--- a/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.ts
+++ b/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.ts
@@ -16,7 +16,9 @@ import { GlobalSearchComponent } from '../global-search/global-search.component'
 export class GlobalHeadbarComponent {
     public isSearchEnabled = true;
 
-    public committeeLink = ``;
+    public get committeeLink(): string {
+        return `/committees/` + this.activeMeeting?.meeting?.committee_id;
+    }
 
     public get displayName(): string {
         if (this.activeMeeting.meeting) {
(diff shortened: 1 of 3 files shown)


client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.ts
```diff
@@ -15,6 +16,8 @@ import { GlobalSearchComponent } from '../global-search/global-search.component'
 export class GlobalHeadbarComponent {
     public isSearchEnabled = true;
 
+    public committeeLink = ``;
```
bastianjoel:
```suggestion
    public get committeeLink() { .....;
```

bastianjoel (changes requested):

client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.ts
```diff
@@ -25,9 +30,20 @@ export class GlobalHeadbarComponent {
...
+        ) {
+            return true;
+        }
+        return false;
+    }
+
     public constructor(
         private activeMeeting: ActiveMeetingService,
         private orgaService: OrganizationService,
+        private operatorService: OperatorService,
```
bastianjoel: The preferred way to inject services is now via `inject`. 

E.g. `private operatorService = inject(OperatorService);`

bastianjoel (approved):

bspekker (changes requested):
Function is great!
Can you please make the icon a bit smaller (18px) it looks very big and not balanced right now.

bastianjoel (approved):

bspekker (approved):

PR #6284, opened 2026-06-12
Add aria labels and translate them

### LINKED ISSUE #6140: Add aria-labels to LoS buttons
**Current behavior**
The buttons from a LoS do not have a label and are therefore a bit harder to use

**Wanted behavior**
Every button from a LoS should have an aria-label which is the text on the button

### LINKED ISSUE #6283: Add aria labels candidate list and global search
**Describe the bug**
same as https://github.com/OpenSlides/openslides-client/issues/6140

The add me and remove me labels are missing in the assignment list

 and the global search label is not being translated

**Expected behavior**

The missing labels should be added and the not translated label should be translated

Elblinator:
resolves https://github.com/OpenSlides/openslides-client/issues/6140
resolves https://github.com/OpenSlides/openslides-client/issues/6283

Adds Aria label to every button from the LoS and assignment list and makes translation of gloabel search label possible

### DIFF
--- a/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.html
+++ b/client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.html
@@ -11,7 +11,7 @@
     <div class="flex-center">
         <!-- Button to open the global search -->
         @if (isSearchEnabled) {
-            <button aria-label="Global Search" mat-icon-button role="Search" (click)="openSearch()">
+            <button aria-label="{{ 'Global search' | translate }}" mat-icon-button role="Search" (click)="openSearch()">
                 <mat-icon>search</mat-icon>
             </button>
         }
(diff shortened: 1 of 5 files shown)


client/src/app/site/modules/global-headbar/components/global-headbar/global-headbar.component.html
```diff
@@ -11,7 +11,7 @@
     <div class="flex-center">
         <!-- Button to open the global search -->
         @if (isSearchEnabled) {
-            <button aria-label="Global Search" mat-icon-button role="Search" (click)="openSearch()">
+            <button aria-label="{{ 'Global search' | translate }}" mat-icon-button role="Search" (click)="openSearch()">
```
bastianjoel: Never use template tags within attributes.

`[attr.aria-label]="'Global search' | translate"`

bastianjoel (changes requested):

bastianjoel (approved):

bspekker (approved):

PR #5693, opened 2025-12-23
Motion card content fixed for all screen sizes

Borreg0:
Should resolve #5026

Issue #5026, opened 2025-05-26
Motions: Prevent content to extend the motion card


bspekker:
**Describe the bug**
If some content in a motion does not wrap and extends the boundaries of the card, this might cause other elements to extend the card as well (e.g. reason, change recs, summary of changes, motion text toolbar) 

**How to Reproduce**
1. create a motion with a reason that contains one string without spaces.
2. save the motion
3. create a change recommendation to the motion

--> the display of the change rec and reason will look similar to this:
![Image](https://github.com/user-attachments/assets/a919d465-bb28-4fdb-8276-316eb0e61a46)

**Expected behavior**
Make sure that content in the motion card will never extend the boundaries of the motion card.

PR #6285, opened 2026-06-12
Revert changes from #5997 and #5693

### LINKED ISSUE #6256: Fix line breaks in changes to lists in motions
**Describe the bug**
If a motion has changes to lists, the line breaks in the middle of words. That is not nice.

Example with 1. change rec to unordered list, 2. amendment to ul, 3. change rec to ol, 4. amendment to ol:
Image

**How to Reproduce**
1. create a motion with an ordered/unordered list
2. create amendments/change recs to these lists (with a bit longer text)
3. observe the line breaks

**Expected behavior**
Line breaks don't appear in the middle of words - even in lists.

**Additional context**
Changes to "normal" text and the display in the projector are not concerned.

bastianjoel:
resolves #6256

we need to reopen #5982 and #5026 after merge. Both fixes are not valid and could introduce more side effects than the already observed ones.

### DIFF
--- a/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-content/motion-content.component.scss
+++ b/client/src/app/site/pages/meetings/pages/motions/pages/motion-detail/pages/motion-view/components/motion-content/motion-content.component.scss
@@ -1,6 +1,4 @@
 @use 'src/assets/styles/mixins/detail-view-appearance.scss';
-@use 'src/assets/styles/variables.scss';
-
 .extra-data {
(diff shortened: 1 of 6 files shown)


Elblinator (approved):
code

bspekker (changes requested):
#6256 is not resolved and the word wrapping is broken everywhere now (for main motion text as soon as screen is smaller than standard width)

bspekker (approved):
works as intended

bastianjoel:
reopened #5982 and #5026

PR #6303, opened 2026-06-15
Remove self from delete payload in participant-list

### LINKED ISSUE #6190: Fix delete admin-self in multiselect
**Describe the bug**
This is a continuation of #5818 
When the selection in multiselect contains your own account as admin and other participants, it is irritating that the delete button is not available at all.

**How to Reproduce**
1. go to the participant list
2. with multiselect, select your own account and other accounts
3. navigate to menu in top right -> the delete option is not available

**Expected behavior**
Deleting in multiselect is possible and the admin-self is skipped. 
Maybe add a snackbar warning that says something like "X users were deleted, you were skipped"

Elblinator:
Resolves #6190

### DIFF
--- a/client/src/app/site/modules/user-components/components/user-multiselect-actions/user-multiselect-actions.component.html
+++ b/client/src/app/site/modules/user-components/components/user-multiselect-actions/user-multiselect-actions.component.html
@@ -8,7 +8,7 @@
     <span>{{ 'Deselect all' | translate }}</span>
 </button>
 
-@if (canUpdate) {
+@if (canUpdate()) {
     <mat-divider />
     <!-- <ng-container *ngIf="implicitContent">
   <ng-container *ngTemplateOutlet="implicitContent"></ng-container>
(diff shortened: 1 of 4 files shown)


bastianjoel (approved):

avinashkamat48 (commented):
There is a signal migration bug left in `sendInvitationEmails()`: `selectedUsers` is now an `input<ViewUser[]>([])`, but the confirmation text still uses `this.selectedUsers.length`. That reads the function object's arity instead of the selected row count, so the prompt can show the wrong number of emails. This should use `this.selectedUsers().length` (and it is worth checking the template for the same pattern, since signal inputs must be invoked there too).

bspekker (changes requested):
The menu in the multiselect behaves weirdly:

Image

I have selected one user that the meeting admin has full control over. 
-> The greyed out options should be available.

bspekker (changes requested):
Previous problems were solved but there is one regression from https://github.com/OpenSlides/openslides-client/pull/5818: In the participant list view in  meeting via the 3-point-menu on my account, I can remove my own account as admin from the meeting.

bspekker (approved):

bastianjoel (approved):

PR #6307, opened 2026-06-16
Add direct link from motion title to in to motion detail

### LINKED ISSUE #6136: Change H1 place/behaviour of motion-list
**Describe the bug**
In a motion list the H1 structure element is not on the navigatable link from the motion but on the title of the motion.

So the title from the motion can be navigated to (and hence, be read out loud)
But if you navigated to the title (with the structure elements) you have to go one back to be able to navigate to the link for the motion

**How to Reproduce**
1. [use the NVDA screen reader and navigate via the headers]
2. have at least one motion
3. go to the lotion list
4. navigate to the header 1 of the motion
5. press enter
6. nothing happens
7. press shift + tab
8. press enter 
9. you navigate to the motion

**Wanted behavior**
If you navigate to the motion title (which could be read out loud) then you should be able to just press enter to navigate into the motion

**Additional context**
The H1 element probably has be removed and/or moved to the navigateable 'a' element

The visual appearance should not change!!

Elblinator:
resolves #6136 
@bspekker: I am sorry I do not know how to test this correctly in our linux setup
The described behaviour happens if you navigate directly to the H1 (screenreaders lik NVDA do that) and then try to navigate into the motion via enter
I could "test" this while deleting the 'a' via the console and then still being able to navigate to the motion via the title

@bastianjoel: The second commit introduced inject, moved variables and removes some comments. 
I hope you agree on the redundance of the comments I removed and the new order of variables


client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.ts
```diff
@@ -46,8 +44,35 @@ interface TileCategoryInformation {
     standalone: false
 })
 export class MotionListComponent extends BaseMeetingListViewComponent<ViewMotion> implements OnInit {
+    public motionRepo = inject(MotionControllerService);
```
bastianjoel: I would prefer keeping those directly over the constructor.

client/src/app/site/pages/meetings/pages/motions/pages/motion-list/components/motion-list/motion-list.component.html
```diff
@@ -124,7 +124,11 @@ <h1 class="mock-h2" translate>Motions</h1>
                 <os-detail-view-navigator [model]="motion"></os-detail-view-navigator>
             }
             <div class="column-title innerTable overflow-hidden">
-                <h1 aria-label="{{ getAriaLabel(motion) }}" class="title-line ellipsis-overflow">
+                <h1
+                    aria-label="{{ getAriaLabel(motion) }}"
```
bastianjoel: [attr.aria-label]

bastianjoel (changes requested):

bastianjoel (approved):

bspekker (approved):

bastianjoel (approved):

PR #6309, opened 2026-06-17
Adds Maximum amount of yes votes to polls

potatojuicemachine:
This PR adds the maximum amount of yes votes to the frontend.

This PR depends on this PR: https://github.com/OpenSlides/openslides-backend/pull/3623
This PR related to this overarching PR: https://github.com/OpenSlides/OpenSlides/pull/7134

This is how the creation of a poll looks:
image
The maximum amount of yes votes will only appear if the user selects YN or YNA voting. I am not entirely sure if I like the current style by just inserting it into the current grid, but I also didn't want to rework the grid system.

This is how the ballot information looks like:
image

This is how the ballot looks like if the maximum amount of yes votes is reached:
image

To get this functionallity I introduced a system to disable individual vote buttons (isActionButtonDisabled).
When I first implemented this when disabled, the buttons had a alternating color. I disabled this in the global-components-style.scss. Not sure if this is the preffered way.

This PR allows for but doesn't add translations.


bastianjoel:
Closing in favor of #6390

Issue #6326, opened 2026-06-23
Orga settings: Save button active without a change done


MSoeb:
**Describe the bug**
Currently in several setting menus the save button is active, even if no changes are done. 

Affected setting menus:
- Organisation settings

**How to Reproduce**
1. open one of these seeting menus

**Expected behavior**
The Save button should only be clickable if changes are done.

bastianjoel:
This is basically caused by two problems. 

1. Ridiculously over-complicated implementation of time zones
2. The mechanism behind `shouldPropagateOnRegistering` in `base-form-field-control.ts`

Point `1.` will be fixed in #6487 as it is straight forward. For `2.` I will open a new issue. 
This also means that meeting and committee edit mask will be fixed with the new issue. 

I updated the issue description.

bastianjoel:
Follow up issue: #6499

PR #6344, opened 2026-07-01
Organization settings and time zone

### LINKED ISSUE #6343: Organisation settings save button stays disabled after making changes
**Describe the bug**
If i change things on the organisational settings, like changing the name of the organisation, the save button on the top stays greyed out. 

**How to Reproduce**
Version without set time zone
0. No Time zone should be set
1. open orga settings 
2. try changing orga name 
3. save button stays disabled

Version with set time zone
0. Time zone is set
1. open orga settings
2. Time zone is disabled and cannot be set
3. try changing orga name 
4. save button works
5. time zone is still not settable

**Expected behavior**
The save button gets enabled and can be click, changes get saved.

**Screenshots**
Image

reiterl:
Resolve #6343

Work on the organization settings and time zone problem.
This is draft.

Not working:
1. Reset of a timezone change  ( os-list-selector doesn't display the changed value)
2. A new organization (fresh new) doesn't have a time zone, so set 'UTC', but it needs to save it.

### DIFF
--- a/client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
+++ b/client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
@@ -181,10 +181,4 @@ export class OrganizationSettingsComponent extends BaseComponent {
             .then(() => this.markFormAsClean())
             .catch(this.raiseError);
     }
-
-    public observeClickNotFound(): void {
-        // the list search selector checks if something observes clickNotFound
-        // else it will disable itself (sometimes).
-        console.warn('Timezone: Click not found');
-    }
 }
--- a/client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
+++ b/client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component } from '@angular/core';
-import { UntypedFormBuilder, UntypedFormGroup, Validators } from '@angular/forms';
+import { UntypedFormBuilder, UntypedFormGroup } from '@angular/forms';
 import { _ } from '@ngx-translate/core';
 import { TranslateService } from '@ngx-translate/core';
 import { BehaviorSubject } from 'rxjs';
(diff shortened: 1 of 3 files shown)


reiterl:
Some part of the reset settings doesn't work. @Elblinator said, that that part should be done in another issue.

client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
```diff
@@ -179,4 +181,10 @@ export class OrganizationSettingsComponent extends BaseComponent {
             .then(() => this.markFormAsClean())
             .catch(this.raiseError);
     }
+
+    public observeClickNotFound(): void {
+        // the list search selector checks if something observes clickNotFound
+        // else it will disable itself (sometimes).
+        console.warn('Timezone: Click not found');
+    }
```
bastianjoel: I don't see why this should be kept.

Elblinator:
The part that the save button is wrongly enabled is documented in another Issue. 
and will be fixed in the future but not while fixing this bug
https://github.com/OpenSlides/openslides-client/issues/6326

bastianjoel (changes requested):

client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
reiterl: In the ListSearchSelectorComponent L.40 is a 
```
this.disabled = ... && !this.clickNotFound.observed;
```
in my test this sometimes leads to a disabled time zone field. So add this observer to the event emitter.

reiterl: Do you have a better solution?

bastianjoel: So you are exploiting a feature that is used allow values that are not in the list to prevent the input from getting disabled?
This is not a valid fix.

Elblinator (dismissed):
Seems to work

client/src/app/site/pages/organization/pages/settings/modules/settings-detail/components/organization-settings/organization-settings.component.ts
reiterl: The clickNotFound part of this fix is removed.

PR #6346, opened 2026-07-02
Add translation marker for is not locked out

### LINKED ISSUE #6249: Add translation label for "is not locked out"
**Current behavior**
In the participant list, the locked out filter option for not locked out can not be translated to other languages.

**Wanted behavior**
The filter option "Is not locked out" can be translated.

Elblinator:
resolves #6249 

I added the translation marker, the new translation is not yet translated. 

To test the marker you need to add a custom translation to see that the string gets changed by it


bastianjoel (approved):

bastianjoel:
I don't think a review from @bspekker makes sense here.

PR #6363, opened 2026-07-07
Add dependDisableOnEmptyItems and use it in this case

### LINKED ISSUE #6327: Meeting edit: page reload leads to disabled field
**Describe the bug**
If a page reload is performed in the meeting edit mask, the duplication field becomes unusable.

**How to Reproduce**
1. open a meeting edit mask > reload > disabled fields

**Expected behavior**
Even after a reload the mentioned fields should be usable in the meeting edit mask.

reiterl:
Resolve #6327 

Add an option to ignore the permanent disable and set it in the duplicate from case.

### DIFF
--- a/client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
+++ b/client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, Input, Optional, Self, ViewEncapsulation } from '@angular/core';
+import { ChangeDetectionStrategy, Component, Input, input, Optional, Self, ViewEncapsulation } from '@angular/core';
 import { NgControl } from '@angular/forms';
 import { MatFormFieldControl } from '@angular/material/form-field';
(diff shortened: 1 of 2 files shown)


bastianjoel (commented):
.

bastianjoel (changes requested):

client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
```diff
@@ -37,7 +39,10 @@ export class ListSearchSelectorComponent extends BaseSearchSelectorComponent {
                 this.selectableItems = items;
                 if (this.contentForm) {
                     this.disabled =
```
bastianjoel, line 41: `this.disabled` is a field with `@Input` decorator. It is not supposed to be set within such control flows in the first place. 
A possible valid solution to your problem might be to override the getter of the `disabled` property, add a new property `isListEmpty` to the `ListSearchSelectorComponent` and then check in the getter if `this.isListEmpty` is set to `false`.

bastianjoel (changes requested):

reiterl:
Closed in favor of #6380

Issue #6273, opened 2026-06-05
History: Not always usable if elections is empty


MSoeb:
**Describe the bug**
The history can't be used, if elections is empty and the user selects it first. After that, it is not possible to use the history anymore. The second selection box is grayed out.

Example - module field and selections box to the right are active:
<img width="799" height="192" alt="Image" src="https://github.com/user-attachments/assets/3a5706bd-f08c-4b25-8625-82668ece2cca" />

Elections is selected - box to the right is grayed out. affects also other module selections.
<img width="804" height="196" alt="Image" src="https://github.com/user-attachments/assets/7202be19-3c68-4f42-a45c-611317b8b333" />

**How to Reproduce**
Precondition: A meeting with one motion and one user
1. open a meeting
2. open history > select module elections > try to use box to the right

**Expected behavior**
If elections is empty, the history should work properly.

reiterl:
I can reproduce the problem. I think the second dropdown list is not enabled again.

reiterl:
This problem is similar to #6327

Elblinator:
https://github.com/OpenSlides/openslides-client/pull/6380

PR #6380, opened 2026-07-09
Use isListEmpty for disable list-search-selectors

### LINKED ISSUE #6327: Meeting edit: page reload leads to disabled field

reiterl:
Resolve #6327 #6273 

As described in #6363  I have used an `isListEmpty` property and override the get (and set) of disabled.

Tested with:
- meeting-edit (duplicate from)
- participant-list-info (Str lvl)
- history (id field)

### DIFF
--- a/client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
+++ b/client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
@@ -50,7 +50,7 @@ export class ListSearchSelectorComponent extends BaseSearchSelectorComponent {
     }
 
     public override get disabled(): boolean {
-        return this.isListEmpty() || this._disabled;
+        return super.disabled || (this.isListEmpty() && !this.clickNotFound.observed);
     }
 
     @Input()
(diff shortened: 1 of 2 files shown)


client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
```diff
@@ -49,4 +48,13 @@ export class ListSearchSelectorComponent extends BaseSearchSelectorComponent {
     public constructor(@Optional() @Self() ngControl: NgControl) {
         super(ngControl);
     }
+
+    public override get disabled(): boolean {
+        return this.isListEmpty() || this._disabled;
```
bastianjoel:
```suggestion
        return this._disabled || (this.isListEmpty() && !this.clickNotFound.observed);
```

bastianjoel: Also - isn't it possible to use `super.disabled` instead of `this._disabled`?

bastianjoel (changes requested):

client/src/app/ui/modules/search-selector/components/list-search-selector/list-search-selector.component.ts
reiterl: Yes, should be possible, changed.

bastianjoel (approved):

Elblinator (approved):

PR #6382, opened 2026-07-09
Exchange mat-time-picker

Elblinator:
The time picker in a motion changes from the clock to a list plus an input field where you can select the wanted adjust the time

The timepicker is only used in the motion when setting a submission date


bastianjoel:
@Elblinator there are still failing tests

bspekker (approved):

bastianjoel (approved):

PR #6385, opened 2026-07-10
Motion export: sanitize filename

### LINKED ISSUE #6271: Motion/amendment pdf Export: export not possible with dots in number
**Describe the bug**
If a motion or amendment number contains a dot, the single PDF Export won't work. If the user wants to save the pdf the file name will contains the dot structure of the number. Because of this the file name will end with the suffix after the dot number and not with .pdf.

Example: Compare motion titel (with number) and file name
Image

**How to Reproduce**
1. open a meeting
2. open motions
3. create a new motion > edit number of it to §8.12 > save
4. open the motion > three-dot-menu > export it

**Expected behavior**
If motion or amendment number contains a dot, the PDF export of it should be possible.

reiterl:
Resolve #6271 

Sanitize filenames to allow a working export. Dots and another special characters can lead to problems on different
systems.

### DIFF
--- a/client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
+++ b/client/src/app/gateways/export/pdf-document.service/pdf-document.service.ts
@@ -570,7 +570,7 @@ export class PdfDocumentService {
                 landscape: false,
                 imageUrls: imageUrls
             }),
-            filename: `${filetitle}.pdf`,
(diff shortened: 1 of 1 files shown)


bastianjoel (approved):

Elblinator (approved):

Elblinator (approved):
@MSoeb also approves

PR #6390, opened 2026-07-10
Added frontent for maximum available yes votes

potatojuicemachine:
This PR adds the frontend for the maximum available yes votes that has been implemented recently in the backend. It is a response (and hopefully closes) to this issue: https://github.com/OpenSlides/openslides-client/issues/6351 .

Creating a new approval poll will looks like this:
image

I added the information to the meta component so it looks like this:
image

I added the information to the top of the vote form:
image

When enough yes votes are selected the inputs will be disabled so they look like this:
image

Looking forward to your feedback. 

Have a nice day.

### DIFF
--- a/client/src/app/site/pages/meetings/modules/poll/services/voting.service/voting.service.ts
+++ b/client/src/app/site/pages/meetings/modules/poll/services/voting.service/voting.service.ts
@@ -75,6 +75,7 @@ export class VotingService {
      * checks whether the operator can vote on the given poll
      */
     public votingProhibited(poll: ViewPoll, user?: ViewUser): Observable<VotingProhibition | null> {
+        // TODO: [potatojuicemachine::10.07.2026] user being undefined will throw here. Not sure what the correct fix would be.
         return combineLatest([
             this.userRepo.getViewModelObservable(user.id),
             this.pollRepo.getViewModelObservable(poll.id),
--- a/client/src/app/domain/models/poll/poll-config-rating-approval.ts
+++ b/client/src/app/domain/models/poll/poll-config-rating-approval.ts
@@ -9,6 +9,7 @@ export class PollConfigRatingApproval extends BasePollConfigModel<PollConfigRati
 
     public max_options_amount!: number;
     public min_options_amount!: number;
+    public max_yes_amount!: number;
     public allow_abstain!: boolean;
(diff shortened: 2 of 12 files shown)


potatojuicemachine:
Note: I added a TODO comment for a small bug I discovered along the way in client/src/app/site/pages/meetings/modules/poll/services/voting.service/voting.service.ts , that I didn't know how to fix. I can remove the comment if you want.

bastianjoel (approved):
I did some minor changes myself because we are a bit on a schedule here and now is a good time to get it in ;) 
Anyways, thanks for your contribution!

client/src/app/site/pages/meetings/modules/poll/services/voting.service/voting.service.ts
```diff
@@ -75,6 +75,7 @@ export class VotingService {
      * checks whether the operator can vote on the given poll
      */
     public votingProhibited(poll: ViewPoll, user?: ViewUser): Observable<VotingProhibition | null> {
+        // TODO: [potatojuicemachine::10.07.2026] user being undefined will throw here. Not sure what the correct fix would be.
```
bastianjoel, line 78: @potatojuicemachine Did you actually run into this case? As far as I can see there is no way this method is called without an user. 

Is it possible that you ran into the issue that `user.id != user.meeting_user.id`? The call in `this.pollRepo.pollBallotsByUser` was wrong. This is fixed in #6420 https://github.com/OpenSlides/openslides-client/pull/6420/files#diff-08dcb3f8e940159da3a03854f990f799103919cec8d1f7cc424225cdc190c6edR82

potatojuicemachine: No. I came across this when I tried to figure out how the voting part of the frontend works. I wondered why the user can be optional while user.id is not. I tried to fix it but didn't understand enough to fix the function. Then I tried to figure out where this is called and I wasn't able to deduct if I can remove the '?'. So I just wrote a comment so that if someone else comes here they can fix the bug or fix the function declaration.

bastianjoel: Alright, in that case it is now fixed in #6420.

PR #6391, opened 2026-07-13
Add Tooltips and aria-labels to list views

### LINKED ISSUE #6370: Improve accessibility of metainfo column for screenreaders
**Current behavior**
In several meeting menus the lists contain some meta info with an icon and then the entry.
Example in agenda list:

Image

The screenreader does not read out the icon or what they stand for, but only the entry.

**Wanted behavior**
A screenreader should read out the label for the icon and then the entry.

Example from picture above: 
1. Visibility public
2. Visibility internal. Duration 0:25h
3. Tag Samstag. Visibility public

This functionality should be added for all list views in meetings:

- [x] agenda (visibility, duration, tags)
- [x] motions (category, motion block, tags) 
- [x] election **(Should be  a ToolTip phase)**
- [x] participants (groups, principals, proxy, structure level, participant number)
- [x] files (restriction, public file)

**Additional context**
The list views on the orga level also need work on this but first need to ensure that the metainfo can be navigated to at all in those views (committees, accounts, files). An additional issue is needed for this.

Elblinator:
resolves  #6370


Elblinator:
Given my comment in the issue:
>> files (restriction, public file)
>
>As far as I understand the guideline for aria labels:
>Aria labels should only be added if it's a clickable event which is not explained by the content inside
>given that a public file is a) not a clickable evene and b) has already a tooltip I did not add anything for it

I am not positive that the aria-label in the participant-list with the delegatioons is redundant given the same tooltip

bastianjoel (approved):

bspekker (changes requested):
With the changes my screen reader only reads out the icons but not the entries. For example it now says only "Visibility" where it previously said only "internal".

Image

Instead both aspects should be read out leading to "visibility internal"

bspekker (approved):

bastianjoel (approved):

PR #6392, opened 2026-07-13
Angular 22

### LINKED ISSUE #6280: Angular 22 update
This issue should be used to keep track of the angular 22 update. 

- [x] check if experimential api used in `feature/vote` are unchanged
- [ ] Check library compatibility. Following list contains packages not explicitly updated for angular 22. Libraries without update for +1 year are marked as stale. We might want to reevaluate usage of these. 
  - [ ] @coreui/angular-chartjs (pre-release available)
  - [x] ngx-date-fns (stale but might be okay)
  - [x] ngx-device-detector
  - [x] ngx-file-drop (stale)
  - [x] ngx-material-timepicker (stale) (replaced in #6382)
  - [x] ngx-papaparse (stale)
  - [x] @ngx-pwa/local-storage
  - [x] @ngx-translate/core
  - [x] @ngx-translate/http-loader
  - [x] @vendure/ngx-translate-extract
- [ ] Evaluate if switch to `@Service` decorator makes sense and how it can be done automatically

@ngx-translate also released a new major version introducing breaking changes that affect us. This library should be updated together with angular. 

**Notes:**
- [x] [eslint-plugin-unused-imports](https://github.com/sweepline/eslint-plugin-unused-imports) currently has a broken version number specified for `@typescript-eslint/eslint-plugin` peerDependency. Therefore `--force` is needed to update eslint related packages. Check if `@typescript-eslint/no-unused-vars` can just be used instead. 
- [x] Some constructor calls of overloaded classes have changed due to `inject` adaption.
- [ ] `vite` is now the default test runner. We should adapt that change.
- [x] `ngx-material-timepicker` has to get replaced
- [x] `baseUrl` is deprecated in tsconfig as of https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---baseurl. We need to either update our imports or get the replacement working. The suggested update path from the docs does not seem to work for us. 
- [x] In preparation all auto generated `.spec.ts` files that only contain disabled tests should be removed.

bastianjoel:
resolves #6280


Elblinator (commented):
only one thiings was found:
- [x] Orga > settings timezone cannot be changed

bastianjoel:
> Orga > settings timezone cannot be changed

Not related to the update. I fixed it anyway.

Elblinator (approved):

Issue #6224, opened 2026-05-21
Fix long reasons extending the motion card on desktop view


bspekker:
**Describe the bug**
If some content in a motion does not wrap and extends the boundaries of the motion card, this might cause other elements to extend the card as well (e.g. reason, change recs, summary of changes, motion text toolbar).
This was fixed for small screens with #5693 but is broken in desktop view.

**How to Reproduce**
1. create a motion with a reason that contains one string without spaces.
2. save the motion
3. create a change recommendation to the motion

--> the display of the change rec and reason will look similar to this:
<img width="1249" height="645" alt="Image" src="https://github.com/user-attachments/assets/420d709b-ee85-45ae-a105-4614f57623db" />

**Expected behavior**
Make sure that content in the motion card will never extend the boundaries of the motion card.

PR #6410, opened 2026-07-15
Fix checkbox with reason width in amendment

### LINKED ISSUE #5982: Fix text width of reason in amendments

Elblinator:
resolves #5982

As far as I could see removing the class is not schangeing anything from the reason text


bspekker (changes requested):
#6224 is not fixed:

Image

#5982 is fixed and no other side effects seem to appear.

bspekker (approved):

bastianjoel (approved):

PR #6417, opened 2026-07-16
Migrate to Service() and use inject() in the services

### LINKED ISSUE #6407: Migrate services to `@Service()` decorator
Follow up to #6280.

`@Injectable({ providedIn: 'root' })` can get replaced with `@Service()`. Note that this requires the injected services in that class to be injected with `inject` and not via the constructor. Otherwise it will throw an error.

reiterl:
Resolve #6407

### DIFF
--- a/client/src/app/gateways/repositories/action-worker/action-worker-repository.service.ts
+++ b/client/src/app/gateways/repositories/action-worker/action-worker-repository.service.ts
@@ -1,15 +1,14 @@
-import { Injectable } from '@angular/core';
+import { inject, Service } from '@angular/core';
 import { ActionWorker } from '@app/domain/models/action-worker/action-worker';
 
 import { BaseRepository } from '../base-repository';
 import { RepositoryServiceCollectorService } from '../repository-service-collector.service';
 import { ViewActionWorker } from './view-action-worker';
 
-@Injectable({
-    providedIn: `root`
-})
+@Service()
 export class ActionWorkerRepositoryService extends BaseRepository<ViewActionWorker, ActionWorker> {
-    public constructor(repositoryServiceCollector: RepositoryServiceCollectorService) {
+    public constructor() {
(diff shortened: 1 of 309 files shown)


client/src/app/gateways/repositories/action-worker/action-worker-repository.service.ts
```diff
@@ -1,15 +1,14 @@
...
 
-@Injectable({
-    providedIn: `root`
-})
+@Service()
 export class ActionWorkerRepositoryService extends BaseRepository<ViewActionWorker, ActionWorker> {
-    public constructor(repositoryServiceCollector: RepositoryServiceCollectorService) {
+    public constructor() {
+        const repositoryServiceCollector = inject(RepositoryServiceCollectorService);
         super(repositoryServiceCollector, ActionWorker);
```
bastianjoel, line 12: I am a bit confused why you did not use the chance to move the injection into `BaseRepository` while at it. The prevention of unnecessary constructor chaining is the main reason for the migration to `inject` in the first place. 

There are other cases where this would have been possible. I guess most cases should be possible with find and replace. Please go through it and at least fix the simple cases.

bastianjoel (changes requested):

client/src/app/gateways/repositories/action-worker/action-worker-repository.service.ts
reiterl: Moving the injections in the `BaseRepository` is more complicated.
To use `inject` in the base repo, it needs to be a Service or Directive, but that clashes with the way the `ModelConstructor` is given to the class.

bastianjoel (approved):

client/src/app/gateways/repositories/action-worker/action-worker-repository.service.ts
bastianjoel: This is now resolved in #6485.

PR #6437, opened 2026-07-23
Fix livestream double audio

### LINKED ISSUE #6310: m3u8 livestream often initializes playing only audio, component missing in DOM
**Describe the bug**

Especially in Firefox, the video player starts playing an m3u8 livestream (on the same domain), but the video is not or only very briefly (<<1sec) shown.
Instead, the message `Restart livestream` is visible, while audio is playing in the background. If `Restart livestream` is clicked, the component initializes a second time successfully with video - **leading to audio being played twice with a short delay**.

The following warning is logged in the console: `VIDEOJS: WARN: The element supplied is not included in the DOM`

When debugging this a bit, it looks like the whole os-video-player component is missing from the DOM and `ngOnDestroy()` is called shortly after `ngAfterViewInit()`, but before the videojs player was initialized, though I could not find out why the component was destroyed in the first place.

Firefox seems to trigger it almost every time, but I also managed to trigger the issue in Chromium with a few reloads.

**How to Reproduce**
1. set a livestream URL to a m3u8 URL
2. (re-)load the page with an active livestream

**Expected behavior**
The video-player initializes successfully at page load, playing audio and (visible) video.

Running release: 4.3.0

Elblinator:
resolves #6310

### DIFF
--- a/client/src/app/site/pages/meetings/pages/interaction/modules/interaction-container/components/video-player/video-player.component.ts
+++ b/client/src/app/site/pages/meetings/pages/interaction/modules/interaction-container/components/video-player/video-player.component.ts
@@ -8,7 +8,7 @@ import {
     Input,
     OnDestroy,
     Output,
-    ViewChild,
+    viewChild,
     ViewEncapsulation
 } from '@angular/core';
 import { MeetingSettingsService } from '@app/site/pages/meetings/services/meeting-settings.service';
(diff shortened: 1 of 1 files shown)


Elblinator:
@oe1rfc I could reproduce the  bug and in my setup this fix solves the problem. 
You are welcome to test it and give feedback \o/

oe1rfc:
@Elblinator thank you very much, I tried your patch and it looks good, I couldn't reproduce the double-audio issue anymore :tada: In my setup, Firefox sometimes initially doesn't want to start the steam, but I think that's not really an issue.
My feeling is that the main reason for the issue is that the video-player component is sometimes initialized, destroyed and re-initialized again, though I couldn't find out why that would be the case.

bastianjoel (approved):

PR #6446, opened 2026-07-27
Fix duplicate field meeting create form

### LINKED ISSUE #6411: Meeting create form: duplicate from field is not always usable
**Describe the bug**
It can happen, that the 'duplicate from' field is grayed out and a reload is needed to made it usable.

Image

**How to Reproduce**
1. open an instance
2. create a committee
3. create a meeting and set it as template
4. create a new meeting a try to select the template from the 'duplicate from' field. > If the field is not grayed out, reload the page until it is. 

**Expected behavior**
If an account has the permission to use the 'duplicate from' field in the meeting create form, the field should always be available and useable.

Elblinator:
resolves #6411


MSoeb (approved):
Works as intended.

Elblinator:
Bugs fixed by other PR

PR #6467, opened 2026-08-03
Fix action worker watch during paused autoupdate connection

### LINKED ISSUE #6466: Action worker requests should be excluded from subscription merging
Autoupdate subscriptions to the action worker should always be in its own subscription. 

A rule should be added to the shared worker to never merge them with other subscriptions as they constantly receive data on a very limited scope. With big subscriptions this can have negative impacts with the current way the autoupdate is implemented.

### LINKED ISSUE #6505: Autoupdate pausing breaks action worker wait
**Describe the bug**
When importing a lot of participants (15k+) and then switching focus to another browser tab the import is displayed as failed. 

**How to Reproduce**
1. Import 15k participants
2. Focus on another browser tab for 10+ minutes
3. Focus on the tab with the import again
4. An error message should show up

**Expected behavior**
The action worker banner should remain visible and no error should be displayed.

bastianjoel:
This fixes an issue where the current action worker was displayed as failed after the tab was not in focus for some time. 

resolves #6505
Also resolves #6466


Elblinator (approved):

client/src/app/gateways/action-worker-watch/action-worker-watch.service.ts
```diff
@@ -212,6 +214,7 @@ export class ActionWorkerWatchService {
         let hasReportedSlowness = false;
         let lastRefreshedInactivityReport = 0;
         let dataLoaded = false;
+        let lastTabInvisible = 0;
```
Elblinator, line 217: "lastTabInvisible"
describing when a tab  started to be invisible? 
Either change Invisible to Visible
or last to start/startpoint

Elblinator (changes requested):

Elblinator (commented):
My requested changes are just redactorial and do not change functional tesing

client/src/app/gateways/action-worker-watch/action-worker-watch.service.ts
bastianjoel: > describing when a tab started to be invisible?

Correct. 

> or last to start/startpoint

No. It contains the last time the tab was marked invisible. It is not updated after the tab is visible again. Therefore I consider `start` misleading.

Elblinator: Would you be okay with ```lastTabMarkedInvisible```?

bastianjoel: I do not see the necessity.

Elblinator (approved):

bspekker (changes requested):
The fix doesn't fully work. The import worked once but I maybe have not followed the exact reproduction in that case.
If I focus on a different page while preview is loading, I will get an error message (in the console and on the page).
If I focus on the page while preview is loading, then click on Import and focus on a different page I will also get the error message after checking back ~10 minutes later.

Error in console: "Error: Client has stopped watching worker "participant.import": Process has been assumed to be dead"
Error report: {"id":60,"name":"participant.import","state":"running","created":1788261860,"timestamp":1788262161,"closed":1788262564}

Issue #6120, opened 2026-04-24
Improve translation mapping for enum like types


bastianjoel:
We currently do not have a consistent way to handle translation of things like enums. These usually come from the database as strings and need to be mapped to a human readable word. 

One way we currently deal with this is by defining constant objects that can be used throughout the application. When we want to use them in a component we either have to add those as a public attribute or add them to the view model as `<field>Verbose`.
https://github.com/OpenSlides/openslides-client/blob/b5da8307ad7f9075a383c73b2a3bfe5d059576f1/client/src/app/site/pages/meetings/pages/polls/view-models/view-poll.ts#L76

Neither of those options is ideal as this always requires searching for those types and potentially overseeing them. Also both ways are quite verbose. 

To improve the situation a central file that contains keyed translations should be created. 
This file should contain key value pairs of enums/models and everything else that should be mapped from database strings. 

This should be implemented with the missing translation handler. Keyed translations should get a prefix like `.`. 

#5534 will implement this. This issue is for documentation and to keep track of adoption.

PR #6483, opened 2026-08-04
Change error map to handle templated error messages from backend

luisa-beerboom:
Result of https://github.com/OpenSlides/openslides-backend/issues/3674 
Needs https://github.com/OpenSlides/openslides-backend/pull/3705 to work

This probably won't work until what @bastianjoel wrote in https://github.com/OpenSlides/openslides-client/pull/6483#pullrequestreview-4867386358 is done


client/src/app/gateways/error-mapping/error-map-utils.ts
```diff
@@ -63,6 +65,48 @@ const MeetingArchiveErrorMap: ErrorMap = new ErrorMap([
         )
     ]
 ]);
+const json_upload_decimal_error_messages = {
```
bastianjoel: We do not use snake case here.

bastianjoel (commented):
I really don't like extending this. 

We already have https://ngx-translate.org/getting-started/translating-your-components/#with-parameters available in `feature/vote` via https://github.com/OpenSlides/openslides-client/pull/6153/commits/4a32f08deef124ccf16502591b0ad194336a7d2b#diff-27642ca8b5c52c5d46eb8ea14c441a142374e0f4a41f12bcace2c4b6a1561756.

Also https://github.com/OpenSlides/openslides-client/issues/6120 which could be used to parse more generic error strings is already implemented at pipe level there. 

Please check if those things would help. Maybe together with a new action error response format (e.g. something like `[{ error: 'format_error', params: { field: 'is_active' }}]`). 
If it helps I can cherry pick those things to main and if needed add a service for #6120.

client/src/app/gateways/error-mapping/error-map.service.ts
```diff
@@ -22,7 +22,11 @@ export class ErrorMapService {
      * @param errorMessage the error message that should be mapped and translated.
      * @returns A translated and (if possible) mapped "Error: <errorMsg>", or, if the ErrorMap contains an Error-object, said Error-object.
      */
-    public getCleanErrorMessage(errorMessage: string, options: GetHttpErrorMapOptions): string | Error {
+    public getCleanErrorMessage(
+        errorMessage: string,
+        options: GetHttpErrorMapOptions,
+        translation_args: Record<string, any>
```
bastianjoel:
```suggestion
        translationArgs: Record<string, any>
```

bastianjoel (approved):
I am adding a waiting label because we need to add https://github.com/OpenSlides/openslides-client/pull/6153/files#diff-27642ca8b5c52c5d46eb8ea14c441a142374e0f4a41f12bcace2c4b6a1561756 and https://github.com/OpenSlides/openslides-client/pull/6153/files#diff-10241f66631c8b0097464f0f3dd2b398244b57ac6251e1c7c19b7cfc7aae18d2 to main before merging.

client/src/app/gateways/error-mapping/error-map.service.ts
luisa-beerboom: Done

bastianjoel:
Necessary changes added to main and merged into this PR.

Elblinator (approved):

Elblinator:
waiting until backend is ready to be merged/merged

Elblinator:
This PR should be mergable without breaking anything else

PR #6487, opened 2026-08-10
Fix timezone fields causing form to be marked as touched on init

### LINKED ISSUE #6326: Orga settings: Save button active without a change done

bastianjoel:
resolves #6326


client/src/app/site/services/time-zone.service.ts
```diff
@@ -30,24 +29,20 @@ class SearchSelectorHelper implements Selectable {
...
         if (this.timezonesCache) {
             return this.timezonesCache;
         }
-        const timezones = await this.getAvailableTimeZones();
-        let i = 0;
+        const timezones = this.getAvailableTimeZones();
         this.timezonesCache = timezones.map(value => {
-            i++;
-            return new SearchSelectorHelper(value, i);
+            return new SearchSelectorHelper(value, value);
```
Elblinator: I would like the following more together with the changed contructor
```suggestion
            return new SearchSelectorHelper(value, Number(value));
```

client/src/app/site/services/time-zone.service.ts
```diff
@@ -12,9 +11,9 @@ class SearchSelectorHelper implements Selectable {
     public value: string;
     public id: number;
 
-    public constructor(value, id) {
+    public constructor(value: string, id: string) {
```
Elblinator:
```suggestion
    public constructor(value: string, id: number) {
```

client/src/app/site/services/time-zone.service.ts
```diff
@@ -12,9 +11,9 @@ class SearchSelectorHelper implements Selectable {
...
         this.value = value;
-        this.id = id;
+        this.id = id as any;
```
Elblinator:
```suggestion
        this.id = id;
```

Elblinator (approved):

client/src/app/site/services/time-zone.service.ts
bastianjoel: Code included in your CRs is now completely removed.

bspekker (approved):

PR #6494, opened 2026-08-11
Clean-up payload from participant create

### LINKED ISSUE #6444: Participant user create `structure_level_ids` error
**Describe the bug**

Creating a new user in the participants menu under specific conditions causes backend to throw the following error:
`ERROR:openslides_backend.shared.handlers.base_handler:Error occured on index 0: Action user.create: data.structure_level_ids must be array`

Inspecting the Action Data reveals that that the field `structure_levels_ids` is set to None, which is not a valid value

**How to Reproduce**
1. Go to Participants of a Meeting
2. start creating a new User
3. Set the email and username to "test@test.com" (both of them)
4. Click Next (do not select alredy existing account)
5. Click Save
6. Error occurs

This error does *not* occur, when the email is left empty in the first menu and then set in the second menu.

**Expected behavior**

The field `structure_levels_ids` is set to `[]` when no structure level IDs are present or left out entirely, so that user creation can work without error.

**Additional context**

This bug does not occur in the [demo](https://demo.openslides.org/), but does occur in latest `main`.

Elblinator:
resolves #6444 

Given that the reproduction is **very** flaky I won't assign someone for funcitonal testing 
I tested in the setup from @Janmtbehrens who is runninng in this error repeatedly

### DIFF
--- a/client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
@@ -305,9 +305,9 @@ export class ParticipantCreateWizardComponent extends BaseMeetingComponent imple
                         delete payload[field];
                     }
                 }
-                // this.repo.create(payload);
+                this.repo.create(payload);
             }
-            // this.onCancel();
+            this.onCancel();
         };
     }
 
--- a/client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
+++ b/client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
@@ -300,11 +300,6 @@ export class ParticipantCreateWizardComponent extends BaseMeetingComponent imple
                     this.repo.setPresent(true, { ...payload, id: this._accountId }).resolve();
                 }
             } else {
-                for (const field in payload) {
-                    if (payload[field] === null) {
-                        delete payload[field];
-                    }
-                }
                 this.repo.create(payload);
             }
             this.onCancel();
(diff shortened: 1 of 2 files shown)


Janmtbehrens:
This is a working fix for me
Ive tried the steps outlined in https://github.com/OpenSlides/openslides-client/issues/6444
the error occured without this fix, and they did not occur when the fix was in.
Importantly, somehow, the error only appeared when I've written "test@test.com" as a username and then as an email manually and not via copy and paste

client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
```diff
@@ -300,6 +300,11 @@ export class ParticipantCreateWizardComponent extends BaseMeetingComponent imple
                     this.repo.setPresent(true, { ...payload, id: this._accountId }).resolve();
                 }
             } else {
+                for (const field in payload) {
```
bastianjoel: The issue states that there is a bug in the path where an account is preselected from the list. However - you are now applying a fix on the path where NO account is preselected. It might be that there is an issue here too (I don't think so) but in line 289-291 there is a very obvious spot that explains the error.

bastianjoel (changes requested):

client/src/app/site/pages/meetings/pages/participants/pages/participant-detail/pages/participant-detail-manage/components/participant-create-wizard/participant-create-wizard.component.ts
Elblinator: The reproduction from the issue is not great 
Nowhere in the issue is stated that an account has to be preselected and I saw how @Janmtbehrens runs into the error (and I ran into it myself) preselecting is actively not a precondition.

And: 
>  in line 289-291 there is a very obvious spot that explains the error

that remark cannot be true or be relevant here.
either there are two bugs and your place has also one 
But my fix cannot have anything to to with this line. 
If the program runs line 289-291 then my changed lines are skipped.

Elblinator: update the repo

Janmtbehrens:
Appears to be a fix as well, though recreation is flaky. Without the new fix it fails ~50% of the time and with the fix 0% of the time from my local tests

Janmtbehrens (approved):

bastianjoel (approved):

PR #6548, opened 2026-08-26
Fix reason text word breaking

### LINKED ISSUE #6224: Fix long reasons extending the motion card on desktop view

bastianjoel:
resolves #6224


Elblinator (approved):

bspekker (changes requested):
Doesn't work well for tablets: Image

This might be related..? https://github.com/OpenSlides/openslides-client/pull/5693#pullrequestreview-3623120805

bspekker (approved):

Elblinator (approved):

PR #6552, opened 2026-08-31
Disallow attribute interpolation

bastianjoel:
Long term substring interpolation should also be forbidden. Instead proper parametrized translation strings should be used. Right now I am too lazy to do that as this needs manual work. 
All usages of this are currently problematic. Therefore a new issue should be opened.

### DIFF
--- a/client/eslint.config.mjs
+++ b/client/eslint.config.mjs
@@ -141,7 +141,9 @@ export default tseslint.config({
         "@angular-eslint/template/attributes-order": ["error", {
             alphabetical: true,
         }],
-        "@angular-eslint/template/no-interpolation-in-attributes": ["error"],
+        "@angular-eslint/template/no-interpolation-in-attributes": ["error", {
+            allowSubstringInterpolation: true,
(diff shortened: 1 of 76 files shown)


client/eslint.config.mjs
```diff
@@ -141,7 +141,9 @@ export default tseslint.config({
         "@angular-eslint/template/attributes-order": ["error", {
             alphabetical: true,
         }],
-
+        "@angular-eslint/template/no-interpolation-in-attributes": ["error", {
+            allowSubstringInterpolation: true,
+        }],
```
Elblinator, line 146: If you activate it here doesn't the linter then fail, given that you didn't fix every place? 
Or did you fix every not substring?

Elblinator (approved):

client/eslint.config.mjs
bastianjoel: No, what I am asking for as a next step is to set `allowSubstringInterpolation` to `false`.

